PR #1, opened 2021-03-09
merges client code into the repo

Gozala:


client/src/api.d.ts
```diff
@@ -0,0 +1,106 @@
...
+   * time.
+   */
+  deals: Deals[]
+}
+
+/**
+ * Finilized deals. In this state all the deals are finilized and are not going
+ * to change.
+ */
+export interface FinilizedDeals {
```
alanshaw, line 57:
```suggestion
export interface FinalizedDeals {
```

client/src/api.d.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export interface FinalizedDeal {
+  status: "active" | "terminated" 
+  sequence: number
+  lastStatusChangeTimestamp: Date
+  miner: string
+  chainDealId: number
+  dealActivationTimestamp: Date
+  dealExpirationTimestamp: Date
```
alanshaw, line 94: Can we remove "Timestamp" from all of these? Seems overly verbose.

client/src/api.d.ts
```diff
@@ -0,0 +1,106 @@
...
+  lastStatusChangeTimestamp: Date
+  miner: string
+  chainDealId: number
+}
+
+export interface FinalizedDeal {
+  status: "active" | "terminated" 
+  sequence: number
+  lastStatusChangeTimestamp: Date
+  miner: string
```
alanshaw, line 91: Address from @glif/filecoin-address? IDK it seems weird to be returning some library types but not others. Personally I'd remove CID and just return strings until we have a good reason to return a CID object - they are just going to be serialized to string for storage on chain anyway so why not just do the simple thing and keep the bundle size small?

client/src/lib.js
```diff
@@ -0,0 +1,135 @@
...
+   */
+  constructor({ token, endpoint = new URL("https://api.nft.storage") }) {
+    /** @readonly */
+    this.token = token
+    /** @readonly */
+    this.endpoint = endpoint
+  }
+
+  /**
+   * @private
```
alanshaw, line 23: Is not private?

client/src/lib.js
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /**
+   * @param {API.Service} service
+   * @param {CID} cid
+   * @returns {Promise<API.StatusResult>}
+   */
+  static async status({ endpoint, token }, cid) {
+    const url = new URL(`/${cid}`, endpoint)
+    const request = await fetch(url.toString(), {
```
alanshaw, line 72: Can you pass the url object directly to fetch?

client/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "nft.storage",
+  "version": "0.0.0-dev",
+  "description": "nft.storage client",
+  "main": "lib.js",
+  "author": "Irakli Gozalishvili <contact@gozala.io>",
+  "license": "(Apache-2.0 AND MIT)",
+  "type": "module",
+  "dependencies": {
+    "multiformats": "4.5.3",
+    "typescript": "4.2.3"
```
alanshaw, line 11: Should be a devDep?

client/src/api.d.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export interface StatusResult {
+  cid: CID
+  deals: Deals,
+  pin: Pin
+  created: Date
+}
+
+export type Deals = OngoingDeals | FinalizedDeals
```
alanshaw, line 38: What is the reason for the distinction?

client/src/api.d.ts
```diff
@@ -0,0 +1,106 @@
...
+export interface Pin {
+  requestid: string
+  cid: CID
+  name?: string
+  status: PinStatus
+  created: Date
+}
+
+
+export type PinStatus = "queued" | "pinning" | "pinned" | "failed"
```
alanshaw, line 106: Is this better as an enum so people can do comparisons easier?

client/src/lib.js
Gozala: Type checker won’t like it because type signature expects string not an URL instance.

client/src/api.d.ts
Gozala: Broadly speaking it given a simple way for user to tell if this is in progress or done ? More nuanced answer is ongoing deals are not finalized and deals array may change over time, deals added removed, their states update etc... Once it’s finalized it’s done and all deals are `FinalizedDeal` that specific fields that may not be present when in `OngoingDeals` state.

client/src/api.d.ts
Gozala: Not sure what exactly you mean here, if you’re referring to TS enum it doesn’t really work well with ts via jsdoc and unions provide just as good type safety.

client/src/api.d.ts
```diff
@@ -0,0 +1,106 @@
...
+  chainDealId: number
+}
+
+export interface FinalizedDeal {
+  status: "active" | "terminated" 
+  sequence: number
+  lastStatusChangeTimestamp: Date
+  miner: string
+  chainDealId: number
+  dealActivationTimestamp: Date
```
Gozala, line 93:
```suggestion
  dealActivation: Date
```

client/src/api.d.ts
Gozala, line 94:
```suggestion
  dealExpiration: Date
```

client/src/lib.js
Gozala, line 23:
```suggestion
```

client/src/lib.js
```diff
@@ -0,0 +1,135 @@
+import * as API from "./api.js"
+import CID from "multiformats/cid"
+
+
+
+/**
+ * @implements {API.Service}
```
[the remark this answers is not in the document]
Gozala: We could. We we’ve been using TS via JSDoc syntax everywhere else so I just chose to do the same here.

client/src/api.d.ts
Gozala: Mostly I thought CID made sense because it’s used and expected in that form everywhere else, it also allows user to choose a serialization format (base encoding). That said I think you’re making a good argument that most likely it will be just turned into a string so decoding to CID just to serialize it back to string provides little value, while introducing some overhead.

How about we keep CID for now, we can always get rid of them later.

PR #13, opened 2021-03-12
fix: endpoint URL

alanshaw:
This will not be at `api.` for ease...

### DIFF
--- a/client/src/lib.js
+++ b/client/src/lib.js
@@ -15,7 +15,7 @@ export default class NFTStore {
    * @param {string} options.token
    * @param {URL} [options.endpoint]
    */
-  constructor({ token, endpoint = new URL("https://api.nft.storage") }) {
+  constructor({ token, endpoint = new URL("https://nft.storage") }) {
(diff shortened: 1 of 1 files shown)


client/src/lib.js
```diff
@@ -15,7 +15,7 @@ export default class NFTStore {
    * @param {string} options.token
    * @param {URL} [options.endpoint]
    */
-  constructor({ token, endpoint = new URL("https://api.nft.storage") }) {
+  constructor({ token, endpoint = new URL("https://nft.storage") }) {
```
hugomrdias, line 18:
```suggestion
  constructor({ token, endpoint = new URL("https://nft.storage/api") }) {
```

hugomrdias (changes requested):

client/src/lib.js
alanshaw: The methods add /api already!

Gozala (approved):

Gozala (approved):

PR #17, opened 2021-03-16
refactor: move deals into separate KV store

alanshaw:
cc @ribasushi


alanshaw:
Merging this so it can be deployed.

PR #34, opened 2021-03-18
fix: remap exports to make web default

Gozala:


client/src/lib.js
```diff
@@ -15,7 +15,7 @@
  */
 
 import * as API from "./lib/interface.js"
-import { fetch, File, Blob, FormData } from "./platform.js"
+import { fetch, File, Blob, FormData } from "nft.storage/src/platform.js"
```
alanshaw, line 18: Not sure I understand this change?

Gozala: > Not sure I understand this change?

This affects how [package "exports"](https://nodejs.org/api/packages.html#packages_exports) are interpreted in nodejs, specifically if it would ignore the mapping if export uses relative paths, but will not if used qualified path. This change allows us to load `./platform.js` in contexts that disregard package exports but apply those in node to load `./platform.node.js` instead.

alanshaw: How does this work in the browser though? That's not a URL...

Gozala: Your are correct it will not work in browser directly. But tools like unpckg take source like this

https://unpkg.com/browse/@web-std/blob@2.0.1/src/lib.web.js

```js
import { TextEncoder, TextDecoder } from "web-encoding"
....
```

and rewrite it for browsers as follows https://unpkg.com/@web-std/blob@2.0.1/src/lib.web.js?module

```js
import { TextEncoder, TextDecoder } from "https://unpkg.com/web-encoding@1.1.0?module";
```

Gozala: Although this does not seem to work on skypack.dev 😭

PR #35, opened 2021-03-19
feat: add deals types and schemas, remove links

hugomrdias:
Move formatting to root


alanshaw (approved):
LGTM. We need to also update the deal types in the client.

PR #47, opened 2021-03-24
feat: prom metrics

alanshaw:
This PR adds MVP prometheus metrics reporting. It adds a `/metrics` endpoint that renders metrics in prometheus exposition format.

Due to the KV format we have to read every nft and deal to collect metrics. Since the size of the data is small right now it doesn't take long. The result is cached by cloudflare for 10 minutes.

It gives us the ability to build graphs for monitoring and alerting in Grafana:

Screenshot 2021-03-24 at 15 38 20

### DIFF
--- a/site/src/routes/metrics.js
+++ b/site/src/routes/metrics.js
@@ -141,7 +141,7 @@ function exportPromMetrics ({ userMetrics, nftMetrics }) {
     '# HELP nftstorage_nfts_storage_filecoin_total Total number of NFTs stored on Filecoin in active deals.',
     '# TYPE nftstorage_nfts_storage_filecoin_total counter',
     `nftstorage_nfts_storage_filecoin_total ${nftMetrics.storage.filecoin.total}`,
-    '# HELP nftstorage_nfts_storage_filecoin_queued_total Total number of NFTs stored on Filecoin in active deals.',
+    '# HELP nftstorage_nfts_storage_filecoin_queued_total Total number of NFTs queued for the next deal batch.',
     '# TYPE nftstorage_nfts_storage_filecoin_queued_total counter',
     `nftstorage_nfts_storage_filecoin_queued_total ${nftMetrics.storage.filecoin.totalQueued}`,
     ...Object.entries(nftMetrics.storage.filecoin.deals).map(([ntwk, totals]) => [
--- a/site/src/routes/metrics.js
+++ b/site/src/routes/metrics.js
@@ -1,0 +1,153 @@
+import { stores } from '../constants.js'
+import { get as getDeals } from '../models/deals.js'
+
+const MAX_AGE_SECS = 10 // max age of a metrics response in seconds
+
+/**
+ * TODO: basic auth
+ * @param {FetchEvent} event
+ */
+export async function metrics(event) {
+  const cacheKey = new Request(new URL(event.request.url).toString(), event.request)
+  const cache = caches.default
+
+  let res = await cache.match(cacheKey)
+  if (res) return res
+
+  const [userMetrics, nftMetrics] = await Promise.all([getUserMetrics(), getNftMetrics()])
+  res = new Response(exportPromMetrics({ userMetrics, nftMetrics }))
+  // Cache the response for 60s
+  res.headers.append('Cache-Control', `s-maxage=${MAX_AGE_SECS}`)
+  event.waitUntil(cache.put(cacheKey, res.clone()))
+  return res
+}
+
+// TODO: keep running total?
+async function getUserMetrics () {
+  let total = 0
+  let done = false
+  let cursor
+  while (!done) {
+    // @ts-ignore
+    const users = await stores.users.list({ cursor })
+    total += users.keys.length
+    cursor = users.cursor
+    done = users.list_complete
+  }
+  return { total }
+}
+
+class DealTotals {
+  constructor () {
+    this.proposing = { total: 0 }
+    this.accepted = { total: 0 }
+    this.failed = { total: 0 }
+    this.published = { total: 0 }
+    this.active = { total: 0 }
+    this.terminated = { total: 0 }
+  }
+}
+
+class NftMetrics {
+  constructor () {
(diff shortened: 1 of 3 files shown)


site/src/routes/metrics.js
```diff
@@ -0,0 +1,154 @@
...
+import { get as getDeals } from '../models/deals.js'
+
+const MAX_AGE_SECS = 60 * 10 // max age of a metrics response in seconds
+
+/**
+ * TODO: basic auth
+ * @param {FetchEvent} event
+ */
+export async function metrics(event) {
+  const cacheKey = new Request(new URL(event.request.url).toString(), event.request)
```
Gozala, line 11: Mind adding a code explaining when new Request is needed here as opposed to just using `event.request`. Reading this I have 0 guesses.

Gozala (commented):
Left a comment, I'll defer rest to @hugomrdias

site/src/routes/metrics.js
alanshaw: Ya will do. I'm guessing since it's used as the cache key it's to remove varying client headers.

hugomrdias: since this is a pretty basic endpoint you can just use `/metrics` as the key no need to do `new Request` and `new URL().toString()` but i would add a cache busting string at the end to avoid issues when code changes and a new deploy is done. Check this https://github.com/filecoin-project/specs/blob/master/api/src/get.js for and example

alanshaw: hmm, but `event.request` is passed in to the constructor so it'll inherit those headers...

It's based on this btw: https://developers.cloudflare.com/workers/examples/cache-api

site/src/routes/metrics.js
```diff
@@ -0,0 +1,154 @@
...
+  const cache = caches.default
+
+  let res = await cache.match(cacheKey)
+  if (res) return res
+
+  const [userMetrics, nftMetrics] = await Promise.all([getUserMetrics(), getNftMetrics()])
+  res = new Response(exportPromMetrics({ userMetrics, nftMetrics }))
+  // Cache the response for 60s
+  res.headers.append('Cache-Control', `s-maxage=${MAX_AGE_SECS}`)
```
hugomrdias, line 20:
```suggestion
  res.headers.append('Cache-Control', `maxage=${MAX_AGE_SECS}`)
```

no need to you restrict the cache to just cache proxies, like this browser and CF will cache it

site/src/routes/metrics.js
```diff
@@ -0,0 +1,154 @@
+import { stores } from '../constants.js'
+import { get as getDeals } from '../models/deals.js'
+
+const MAX_AGE_SECS = 60 * 10 // max age of a metrics response in seconds
```
hugomrdias, line 4: we can do at least 1 hour here right ?

hugomrdias (changes requested):

site/src/routes/metrics.js
hugomrdias: Actually here i would do `public,max-age=600,stale-while-revalidate=3600`

hugomrdias (changes requested):

site/src/routes/metrics.js
hugomrdias: this will give us more certainty that we get fresh values in the metrics at least every hour

site/src/routes/metrics.js
alanshaw: Yep, depends what resolution we want to see...

site/src/routes/metrics.js
```diff
@@ -0,0 +1,154 @@
...
+export async function metrics(event) {
+  const cacheKey = new Request(new URL(event.request.url).toString(), event.request)
+  const cache = caches.default
+
+  let res = await cache.match(cacheKey)
+  if (res) return res
+
+  const [userMetrics, nftMetrics] = await Promise.all([getUserMetrics(), getNftMetrics()])
+  res = new Response(exportPromMetrics({ userMetrics, nftMetrics }))
+  // Cache the response for 60s
```
hugomrdias, line 19:
```suggestion
  // Cache the response for MAX_AGE_SECS
```

hugomrdias (changes requested):

site/src/routes/metrics.js
```diff
@@ -0,0 +1,155 @@
+import { stores } from '../constants.js'
+import { get as getDeals } from '../models/deals.js'
+
+const MAX_AGE_SECS = 600 // max age of a metrics response in seconds
+const STALE_WHILE_REVALIDATE_SECS = 3600
+const CACHE_KEY = '/metrics'
```
hugomrdias, line 6:
```suggestion
const nanoid = require('nanoid/non-secure')
const cacheBust = nanoid.nanoid()
const CACHE_KEY = '/metrics' + cacheBust
```

the cache bust is important, this will invalidate all caches when we deploy a new worker and avoid issues if the new worker changes the responses

hugomrdias (changes requested):

site/src/routes/metrics.js
alanshaw: Right but, do we need a whole a module to generate an universally unique id? Why not just add a suffix like `/metrics-1` iff we change the code? We only really want to invalidate the cache if we change the code right?

alanshaw: Wouldn't this generate a new code each time a worker is run?

hugomrdias: yes sure, the suggested change would just make this automatic. Im fine with both.

hugomrdias: no, the value of `cacheBust` is kept between worker runs. Each time we deploy a v8 isolate is created and keeps that value stable until a new deploy is made.

alanshaw: Ok well, I think we should just keep it simple for now. Given what this is being used for and the short time involved I can't imagine this being a huge problem even if there were no busting...

PR #56, opened 2021-04-06
feat: new store method that uploads token with all the assets and metadata

Gozala:
Supersedes #50


client/src/lib/interface.ts
```diff
@@ -106,3 +129,146 @@ export interface Pin {
...
+   */
+  locales: string[]
+}
+
+export interface StoreResult<T extends TokenInput> {
+  /**
+   * CID for the token that encloses all of the files including metadata.json
+   * for the stored token.
+   */
+  ipld: CID
```
alanshaw, line 200: IDK, for discussion...I think people are less likely to know what IPLD is.

```suggestion
  ipfs: CID
```

client/src/lib/interface.ts
```diff
@@ -106,3 +129,146 @@ export interface Pin {
...
+  metadata: URL
+
+  /**
+   * Actual token data in ERC-1155 format. It is matches data passed as `token`
+   * argument except Files/Blobs are substituted with corresponding `ipfs://`
+   * URLs.
+   */
+  data: Encoded<T, [[Blob, URL]]>
+
+  embed: Encoded<T, [[Blob, URL]]>
```
alanshaw, line 214: A nice comment for this one too please!

client/src/lib/interface.ts
```diff
@@ -106,3 +129,146 @@ export interface Pin {
...
+  /**
+   * CID for the token that encloses all of the files including metadata.json
+   * for the stored token.
+   */
+  ipld: CID
+
+  /**
+   * URL like `ipfs://bafy...hash/meta/data.json` for the stored token metadata.
+   */
+  metadata: URL
```
alanshaw, line 205: So this maps nicely to the `URI`/`tokenURI` function in the contract?

```suggestion
  uri: URL
```

client/src/lib/interface.ts
```diff
@@ -106,3 +129,146 @@ export interface Pin {
...
+
+  /**
+   * Actual token data in ERC-1155 format. It is matches data passed as `token`
```
alanshaw, line 208:
```suggestion
   * Actual token data in ERC-1155 format. It matches data passed as `token`
```

alanshaw (changes requested):
This is awesome work 🚀

I think that where we have `Token` in this code we should replace with `Metadata` (or similar) so we don't confuse with our API key `token`.

Please could we get a really good example added to the README.md in place of what's currently there? 🙏

client/src/lib.js
```diff
@@ -13,10 +13,14 @@
  * ```
  * @module
  */
-
+import { CID } from 'multiformats'
 import * as API from './lib/interface.js'
+import * as Token from './token.js'
```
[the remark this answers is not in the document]
Gozala: Since we are moving to different wallet based auth, I was expecting our auth `token` will go away so there will be no conflict 😅😳 ?

client/src/lib/interface.ts
Gozala: I honestly think it should be `cid`, but @mikeal thought `ipld` was better. If you both agree on `ipfs` that's good enough for me.

client/src/lib.js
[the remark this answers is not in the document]
hugomrdias: we will still have tokens

client/src/lib.js
```diff
@@ -112,6 +116,71 @@ class NFTStorage {
...
+
+    /** @type {API.StoreResponse<T>} */
+    const result = await response.json()
+
+    if (result.ok === true) {
+      const { value } = result
+      const data = Token.decode(value.data)
+
+      return {
+        ipld: CID.parse(value.ipld),
```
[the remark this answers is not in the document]
hugomrdias: yeah let just use cid strings

client/test/service.js
```diff
@@ -32,6 +35,79 @@ const importUpload = async (request) => {
   }
 }
 
+/**
+ * @param {File} file
+ * @returns {Promise<CID>}
+ */
+const importAsset = async (file) => {
+  const { cid } = await importDirectory([file])
```
Gozala, line 43: This will be one the backend

```js
ipfs.add([{ path: file.name, content: file.stream() ])
```

client/test/service.js
```diff
@@ -32,6 +35,79 @@ const importUpload = async (request) => {
...
+const importToken = async (request) => {
+  const contentType = request.headers.get('content-type') || ''
+  if (contentType.includes('multipart/form-data')) {
+    const form = await request.formData()
+
+    const data = JSON.parse(/** @type {string} */ (form.get('meta')))
+    const dag = JSON.parse(JSON.stringify(data))
+    const metadata = JSON.parse(JSON.stringify(data))
+
+    for (const [path, content] of form.entries()) {
```
Gozala, line 59: Example paths

```
image
properties.video.src.0
```

client/test/service.js
```diff
@@ -32,6 +35,79 @@ const importUpload = async (request) => {
...
+      },
+    }
+
+    return result
+  } else {
+    throw Error('/api/store expects multipart/form-data')
+  }
+}
+
+/**
```
Gozala, line 93:
```
setAt('a.b.c', { a: { b: { c: null }}}, 5) // sets c to 5
```

client/test/lib.spec.js
```diff
@@ -89,6 +89,173 @@ describe('client', () => {
...
+        description: 'stuff',
+        image: new File(['fake image'], 'cat.png', { type: 'image/png' }),
+        properties: {
+          extra: 'meta',
+          src: [
+            new File(['hello'], 'hello.txt', { type: 'text/plain' }),
+            new Blob(['bye']),
+          ],
+        },
+      })
```
Gozala, line 218: const data = new FormData()
data.set('image', File())
data.set('properties.src.0', new File())
data.set('properties.src.1': new Blob())
data.set('meta', JSON.stringify(metadata))
```

File: client/test/lib.spec.js [hunk 11]
```diff
@@ -89,6 +89,173 @@ describe('client', () => {
...
+    })
+
+    it('store with properties', async () => {
+      const client = new NFTStorage({ token, endpoint })
+      const result = await client.store({
+        name: 'name',
+        description: 'stuff',
+        image: new File(['fake image'], 'cat.png', { type: 'image/png' }),
+        properties: {
+          extra: 'meta',
```
Gozala, line 212:
```suggestion
          bla: 'ipfs://...'
```

client/test/service.js
```diff
@@ -32,6 +35,79 @@ const importUpload = async (request) => {
...
+    const dag = JSON.parse(JSON.stringify(data))
+    const metadata = JSON.parse(JSON.stringify(data))
+
+    for (const [path, content] of form.entries()) {
+      if (path !== 'meta') {
+        const file = /** @type {File} */ (content)
+        const cid = await importAsset(file)
+        const href = `ipfs://${cid}/${file.name}`
+        setAt(path.split('.'), dag, cid)
+        setAt(path.split('.'), data, { '@': 'URL', href })
```
Gozala, line 65:
```suggestion
        setAt(path.split('.'), data, { '@': 'CID', cid: cid.toString() })
```

client/src/lib.js
```diff
@@ -112,6 +116,71 @@ class NFTStorage {
...
+        metadata: new URL(value.metadata.href),
+        data,
+        embed: Token.embed(data, {
+          gateway: GATEWAY,
+        }),
+      }
```
Gozala, line 179: What about this return type

```ts
interface Token {
  cid: CID
  medataURL: URL
  toJSON(options?: { gateway?: URL }):TokenJSON
}
```

client/package.json
```diff
@@ -49,7 +50,7 @@
     "ipld-in-memory": "8.0.0",
     "mocha": "8.3.2",
     "multicodec": "3.0.1",
-    "multiformats": "4.5.3",
+    "multiformats": "^4.5.3",
```
Gozala, line 53:
```suggestion
    "multiformats": "^4.5.3",
    "@ipld/dag-cbor": "^4.0.0",
```

client/package.json
```diff
@@ -34,6 +34,7 @@
     "prepare": "npm run build"
   },
   "dependencies": {
+    "@ipld/dag-cbor": "^4.0.0",
```
Gozala, line 37:
```suggestion
```

client/src/lib.js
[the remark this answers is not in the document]
Gozala: > we will still have tokens

Could those be called something else like `auth` or `claim` or someting ? Because NFT does stand for Non-fungible token, all the metadata specs refer to token so calling it something else would be more confusing.

If we can't change the auth token name, then I suggest renaming this to `NFT` but instead.

client/src/lib.js
Gozala: Context: @hugomrdias did not liked that data and embed were basically same data but with different URL encodings. We have talked bunch on what a better representation would be and I've wrote it down here for further discussion.

client/src/lib/interface.ts
```diff
@@ -106,3 +129,146 @@ export interface Pin {
...
+  decimals?: number
+
+  /**
+   * Arbitrary properties. Values may be strings, numbers, nested object or
+   * arrays of values. It is possible to provide a `File` or a `Blob` instance
+   * as property value, in which case it is stored on IPFS and metadata will
+   * contain URL to it in form of `ipfs://bafy...hash/name.png` or
+   * `ipfs://bafy...file/` respectively.
+   */
+  properties?: Object
```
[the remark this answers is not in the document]
Gozala: I think we really should at least request it to be an object otherwise all of the nice type inference that took me a while to get will break. And really we should discourage sticking primitives or files in there.

client/src/lib/interface.ts
```diff
@@ -106,3 +129,146 @@ export interface Pin {
...
+
+export interface StoreResult<T extends TokenInput> {
+  /**
+   * CID for the token that encloses all of the files including metadata.json
+   * for the stored token.
+   */
+  ipld: CID
+
+  /**
+   * URL like `ipfs://bafy...hash/meta/data.json` for the stored token metadata.
```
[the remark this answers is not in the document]
Gozala: The reason is that in terms of IPLD node we have `meta` link to dag-pb dir with `data.json` as show here

image

We could instead create `metadata.json` file and a link in dag-cbor named `metadata.json`, but I felt that wrapping it in directory just like any other asset was a better compromise. It also creates a room for sticking some other metadata files if they will come relevant in the future.

That said I don't feel strongly either way and can update to that if you feel like it's better despite arguments above.

client/src/lib/interface.ts
Gozala: I was inclined to do exactly that (or `url` rather) but chose not to, because I think that `tokenURI` really should be pointing to the ipld dag as opposed to metadata file. That way we can in a future recommend that and expose `url` property as well in hope of getting folks to upgrade from metadata URL to actual token URL.

client/test/service.js
```diff
@@ -32,6 +35,79 @@ const importUpload = async (request) => {
...
+    }
+
+    return result
+  } else {
+    throw Error('/api/store expects multipart/form-data')
+  }
+}
+
+/**
+ * @template V
```
Gozala, line 94:
```suggestion
 * Sets a given `value` at the given `path` on a passed `object`.
 *
 * @example
 * ```js
 * const obj = { a: { b: { c: 1 }}}
 * setAt('a.b.c', obj, 5)
 * obj.a.b.c //> 5
 * ```
 * 
 * @template V
```

client/test/lib.spec.js
Gozala: Added this example to jsdoc comment of `Token.encode`

client/test/service.js
Gozala: added this to jsdoc example

client/src/lib.js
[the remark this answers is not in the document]
Gozala: Sync discussion we've decided to use different name for auth token.

Gozala:
I have added code to implement backend piece of this that stores NFT and all it's assets in our own IPFS node. However it is not working and I'm unable to make progress because:

1. Cloudfare seems to forbid http requests to IP addresses.
2. Cloudfare does not seem to like DNS record for our IPFS node (Here is their explanation https://support.cloudflare.com/hc/en-us/articles/115003011431-Troubleshooting-Cloudflare-5XX-errors#522error)

I've still pushed these changes to get more eyes on them. If have any ideas how to overcome stated issues please let me know.

Gozala:
Finally got things working, turns out cloudfare ignores ports which was the reason why requests were timing out 😖 With the help of @ribasushi I got things working and was able to run tests locally to confirm that content is created and pinned on our ipfs node here is an example:

http://experimental-ipfs.nft.storage:28080/ipfs/bafyreib75ot3oyo43f7rhdk6xlv7c4mmjwhbjjnugrw3yqjvarpvtzxkoi/
http://experimental-ipfs.nft.storage:28080/ipfs/bafyreib75ot3oyo43f7rhdk6xlv7c4mmjwhbjjnugrw3yqjvarpvtzxkoi/metadata.json

However there is still a problem that other ipfs nodes don't seem able to connect to our node. My local node fails to do so and content does not makes it's way neither to public gateway nor to pinata gateway. Once those issues are resolved there still going to be things we need to rethink:

1. When we tell pinata to pin by CID it's not going to keep the connection open. It will try and eventually either succed or give up.
   - We block and poll pinata, but that does not seems like a good idea. And I'm not even sure it would be compatible with cloudworker design.
2. We probably need a job that occasionally goes over the the pending pins and checks that they got picked up by pinata. Then update our pin state.
   - Our keyvaule store isn't going to fit this model because only one writing to pin status should be that job.
   - Do we retry pins that failed on pinata ? How many times do we retry ?
3. It would be difficult to not leak the complexity to the end user.
    - Does `client.store({ ... })` returns when store is complete or when it's submitted ? 
    - Do users need to keep polling to see when store is complete ?
    - Seems like we're trying to work around the fact that pinata has no API to ingest arbitrary dags. Maybe we should talk to pinata to see if we can get some other access to avoid introducing this complexity.

Gozala:
Update: now that node connectivity was resolved both pinata and gateway are picking things up when I test them locally.

site/src/constants.js
```diff
@@ -19,6 +19,10 @@ export const pinata = {
   jwt: PINATA_JWT,
 }
 
+export const ipfs = {
+  host: IPFS_HOST,
+}
```
ribasushi, line 24: @Gozala @alanshaw want to raise a point here, that for launch we should use a *slightly* better configured go-ipfs node ( with a behind-DNS IP etc )
You *could* aunch with what we have right now, but flagging that a new one is coming EOW/weekend when I get a chance to set it all up.

cc @mikeal

Gozala:
I have updated pull request so it now uses cluster and it is ready to be reviewed.

site/src/routes/nfts-store.js
```diff
@@ -0,0 +1,98 @@
...
+    },
+    codec: CBOR,
+    hasher: sha256,
+  })
+  const car = await CAR.encode([block.cid], [block])
+  const { cid, size } = await cluster.add(car)
+
+  // We do want worker to wait for this, but we do not want to
+  // block response waiting on this.
+  event.waitUntil(
```
Gozala, line 57: @alanshaw do we need to provide anything else to ensure that pinata will pin the CBOR nodes here, or will it just work out of the box ?

site/src/utils/utils.js
```diff
@@ -25,3 +25,30 @@ export async function timed(fn, label, ctx) {
...
+ * setIn(obj, ['a', 'b', 'c'], 5)
+ * obj.a.b.c //> 5
+ * ```
+ *
+ * @template V
+ * @param {any} object
+ * @param {string[]} path
+ * @param {V} value
+ */
+export const setIn = (object, path, value) => {
```
alanshaw, line 44: Can `client/test/service.js` use this?

alanshaw (approved):
This looks incredible! A few things that can be sent as followup PRs if you like:

* We need an update to the [OpenAPI schema](https://github.com/ipfs-shipyard/nft.storage/blob/main/website/public/schema.yml)
* We should alter the docs for the JS client on the homepage to use the new `store` method
* We should change the example [here](https://github.com/ipfs-shipyard/nft.storage/tree/main/client#usage) to use the new API
* We should add example(s?) [here](https://github.com/ipfs-shipyard/nft.storage/tree/main/client/examples)

site/src/cluster.js
```diff
@@ -1,5 +1,6 @@
 import { Cluster } from '@nftstorage/ipfs-cluster'
 import { cluster } from './constants.js'
+import { CID } from 'multiformats'
```
hugomrdias, line 3: can we just use the cid string ?

site/src/utils/utils.js
hugomrdias, line 44: can we just use https://github.com/angus-c/just#just-safe-set instead?

site/tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "moduleResolution": "node",
     "sourceMap": true,
     "esModuleInterop": true,
+    "skipLibCheck": true,
```
hugomrdias, line 12: why are we adding this ?

hugomrdias (changes requested):

site/tsconfig.json
Gozala: https://github.com/multiformats/js-multiformats#typescript-support

client/src/token.js
```diff
@@ -0,0 +1,188 @@
+import * as API from './lib/interface.js'
```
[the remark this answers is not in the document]
Gozala: https://github.com/ipfs-shipyard/nft.storage/pull/56/files#diff-b6a841a788ab169e9be137340e8760874cce89d1af73c1a73bee22d19b82fd27

site/src/cluster.js
Gozala: We need CIDs to assemble an IPLD node. I could move the the CID string parsing to the other module, if that is what you want but CID is needed one way or the other.

@hugomrdias please let me know if this is ok, or if you want me to move the parsing to other module.

site/src/utils/utils.js
Gozala: I was not sure how to go about without introducing client to server dependency. I'll just use package hugo linked

site/src/routes/nfts-store.js
Gozala: There were some problems, but now they're fixed and things are working 

image

site/src/utils/utils.js
alanshaw: oh, oops yes sorry ignore me.

site/src/cluster.js
hugomrdias: i would prefer to remove this here but its not a blocker

hugomrdias (changes requested):
LGTM, but i'm getting errors running api ts check 

did you enable the git hooks like this https://github.com/ipfs-shipyard/nft.storage#development ?

i added an issue to add a github action to run api tests in the CI to avoid these problems in the future

site/tsconfig.json
```diff
@@ -9,6 +9,10 @@
     "moduleResolution": "node",
     "sourceMap": true,
     "esModuleInterop": true,
+    // Need to disable this because generated typedefs omit generics which
+    // creates a problew for multiformats
+    // @see https://github.com/multiformats/js-multiformats#typescript-support
```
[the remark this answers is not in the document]
alanshaw: This PR updates to v7

alanshaw:
I upgraded to multiformats@8 - the same version `@ipld/car` uses and typescript seems to be happier.

site/tsconfig.json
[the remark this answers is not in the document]
Gozala: > but we are on 4.2

Generated types don't have type parameters possibly because multiformats is either on older version of TS or maybe that bug has not been fixed yet.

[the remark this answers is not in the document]
alanshaw: Oh 🤣 I just realised you meant typescript 4.2 - sorry ignore me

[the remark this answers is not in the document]
Gozala: I looked into this and seems that problem is not with multiformats, but rather with dag-cbor.

PR #60, opened 2021-04-20
feat: persist to IPFS Cluster

alanshaw:
TODO:

* Get DAG size from IPFS proxy (will send separate PR)

### DIFF
--- a/site/src/constants.js
+++ b/site/src/constants.js
@@ -18,6 +18,7 @@ export const auth0 = {
 export const cluster = {
   url: CLUSTER_URL,
   token: CLUSTER_TOKEN,
+  addrs: Object.freeze(CLUSTER_ADDRS.split(',')),
 }
 
 export const pinata = {
--- a/site/src/constants.js
+++ b/site/src/constants.js
@@ -26,7 +26,7 @@ export const cluster = {
     typeof CLUSTER_IPFS_PROXY_BASIC_AUTH_TOKEN !== 'undefined'
       ? CLUSTER_IPFS_PROXY_BASIC_AUTH_TOKEN
       : '',
-  addrs: Object.freeze(CLUSTER_ADDRS.split(',')),
+  addrs: Object.freeze(CLUSTER_ADDRS.split(',').filter(Boolean)),
 }
 
 export const pinata = {
--- a/site/src/constants.js
+++ b/site/src/constants.js
@@ -16,8 +16,16 @@ export const auth0 = {
 }
 
 export const cluster = {
-  url: CLUSTER_URL,
-  token: CLUSTER_TOKEN,
+  apiUrl: CLUSTER_API_URL,
+  basicAuthToken:
+    typeof CLUSTER_BASIC_AUTH_TOKEN !== 'undefined'
+      ? CLUSTER_BASIC_AUTH_TOKEN
+      : '',
+  ipfsProxyApiUrl: CLUSTER_IPFS_PROXY_API_URL,
+  ipfsProxyBasicAuthToken:
+    typeof CLUSTER_IPFS_PROXY_BASIC_AUTH_TOKEN !== 'undefined'
+      ? CLUSTER_IPFS_PROXY_BASIC_AUTH_TOKEN
+      : '',
   addrs: Object.freeze(CLUSTER_ADDRS.split(',')),
 }
 
--- a/site/src/constants.js
+++ b/site/src/constants.js
@@ -15,6 +15,11 @@ export const auth0 = {
   salt: SALT,
 }
 
+export const cluster = {
+  url: CLUSTER_URL,
+  token: CLUSTER_TOKEN,
+}
+
 export const pinata = {
   jwt: PINATA_JWT,
 }
--- a/site/src/routes/pins-add.js
+++ b/site/src/routes/pins-add.js
@@ -65,8 +65,7 @@ export async function pinsAdd(event) {
       size = parseInt(pin.metadata.size)
     }
     // we have this pin already allocated in our cluster, get the status...
-    const { peerMap } = await cluster.status(pin.cid)
-    status = cluster.toPSAStatus(Object.values(peerMap))
+    status = cluster.toPSAStatus(await cluster.status(pin.cid))
     // if this failed to pin, try again
     if (status === 'failed') {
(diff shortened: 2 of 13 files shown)


site/src/routes/pins-list.js
```diff
@@ -1,40 +1,199 @@
...
+      prefix: user.sub,
+      cursor,
+      limit: 1000,
+    })
+    for (const key of nftList.keys) {
+      if (!filter(key)) continue
+      count++
+      key.metadata.createdSort = key.metadata.created
+        ? new Date(key.metadata.created).getTime()
+        : 0
```
hugomrdias, line 52: maybe we should put the `created` in seconds when we add nfts instead of iso string, we could run a migration to change it to `s` or `ms` plus make sure all of them have it.

hugomrdias (commented):
left some comments inline.

where do we sync the pin status between the nfts kv and cluster ? does the pin add endpoint wait for it ?

site/src/constants.js
```diff
@@ -15,6 +15,20 @@ export const auth0 = {
   salt: SALT,
 }
 
+export const cluster = {
+  apiUrl: CLUSTER_API_URL,
+  basicAuthToken:
```
hugomrdias, line 20: we need to add these as secrets right ? if yes we should update readme instructions

site/src/routes/pins-add.js
```diff
@@ -3,43 +3,114 @@ import { verifyToken } from '../utils/utils.js'
...
+    if (typeof pinData.meta !== 'object' || Array.isArray(pinData.meta)) {
+      return new JSONResponse(
+        { error: { reason: 'INVALID_PIN_DATA', details: 'invalid metadata' } },
+        { status: 400 }
+      )
+    }
+    meta = Object.fromEntries(
+      Object.entries(pinData.meta).filter(([, v]) => typeof v === 'string')
+    )
   }
```
hugomrdias, line 46: can't we just try catch JSON.stringify ? plus kv metadata can only be 1024 bytes in length

site/src/routes/pins-add.js
```diff
@@ -3,43 +3,114 @@ import { verifyToken } from '../utils/utils.js'
...
+      created: created.toISOString(),
+    },
+  }
+  const metadata = {
+    name,
+    meta,
+    pinStatus: status,
+    size,
+    created: created.toISOString(),
   }
```
hugomrdias, line 104: we duplicate a bunch of stuff here, we should only keep the status and created in the metadata because we need that to filter in the list endpoint and deduplicate the rest inside the kv value

hugomrdias: i guess psa filters by meta too..

site/src/routes/pins-delete.js
```diff
@@ -9,29 +9,32 @@ import * as nfts from '../models/nfts.js'
...
+    )
   }
 
-  const res1 = await PinataPSA.pinsDelete(params.requestid)
-  if (!res1.ok) {
-    return PinataPSA.parseErrorResponse(res1.error)
-  }
-
-  await nfts.remove({ user, cid: pinStatus.pin.cid })
+  await nfts.remove({ user, cid })
```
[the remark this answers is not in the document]
alanshaw: It could have been pinned by someone else. Garbage collection is best done as a maintenance task. Also low priority, considering this is for NFTs, and they actually shouldn't ever be unpinned.

site/src/routes/pins-delete.js
```diff
@@ -9,29 +9,32 @@ import * as nfts from '../models/nfts.js'
...
-  if (!res0.ok) {
-    return PinataPSA.parseErrorResponse(res0.error)
+  if (!nft) {
+    // maybe this is an old Pinata pin?
+    const res = await PinataPSA.pinsGet(params.requestid)
+    if (res.ok) {
+      cid = res.value.pin.cid
+      nft = await nfts.get({ user, cid })
+    }
   }
```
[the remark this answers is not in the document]
alanshaw: So we're using the CID as the `requestid` to make this all work. We never stored `requestid`'s when proxying to Pinata and there's no way to get hold of them. So the user may still be holding on to them and so we _have_ to check.

site/src/routes/pins-add.js
```diff
@@ -3,43 +3,114 @@ import { verifyToken } from '../utils/utils.js'
...
+    // if we cached the size then use it
+    if (pin.metadata && pin.metadata.size) {
+      size = parseInt(pin.metadata.size)
+    }
+    // we have this pin already allocated in our cluster, get the status...
+    const { peerMap } = await cluster.status(pin.cid)
+    status = cluster.toPSAStatus(Object.values(peerMap))
+    // if this failed to pin, try again
+    if (status === 'failed') {
+      await cluster.recover(pinData.cid)
```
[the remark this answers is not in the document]
alanshaw: Firstly, because we're passing a CID this'll only recover a single pin. Recovery just re-attempts to pin the data if it has an error status. For example it may have tried for 24 hours (or whatever the timeout is) and then given up.

So this is just incase we tried to pin a thing before but it failed and we're attempting to pin the same thing again.

site/src/routes/pins-list.js
alanshaw: Yeah, this needs to be made good but we need a better database to do this right...

site/src/routes/pins-add.js
alanshaw: The pin `meta` MUST be an object and it MUST be a `Record<string, string>`. Any values that are not strings are ignored and not persisted. This is what the spec says.

This is just checking that the passed `meta` is a plain object and then filtering out any values that are not, in order to conform to the spec.

My gut instinct is that the majority of people will not encounter the 1024 bytes limit. If we find lots of people are seeing it then we'll prioritise moving to a new solution for listing pins sooner (the pin metadata is in the KV metadata so we can filter by it in the pin listing endpoint withough having to fetch every single NFT in the KV).

site/src/routes/pins-add.js
alanshaw: Yep, and then we also need `size` as it is used by the metrics endpoint in order to count the total size of all NFTs.

hugomrdias (approved):
lets merge it

PR #62, opened 2021-04-23
feat: cron jobs for metrics and pin status tracking

alanshaw:
This PR adds cron jobs for metrics generation and pin status / size tracking.

Previously the metrics were counted when the user hit the `/metrics` endpoint and then cached for 10 minutes. Now they are counted and stored in a new KV "METRICS". The `/metrics` endpoint has been updated to just read the latest values from the KV store.

The code to count metrics has also been updated to count filecoin deals in the "DEALS" KV, which resolves https://github.com/ipfs-shipyard/nft.storage/issues/59

There was a hack in the `/metrics` endpoint that was updating pin status and size for NFTs that were stored via the pinning services API. This has been pulled out into it's own dedicated cron job.

### DIFF
--- a/site/src/index.js
+++ b/site/src/index.js
@@ -45,6 +45,14 @@ r.add('get', '/new-key', newKey)
 r.add('get', '/new-file', newFile)
 // Monitoring
 r.add('get', '/metrics', metrics)
+// Cron testing
+// r.add('get', '/cron', async event => {
+//   await timed(updateUserMetrics, 'updateUserMetrics')
+//   await timed(updateNftMetrics, 'updateNftMetrics')
+//   await timed(updateNftDealMetrics, 'updateNftDealMetrics')
+//   await timed(updatePinStatuses, 'updatePinStatuses')
+//   return new Response('done')
+// })
 
 r.add('options', '/api/*', cors)
 // Remote Pinning API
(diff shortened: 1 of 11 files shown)


site/src/index.js
```diff
@@ -39,6 +45,14 @@ r.add('get', '/new-key', newKey)
...
 // Monitoring
 r.add('get', '/metrics', metrics)
+// Cron testing
+// r.add('get', '/cron', async event => {
+//   await timed(updateUserMetrics, 'updateUserMetrics')
+//   await timed(updateNftMetrics, 'updateNftMetrics')
+//   await timed(updateNftDealMetrics, 'updateNftDealMetrics')
+//   await timed(updatePinStatuses, 'updatePinStatuses')
+//   return new Response('done')
+// })
```
hugomrdias, line 55:
```suggestion
```

site/src/index.js
```diff
@@ -60,3 +74,19 @@ r.add('delete', '/api/internal/tokens', tokensDelete)
...
+      await timed(updateNftMetrics, 'CRON updateNftMetrics')
+    })()
+  )
+)
+addEventListener('scheduled', (event) =>
+  event.waitUntil(timed(updateNftDealMetrics, 'CRON updateNftDealMetrics'))
+)
+addEventListener('scheduled', (event) =>
+  event.waitUntil(timed(updatePinStatuses, 'CRON updatePinStatuses'))
+)
```
hugomrdias, line 92: why 3 listeners ?

hugomrdias (changes requested):

site/src/index.js
alanshaw: max 3 cron per script according to the docs

PR #64, opened 2021-04-26
refactor: switch to Pinata pinByHash

alanshaw:
This endpoint now supports CBOR data.

Also adds try/catch around calls to prevent response being sent early on failure.


hugomrdias (approved):
LGTM

PR #65, opened 2021-04-26
feat: new KV to index NFTs ORDER BY user ID, created DESC

alanshaw:
Adds a new KV that acts as an index of NFTs ordered by user ID then created. It allows us to list NFTs for a user in descending date order.

resolves https://github.com/ipfs-shipyard/nft.storage/issues/58

TODO:

* Needs https://github.com/ipfs-shipyard/nft.storage/pull/61 to merge for pagination in the UI (will send separate PR)
* ~~Will send separate PR for migration of existing NFTs~~ Added here


site/src/models/nfts.js
```diff
@@ -52,23 +79,59 @@ export const get = async (key) => {
...
+  }
+
+  const before = options.before || new Date()
+  if (!(before instanceof Date) || isNaN(before.getTime())) {
+    throw new Error('invalid filter')
+  }
+
+  let done = false
+  let cursor
+  /** @type Promise<KVValue<NFT>>[] */
```
hugomrdias, line 111: why did you change the type here ?

site/src/models/nfts.js
```diff
@@ -52,23 +79,59 @@ export const get = async (key) => {
...
+
+    if (nfts.length >= limit) {
+      break
+    }
+    cursor = nftIdxList.cursor
+    done = nftIdxList.list_complete
   }
-  return []
+
+  return Promise.all(nfts)
```
hugomrdias, line 136: i think we need to return the cursor here, frontend needs to know the cursor to do next page right ?

site/src/models/nfts.js
```diff
@@ -28,7 +49,13 @@ export const has = async (key) => {
 export const set = async (key, value, options) => {
   const savedValue = await get(key)
   if (savedValue === null) {
-    await stores.nfts.put(encodeKey(key), JSON.stringify(value))
+    const kvKey = encodeKey(key)
+    await stores.nfts.put(kvKey, JSON.stringify(value))
+    await stores.nftsIndex.put(
+      encodeIndexKey({ ...key, created: new Date(value.created) }),
```
hugomrdias, line 55: we should probably make an issue to change all these Dates to numbers and stop this back and forth between Date and number (ms)

hugomrdias (changes requested):

site/src/models/nfts.js
alanshaw: They just need to know the created date of the last item, which they can use as the `before` param for subsequent requests.

site/src/models/nfts.js
alanshaw: Since we `get` every key there's a small chance that the `get` will return null so really it's an array of `NFT|null` but we were ts-ignoring it before and it would have errored if that happened.

hugomrdias:
can you please rebase again, just to make sure the api deploy action works ?

site/src/models/nfts.js
alanshaw: https://github.com/ipfs-shipyard/nft.storage/issues/67

migrations/001-nfts-index.js
```diff
@@ -0,0 +1,75 @@
...
+  const indexTables = namespaces.filter((ns) => ns.title.includes('NFTS_IDX'))
+  const tables = indexTables
+    .map((idxTable) => {
+      const title = idxTable.title.replace('NFTS_IDX', 'NFTS')
+      return {
+        table: namespaces.find((t) => t.title === title),
+        index: idxTable,
+      }
+    })
+    .filter((t) => t.table.title.includes('staging'))
```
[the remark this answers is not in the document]
alanshaw: Good idea!

hugomrdias (approved):

PR #71, opened 2021-04-27
fix: update deal types to reflect actual structure

Gozala:
fixes: #39

### DIFF
--- a/client/src/lib.js
+++ b/client/src/lib.js
@@ -129,7 +129,7 @@ class NFTStorage {
     if (result.ok) {
       return {
         cid: result.value.cid,
-        deals: result.value.deals,
+        deals: decodeDeals(result.value.deals),
         size: result.value.size,
         pin: result.value.pin,
         created: new Date(result.value.created),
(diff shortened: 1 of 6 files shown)


alanshaw (changes requested):
Changes LGTM but we also need to update the status endpoint in the worker so it responds as the types suggest!

alanshaw (commented):
Shall we have lastChanged as a Date like we do for created at the top level?

Gozala:
> Changes LGTM but we also need to update the status endpoint in the worker so it responds as the types suggest!

It will because it just responds with what we have in kv store and that is what types reflect.

Gozala:
> Shall we have lastChanged as a Date like we do for created at the top level?

I do personally like using types representing values, however we moved to string representation for CID (for good reason) and also for URLs in #56 so I was no longer sure. I've added last commit that turns timestamps into Dates and I'll live it up to you to decide what's best here. If we decide to go with string we should probably change `created` to string as well.

alanshaw:
> > Changes LGTM but we also need to update the status endpoint in the worker so it responds as the types suggest!
> 
> It will because it just responds with what we have in kv store and that is what types reflect.

It responds with the previous structure described by the previous types: https://github.com/ipfs-shipyard/nft.storage/blob/main/site/src/routes/nfts-get.js#L32

> > Shall we have lastChanged as a Date like we do for created at the top level?
> 
> I do personally like using types representing values, however we moved to string representation for CID (for good reason) and also for URLs in #56 so I was no longer sure. I've added last commit that turns timestamps into Dates and I'll live it up to you to decide what's best here. If we decide to go with string we should probably change `created` to string as well.

Lets leave it as Date objects for now so this change is not more breaking than it needs to be.

alanshaw (approved):

Gozala:
> It responds with the previous structure described by the previous types: https://github.com/ipfs-shipyard/nft.storage/blob/c0a2a80a4e0115b24ee812b7b537ba37b3d3d12e/site/src/routes/nfts-get.js#L19-L28

I do not believe that is true, you can see `getDeals` used there just pulls it from kv store

https://github.com/ipfs-shipyard/nft.storage/blob/c0a2a80a4e0115b24ee812b7b537ba37b3d3d12e/site/src/models/deals.js#L11-L14

And KV holds just arrays of deals that have a same structure as typed here.

alanshaw:
I fixed it here https://github.com/ipfs-shipyard/nft.storage/commit/708778d87e4ceb2de8b7f19b3253f047bac5dae8

PR #75, opened 2021-04-29
feat: public nft checker api

alanshaw:
This PR adds a new API `GET /check/{cid}` which allows unauthenticated users to check the IPFS pin status and Filecoin deal status of _any_ NFT that has been added to nft.storage.


migrations/deals-list.js
```diff
@@ -0,0 +1,31 @@
...
+    if (env === 'production') {
+      return !t.title.includes('dev') && !t.title.includes('staging')
+    }
+    return t.table.title.includes(env)
+  })
+
+  let i = 0
+  for await (const keys of cf.fetchKVKeys(table.id)) {
+    i += keys.length
+    keys.forEach((k) => console.log(`${k.name}: ${JSON.stringify(k.metadata)}`))
```
Manik-Jain, line 26: I think we should avoid using console log in PROD. Rather we can log to a file.

alanshaw: Ah thank you but I did not mean for this to be committed - removed now.

site/src/routes/nfts-check.js
```diff
@@ -0,0 +1,34 @@
...
+
+  /** @type {import('@nftstorage/ipfs-cluster').StatusResponse} */
+  let status
+  /** @type {Deal[]} */
+  let deals
+  try {
+    ;[status, deals] = await Promise.all([cluster.status(cid), getDeals(cid)])
+  } catch (err) {
+    if (err.response && err.response.status === 404) {
+      return HTTPError.respond(new HTTPError('not found', 404))
```
hugomrdias, line 25:
```suggestion
     throw new HTTPError('not found', 404)
```
the error handler already does the `HTTPError.respond` bit

client/test/lib.spec.js
```diff
@@ -255,4 +255,39 @@ describe('client', () => {
...
+    const client = new NFTStorage({ token, endpoint })
+
+    /** @type {string[]} */
+    let preloaded
+    beforeEach(async () => {
+      preloaded = [
+        // QmVRC63ZHwHQBC4pkqbiQjPjdHVKueafTpwmkNHhnVfwLQ
+        await client.storeBlob(new Blob(['preload check'])),
+        // QmTPFUEcZvqKBYqJM3itqkDiqJaApYzLJ1ht6iBD4d6M28
+        // await client.storeBlob(new Blob(['missing']))
```
hugomrdias, line 269: do we need these comments ?

hugomrdias (changes requested):

client/test/lib.spec.js
alanshaw: Probably not, copied from above!

hugomrdias (approved):

client/test/service.js
```diff
@@ -53,7 +53,7 @@ export const init = (
 export const handle = async (request, { store, AUTH_TOKEN }) => {
   const url = new URL(request.url)
 
-  const [_, param] = url.pathname.split('/')
+  const [_, ...pathParts] = url.pathname.split('/')
```
Gozala, line 56: Nit: I think it would be easier to read later if it was

```suggestion
  const [_, base, param, ...etc] = url.pathname.split('/')
```

site/src/routes/nfts-check.js
```diff
@@ -0,0 +1,34 @@
...
+export const check = async (event, params) => {
+  const { cid } = params
+
+  /** @type {import('@nftstorage/ipfs-cluster').StatusResponse} */
+  let status
+  /** @type {Deal[]} */
+  let deals
+  try {
+    ;[status, deals] = await Promise.all([cluster.status(cid), getDeals(cid)])
+  } catch (err) {
```
Gozala, line 23: Aside: seems like something router should be doing this.

Gozala (approved):
Looks good to me, only thing I'm wondering about is: This looks pretty much what `client.status`, only differences are (as far as I can tell)

1. `client.status` requires auth.
2. `client.status` includes `size`
3. `client.status` includes `created` timestamp.

I do not see a reason why we could not expose `size` without auth & I'm not entirely sure I see a reason why `created` timestamp needs to be a secret either. And if both are true we could probably make `client.status` not require `auth`. It would probably need switch to use this implementation instead, but smaller API surface seems like a win.

client/test/service.js
alanshaw: I kinda agree but I didn't do it like that because the `param` is not always the 3rd item - e.g. `/{cid}` (status - 1) and `/check/{cid}` (check - 2).

site/src/routes/nfts-check.js
alanshaw: > I do not see a reason why we could not expose `size` without auth & I'm not entirely sure I see a reason why `created` timestamp needs to be a secret either. And if both are true we could probably make `client.status` not require `auth`. It would probably need switch to use this implementation instead, but smaller API surface seems like a win.

Just surfacing our convo:

We respond with more data in the HTTP API than we return from the JS client https://github.com/ipfs-shipyard/nft.storage/blob/4dfcb5b3537c0a7f3628d70c8ca97c7cda2d93f3/website/public/schema.yml#L147-L174 

for example `scope` is the name of the token and can be considered private.

Also, for things pinned through the pinning service API it'll carry `name` and `meta` in the `pin` object which is user specific.

So it would be a breaking change to remove that data.

In a way it's good because we can support the concept of "user private" data in the "status API" and have public stuff in the "check API"...

alanshaw:
> I do not see a reason why we could not expose `size` without auth & I'm not entirely sure I see a reason why `created` timestamp needs to be a secret either. And if both are true we could probably make `client.status` not require `auth`. It would probably need switch to use this implementation instead, but smaller API surface seems like a win.

Just surfacing our convo:

We respond with more data in the HTTP API than we return from the JS client https://github.com/ipfs-shipyard/nft.storage/blob/4dfcb5b3537c0a7f3628d70c8ca97c7cda2d93f3/website/public/schema.yml#L147-L174 

for example `scope` is the name of the token and can be considered private.

Also, for things pinned through the pinning service API it'll carry `name` and `meta` in the `pin` object which is user specific.

So it would be a breaking change to remove that data.

In a way it's good because we can support the concept of "user private" data in the "status API" and have public stuff in the "check API"...

PR #86, opened 2021-04-30
fix: faq typos

hugomrdias:


alanshaw:
I think provably is intentional. "probably permanent" is like an oxymoron 🤣

alanshaw (approved):

PR #97, opened 2021-05-05
feat: pin tracking

alanshaw:
# Problems:

1. We currently have a cron job that updates pin status and size for every NFT. It has to crawl every single NFT in the table and eventually it will be too big and hit API limits.
2. Every call to `/check/{cid}` results in a call to IPFS Cluster.
3. Our total data size metric is incorrect because it double counts two NFTs that have uploaded the same content.
4. Pinning Services API pins listing very inefficiently crawls the unordered list of NFTS for a given user and orders them on the  fly.

# Solutions:

1. We store every pin that needs to be followed up on in a separate KV so we only need to iterate over pins that need to be followed up on and remove the ones that are done.
2. We store pins (their size and status) in a separate KV store that can be easily queried by CID and do not have to talk to IPFS Cluster.
3. The separate KV store for pins keyed by CID means we won't double count when two NFTs are pinning the same file.
4. The metadata previously stored in the NFTS KV store has moved to the NFTS_IDX store. NFTS_IDX is date ordered per user so allows for significantly more efficient listing of NFT data.

## Additions in this PR

* Two new KV namespaces: PINS and FOLLOWUPS

    **PINS**: CIDs that are pending or pinned to the cluster
    key: `cid`
    value: `empty`
    metadata: `{ status: PinStatus, size: number, created: string }`

    **FOLLOWUPS**: CIDs that are being pinned to the cluster that need to be re-checked for completion/failure
    key: `cid`
    value: `empty`
    metadata: `{ status: PinStatus, size: number, created: string }`

* Migrations to populate these new KVs.

* New cron job to process items in the **FOLLOWUPS** KV

* New cron job to keep **NFTS_IDX** metadata up to date

## Changes in this PR

* Records in the **NFTS** KV no longer have a `size` property and their `pin` property is optional and may only contain optional `name: string` and `meta: Record<string, string>` values. When returning an NFT from the API we "inflate" the `size` and `pin` properties by fetching the data from **PINS**.

* Creating a new pin will cause a record to be added to **FOLLOWUPS** if the CID has not already been pinned.

* Records in the **NFTS** KV no longer have metadata - it has moved to **NFTS_IDX**.

* The Pinning Services API pin listing (`GET /pins`) now uses **NFTS_IDX**.

### DIFF
--- a/site/src/models/nfts.js
+++ b/site/src/models/nfts.js
@@ -1,5 +1,4 @@
 import { stores } from '../constants.js'
-import merge from 'merge-options'
 import * as nftsIndex from './nfts-index.js'
 
 /**
@@ -22,21 +21,23 @@ export const has = async (key) => {
 
 /**
  * @param {Key} key
- * @param {NFT} value
- * @param {any} [options]
+ * @param {NFT} nft
+ * @param {import('./pins.js').Pin} pin
  * @returns {Promise<NFT>}
  */
-export const set = async (key, value, options) => {
-  const savedValue = await get(key)
-  if (savedValue === null) {
-    const kvKey = encodeKey(key)
-    await stores.nfts.put(kvKey, JSON.stringify(value), options)
-    await nftsIndex.set({ ...key, created: value.created }, { key: kvKey })
-    return value
-  }
-  const data = merge(savedValue, value)
-  await stores.nfts.put(encodeKey(key), JSON.stringify(data), options)
-  return data
+export const set = async (key, nft, pin) => {
+  const kvKey = encodeKey(key)
+  await stores.nfts.put(kvKey, JSON.stringify(nft))
+  await nftsIndex.set(
+    { ...key, created: nft.created },
+    {
+      key: kvKey,
+      ...(nft.pin || {}), // name/meta
+      pinStatus: pin.status,
+      size: pin.size,
+    }
+  )
+  return nft
 }
 
 /**
@@ -65,7 +66,7 @@ export const remove = async (key) => {
 /**
  * @param {any} prefix
  * @param {{ limit?: number, before?: Date }} [options]
- * @returns {Promise<Array<KVValue<NFT>>>}
+ * @returns {Promise<Array<import('../bindings').NFTResponse|null>>}
  */
 export async function list(prefix, options) {
   options = options || {}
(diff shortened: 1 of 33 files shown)


site/src/models/nfts.js
```diff
@@ -109,32 +81,37 @@ export async function list(prefix, options) {
...
+            ...nft,
+            size: indexData.size,
+            pin: {
+              cid: nft.cid,
+              ...(nft.pin || {}),
+              status: indexData.pinStatus,
+              size: indexData.size,
+              created: nft.created,
+            },
+            deals: [], // FIXME: how to get deals data?
```
hugomrdias, line 112: we cant get from the deals KV ?

site/src/routes/nfts-get.js
```diff
@@ -16,14 +17,27 @@ export const status = async (event, params) => {
...
-  return new JSONResponse({ ok: true, value: { ...nft, deals } })
+  if (pin == null) {
+    return HTTPError.respond(new HTTPError('pin not found', 404))
+  }
+
+  /** @type {import('../bindings').NFTResponse} */
+  const res = {
+    ...nft,
+    size: pin.size,
+    pin: { cid: nft.cid, ...(nft.pin || {}), ...pin },
```
hugomrdias, line 38: can't we simplify this pin structure everywhere ? avoid  `{ cid: nft.cid, ...(nft.pin || {}), ...pin }` and just have everything in the output from getPin ?

hugomrdias (approved):
LGTM, just a couple of comments inline

hsanjuan:
As a side note: you can store arbitrary pin metadata directly in cluster. I'm not sure if this is known.

site/src/models/nfts.js
alanshaw: Yes we could, but we risk hitting API limits...this endpoint currently does not return this information so I was punting the solution until I had a better idea!

alanshaw:
> As a side note: you can store arbitrary pin metadata directly in cluster. I'm not sure if this is known.

Is known, and we cache the size in meta if known. It's possible multiple users upload the same file so we can't put user provided meta here.

site/src/routes/nfts-get.js
alanshaw: We could include the `cid` but `nft.pin` is user provided pin `name` and `meta` so we'd still have to merge it here.

alanshaw: I will add the `cid` because that will help a little!

site/src/routes/pins-replace.js
```diff
@@ -72,33 +73,19 @@ export async function pinsReplace(event, params) {
...
-    pin = await cluster.allocation(pinData.cid)
-  } catch (err) {
-    // if 404 we can continue, we have not pinned this CID
-    if (!err.response || err.response.status !== 404) {
-      throw err
-    }
-  }
-
+  const created = new Date().toISOString()
+  let pin = await pins.get(pinData.cid)
```
Gozala, line 77: This looks similar to what pins-add does and suggested refactor there would also enable code reuse.

Gozala (approved):
Looks good! I think `./migration/tools` are super useful maybe the should just be `./tools` a they seem useful beyond just migrations.  I have also suggested some **optional** improvements.

site/src/models/nfts.js
hugomrdias: KV calls limit ?

alanshaw: Seems to be undocumented but definitely exists.

PR #106, opened 2021-05-10
test: site tests

### LINKED ISSUE #96: Create test suite for ./site and run it in CI on commits / pulls

alanshaw:
Adds e2e tests for the API, using playwright.

Eventually the following repeated lines in tests will be replaced:

```js
    const request = new Request(`http://localhost/check/${cid}`)
    const event = new TestFetchEvent('fetch', { request })
    globalThis.dispatchEvent(event)
    const res = await event.respondWithPromise
```

...with a call to `fetch` when playwright supports service workers (currently only regular workers):

```js
    const res = await fetch(`http://localhost/check/${cid}`)
```

Later, these same tests can be run against a real life deployed API on Cloudflare.

resolves #96


hugomrdias (approved):
LGTM 5 ⭐️

PR #108, opened 2021-05-11
feat: metadata API examples

alanshaw:
...also a small fix in the client to parse the date in the `Pin` object returned by `storage.status()` to match the type information.


Gozala (approved):
Thanks you

PR #113, opened 2021-05-12
rejects empty directory

Gozala:
This is pull against deployed version that will reject uploads with no payloads


Gozala:
Given that new changes had been deployed this is probably no longer relevant and we can just do #114

PR #114, opened 2021-05-12
fix: reject empty uploads both on client and site

Gozala:


site/src/cluster.js
```diff
@@ -28,6 +29,12 @@ export async function add(data) {
  */
 export async function addDirectory(files) {
   const size = files.reduce((total, f) => total + f.size, 0)
+  if (size === 0) {
+    throw new HTTPError(
+      `Content added contains ${size} bytes. Please make sure that files are encoded correctly`
+    )
```
alanshaw, line 35:
```suggestion
    throw new HTTPError('Content added contains 0 bytes. Please make sure that files are encoded correctly')
```

alanshaw (approved):
LGTM

eazydripz:
Thank you

PR #117, opened 2021-05-13
test: in service worker

### LINKED ISSUE #112: Add sw support and fix issues with playwright-test
fix: fix race in uvu runner, reported by @Gozala 
https://github.com/hugomrdias/playwright-test/pull/205

fix: use tape instead of fresh-tape, this was causing CI failures when doing yarn install
https://github.com/hugomrdias/playwright-test/pull/206

feat: add support for service worker so with can have a better test env
https://github.com/hugomrdias/playwright-test/pull/211

alanshaw:
Uses new service worker test mode in playwright.

resolves  #112


site/src/utils/router.js
```diff
@@ -160,7 +160,15 @@ class Router {
    * @param {FetchEvent} event
    */
   listen(event) {
-    event.respondWith(this.route(event))
+    // when testing just respond to the fake testing domain
+    if (process.env.PW_TEST) {
+      const url = new URL(event.request.url)
+      if (url.hostname === 'testing.com') {
```
alanshaw, line 166: @hugomrdias we should probably use `testing.nft.storage`

```suggestion
      if (url.hostname === 'testing.nft.storage') {
```

site/test/version.spec.js
```diff
@@ -1,13 +1,11 @@
 import assert from 'assert'
-import '../src/index.js'
-import { TestFetchEvent } from './scripts/events.js'
+import { VERSION, COMMITHASH, BRANCH } from './scripts/worker-globals.js'
+
+const endpoint = 'http://testing.com'
```
alanshaw, line 4:
```suggestion
const endpoint = 'http://testing.nft.storage'
```

site/test/nfts-check.spec.js
```diff
@@ -1,9 +1,8 @@
 import assert from 'assert'
-import '../src/index.js'
-import { TestFetchEvent } from './scripts/events.js'
-import { set as setPin } from '../src/models/pins.js'
+import stores from './scripts/stores.js'
 import { clearStores } from './scripts/helpers.js'
 
+const endpoint = 'http://testing.com'
```
alanshaw, line 5:
```suggestion
const endpoint = 'http://testing.nft.storage'
```

PR #145, opened 2021-05-21
feat: pinata pin sync

alanshaw:
1. A KV store (`PINATA_QUEUE`) that new CIDs for pinning to pinata are placed in.
1. A cron job that reads from the queue and submits to Pinata.
1. A migration that adds all existing pinned CIDs to the queue to ensure they are pinned remotely.
1. Code changes to add to the `PINATA_QUEUE` instead of calling `pinata.pinByHash`


alanshaw:
I'm going to merge this because we _really_ need to start ensuring our pins are backed up to Pinata. I have run the migration so the queue is ready to be processed by the cron job.

PR #162, opened 2021-05-27
chore: avoid deprecated url-regex-safe version

### LINKED ISSUE #160: root yarn install fails on node 16
The `re2` dep fails to build under node 16, looks like a fix landed recently https://github.com/uhop/node-re2/issues/100

```console
~/Code/ipfs-shipyard/nft.storage on main  
❯ yarn 
...snip...
In file included from /Users/oli/Library/Caches/node-gyp/16.1.0/include/node/v8.h:30:
/Users/oli/Library/Caches/node-gyp/16.1.0/include/node/v8-internal.h:452:38: error: no template named 'remove_cv_t' in namespace 'std'; did you mean 'remove_cv'?
            !std::is_same<Data, std::remove_cv_t<T>>::value>::Perform(data);
                                ~~~~~^~~~~~~~~~~
                                     remove_cv
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1/type_traits:776:50: note: 'remove_cv' declared here
template <class _Tp> struct _LIBCPP_TEMPLATE_VIS remove_cv
                                                 ^
1 error generated.
make: *** [Release/obj.target/re2/lib/addon.o] Error 1
gyp ERR! build error 
```

olizilla:
- force yarn to install url-regex-safe@^2.0.2 to avoid pulling in a version of re2 that won't compile on node 16
- avoid deprecated url-regex-safe@1.1.0 which was previously pinned

From local testing, the problem is caused by yarn's dependency resolution pulling in deprecated modules. The direct dep `get-urls` builds fine on node 16 unchaged under npm, but fails under yarn, as npm does not included deprecated versions in its resolution.

`<grumble>`why are we using yarn? this always happens.`</grumble>`

Fixes #160 

License: MIT

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,5 +34,8 @@
     "rimraf": "^3.0.2",
     "simple-git-hooks": "^2.3.1",
     "typescript": "^4.2.4"
+  },
+  "resolutions": {
+    "url-regex-safe": "^2.0.2"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -450,6 +450,14 @@
     "@nodelib/fs.scandir" "2.1.4"
     fastq "^1.6.0"
 
+"@npmcli/move-file@^1.0.1":
+  version "1.1.2"
+  resolved "https://registry.yarnpkg.com/@npmcli/move-file/-/move-file-1.1.2.tgz#1a82c3e372f7cae9253eb66d72543d6b8685c674"
+  integrity sha512-1SUf/Cg2GzGDyaf15aR9St9TWlb+XvbZXWpDx8YKs7MLzMH/BCeopv+y9vzrzgkfykCGuWOlSu3mZhj2+FQcrg==
+  dependencies:
+    mkdirp "^1.0.4"
+    rimraf "^3.0.2"
+
 "@octokit/auth-token@^2.4.4":
   version "2.4.5"
   resolved "https://registry.yarnpkg.com/@octokit/auth-token/-/auth-token-2.4.5.tgz#568ccfb8cb46f36441fac094ce34f7a875b197f3"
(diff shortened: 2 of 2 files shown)


olizilla:
I can just stick with node 15 and we can ignore this. But it'll save us some pain in the long run to stick with a single version of node for all the things. From the ci workflows I can see we're using nodes 14, 15 and 16 depending on the project. Shall we go all in on node 16? (except for the client, tesing on 14 & 16 is a good idea there) ONLY FORWARDS!

olizilla:
i'm new here, so i'm not really proposing we change everything, I'm just flagging a frustration I'm having as a new dev to the project. sorry for the grumbles.

vasco-santos (commented):
I can confirm this fixes the build using Node 16. Can we update it in CI or have both 14 and 16?

hugomrdias:
so npm ignores a deprecated version even if it means jumping to a new major ?

hugomrdias:
> I can just stick with node 15 and we can ignore this. But it'll save us some pain in the long run to stick with a single version of node for all the things. From the ci workflows I can see we're using nodes 14, 15 and 16 depending on the project. Shall we go all in on node 16? (except for the client, tesing on 14 & 16 is a good idea there) ONLY FORWARDS!

100% feel free to go all in on 16 except for the client which should test 14 and 16 (kill 15 with fire)

olizilla:
> so npm ignores a deprecated version even if it means jumping to a new major ?

Much more boring! it uses the highest semver that remains compatible with the specfied range, so in this case npm installs `url-regex-safe@1.0.2` which is the last, non-deprecated, version compatible with the [get-urls package.json](https://github.com/sindresorhus/get-urls/blob/db3f04ae49bb4487a8472f5562e0d09d1952796e/package.json#L35) meanwhile, yarn is installing the highest available semver compatible version `url-regex-safe@1.10.0` [ignoring that it is deprecated](https://www.npmjs.com/package/url-regex-safe/v/1.1.0)

see: https://github.com/ipfs-shipyard/nft.storage/issues/160#issuecomment-849631511

hugomrdias:
wait what ? `url-regex-safe@1.0.2` works in node 16  but `url-regex-safe@1.10.0` does not ?

olizilla:
It appears so. It's a fun game right!? This is why i stick to just one module resolvers set of exciting bug/features. It's the transitive dep on re2 that fails to build under node 16 and there were some shinanigans in the package.json history for `url-regex-safe` where the older 1.0.2 version has a more permissive range dep on re2, allowing it to pull in the more recent version, while the deprecated 1.10.0 version pins re2 to exactly the bad unbuildable version. 🎉

hugomrdias (approved):
lets get this in and make a PR to `get-urls` to see if sindre updates if not we can use `url-regex-safe` directly

PR #163, opened 2021-05-27
test: add test for uploading a single file

alanshaw:
* Adds a mock cluster server using [smoke](https://npm.im/smoke)
* Adds a test that uploads a file, ensures the response is correct and checks it was added to the NFTS and PINS stores
* Minor changes to the JWT utility module to allow it to be used in tests

### DIFF
--- a/packages/api/test/mocks/cluster/post_add.js
+++ b/packages/api/test/mocks/cluster/post_add.js
@@ -1,6 +1,11 @@
+// @ts-ignore
 const Hash = require('ipfs-only-hash')
 
-// https://github.com/sinedied/smoke#javascript-mocks
+/**
+ * https://github.com/sinedied/smoke#javascript-mocks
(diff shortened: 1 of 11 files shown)


packages/api/test/mocks/cluster/post_add.js
```diff
@@ -0,0 +1,22 @@
+// @ts-ignore
```
vasco-santos, line 1: nit: usually nice to add the reason of the ignore next to this

vasco-santos (approved):
This looks good!
Left a nit comment and what about `packages/cron/src/bin/pinata.js`? It seems empty, should we delete the file?

hugomrdias (approved):
LGTM

alanshaw:
> Left a nit comment and what about `packages/cron/src/bin/pinata.js`?

Not empty? https://github.com/ipfs-shipyard/nft.storage/blob/main/packages/cron/src/bin/pinata.js

PR #176, opened 2021-06-07
chore: ensure website uses local client

olizilla:
get yarn workspace to symlink in nft.storage from the local workspace by making the dependency version range match.

License: MIT


vasco-santos (approved):
LGTM

alanshaw (approved):

PR #178, opened 2021-06-08
feat: hande car uploads

### LINKED ISSUE #172: Support CAR uploads 🚘✨⤴️
[Content Addressed Archives (CAR)](https://github.com/ipld/specs/blob/master/block-layer/content-addressable-archives.md) package up blocks with their CIDs. Users could pre-compute the root CID for their assets by creating the CAR file locally and uploading that. nft.storage would store the exact set of blocks provided in the CAR, and the single root CID listed in the CAR header would be used as the asset CID. 

This removes the need to trust nft.storage. The client chunks and hashes the file to derive the CID, nft.storage just stores it as provided. A local IPFS node can be used to create a CAR via `ipfs dag export <root CID>` or tools like [`ipfs-car`](https://github.com/vasco-santos/ipfs-car) or the web based https://car.on.fleek.co

It also allows users to begin minting contracts with the CID they created and upload the CAR file to nft.storage in parallel. They would be responsible for preserving a copy of the CAR until the upload to nft.storage is confirmed, so it can be retried in the event of an error, but the guarantee of content addressing means that they can safely use the CID they created in a contract, and that same CID will be available over IPFS once the CAR uploads and import completes.

## TODO

- [x] add car upload test to [packages/api/test/nfts-upload.spec.js](https://github.com/ipfs-shipyard/nft.storage/blob/main/packages/api/test/nfts-upload.spec.js)
- [x] add api route `/uploadCar` or update `/upload` to check for a `Content-Type: application/car` header (or other flag, TBC)
	- [ ] verify the user provided CAR has a single root in the header.
	- [ ] (TBC) verify the CAR is complete? I assume cluster will try and find any missing blocks, but we probably don't want that behaviour, or I could upload a CAR with a single block for the root of wikipedia, and have cluster pull in a huge DAG.
	- [x] ensure the car blob passed to cluster.add has the content-type `application/car` so it is added to ipfs-cluster as a CAR rather than a regular file.
	- [ ] verify the [cid returned from cluster](https://github.com/ipfs-shipyard/nft.storage/blob/ff032f1928b2493d3e040f971732f81879489137/packages/api/src/routes/nfts-upload.js#L52) matches the root cid in the user provided CAR 
	- [x] pin the car file root to cluster if not already pinned. (TBC Does cluster do this automatically with a car file?)
	- [x] check if the user has previously uploaded this CID before
	- [x] if not, persist it via the nft model and add the CID to the pinataQueue.
	- [x] figure out how to update the openapi spec for the api as necessary [openapi schema](https://github.com/ipfs-shipyard/nft.storage/blob/main/packages/website/public/schema.yml)
	- [ ] regenerate the openapi clients if needed.
- [x] update js client `storeBlob` method to verify car file, or add a `storeCar` method
	- [ ] add client test for car files
	- [x] [example added](https://github.com/ipfs-shipyard/nft.storage/tree/main/packages/client/examples) to the JS client
- [ ] update docs as needed
- [x] update website to explain how to create an upload a CAR
- [x] nice to have... add a checkbox to file upload to say "this is a CAR", otherwise you end up with a CID for the CAR file itself.

olizilla:
A first pass at handling CAR uploads, smaller than 100MB. (no car-chunking here). The aim was the smallest change that woud allow users to create and upload their own CARs and have nft.storage handle it correctly, to unblock users who want to pre-compute the root CID for their files. Turns out it was almost supported already. Good work Spark!

- api: check for Content-Type: application/car and handle cluster size vs bytes difference and test it.
- api: add a `smoke` mock for posting small car files to cluster.
- client: add isCar flag to storeBlob to force application/car content type on upload.
- website: add is CAR? check box to file upload form to try it out, and an explainer.
- website: update openAPI schema

Fixes #172 

| upload form | details expanded  |
|-------------|-------------------|
| Screenshot 2021-06-07 at 14 34 39 |  Screenshot 2021-06-07 at 14 34 50

License: MIT


olizilla:
I will fix the CI complaints tomorrow.

packages/api/src/routes/nfts-upload.js
```diff
@@ -45,19 +45,27 @@ export async function upload(event, ctx) {
...
       throw new HTTPError('Empty payload', 400)
     }
-    const { cid, size } = await cluster.add(blob)
+    if (
+      contentType.includes('application/car') &&
+      blob.type !== 'application/car'
+    ) {
+      // force the content type on the blob without duplicating content
+      blob = blob.slice(0, blob.size, 'application/car')
+    }
```
olizilla, line 58: This may not strictly be needed, it is a left over from when I was struggling to get the content-type:application/car to show up in the worker when i set it on the upload... I'm not clear on if the blob as provided in the worker context is guaranteed to always have it's content-type set to match the content-type header on the POST that got it there.

olizilla: ...and it's vital for us that it does here as it's the only flag to the cluster client to treat the add as a CAR add.

olizilla:
oh my. i will fix the rest of the CI complaints tomorrow.

packages/api/src/routes/nfts-upload.js
```diff
@@ -45,19 +45,27 @@ export async function upload(event, ctx) {
...
+    const { cid, size, bytes } = await cluster.add(blob)
     nft = {
       cid,
       created,
       type: blob.type,
       scope: tokenName,
       files: [],
     }
-    nftSize = size
+    nftSize = size || bytes
```
vasco-santos, line 68: This could lead into misleading size for the NFT, if we do not provide all the blocks, right?

packages/client/examples/node.js/storeCar.js
```diff
@@ -0,0 +1,38 @@
...
+import { Blob } from '../../src/platform.js'
+
+const endpoint = 'https://api.nft.storage' // the default
+const token = 'API_KEY' // your API key from https://nft.storage/manage
+
+async function main() {
+  const storage = new NFTStorage({ endpoint, token })
+
+  // locally chunk'n'hash the file to get the CID and pack the blocks in to a CAR
+  const { root, out } = await pack({
```
vasco-santos, line 13: We are complicating things here because of the test setup with Node, given that `ipfs-car` already offers a `packToBlob` function (for browser). Wondering if we should have also a browser examples folder?

packages/client/src/lib.js
```diff
@@ -70,18 +70,26 @@ class NFTStorage {
   /**
    * @param {API.Service} service
    * @param {Blob} blob
+   * @param {boolean} [isCar=false]
    * @returns {Promise<API.CIDString>}
    */
-  static async storeBlob({ endpoint, token }, blob) {
+  static async storeBlob({ endpoint, token }, blob, isCar = false) {
```
vasco-santos, line 76: Why not having a new function to `storeCar`?
Specially has the logic will increase for chunking, I feel like we should add a new function, rather than having users starting to use this function with a CAR and then need to change.
Also, as we will only support larger uploads with CAR files, we might reach a point that we want to deprecate (not recommend) storing Blobs?

packages/api/test/scripts/car.js
```diff
@@ -0,0 +1,17 @@
+import * as Block from 'multiformats/block'
+import * as Raw from 'multiformats/codecs/raw'
+import { sha256 } from 'multiformats/hashes/sha2'
+import * as CAR from '../../src/utils/car.js'
+
+// @ts-ignore
```
vasco-santos, line 6: nit: I usually like to add the reasons for adding ts-ignore next to it. This helps us cleaning them up in the future. I don't know the exact pattern in this project though

packages/api/src/routes/nfts-upload.js
olizilla: Yes, if the same code path is used for uploading partial CARs then the bytes value would likely be inaccruate

packages/client/examples/node.js/storeCar.js
olizilla: there is a browser examples folder! I can add a browser example once this PR is generally acceptable.

packages/client/src/lib.js
olizilla: Yes, that's a good idea. I was going to create a storeCar function, as an alternative, but then got bogged down in figuring out all the various testing shims and typescript complaints. I went with the boolean flag initial as the original plan was to punt on having the client create CAR files itself, and I wanted to keep the change as small as possible... but boolean params are horrible `storeBlob(foobar, true)` is too mysterious.

packages/api/test/nfts-upload.spec.js
```diff
@@ -45,18 +53,61 @@ describe('/upload', () => {
...
+  })
+
+  it('should upload a single CAR file', async () => {
+    const { token, issuer } = await createTestUser()
+
+    const { root, car } = await createCar('hello world!')
+    // expected CID for the above data
+    const cid = 'bafkreidvbhs33ighmljlvr7zbv2ywwzcmp5adtf4kqvlly67cy56bdtmve'
+    assert.strictEqual(root.toString(), cid, 'car file has correct root')
+    const res = await fetch(new URL(`upload`, endpoint).toString(), {
```
alanshaw, line 82:
```suggestion
    const res = await fetch(new URL('upload', endpoint).toString(), {
```

packages/api/test/scripts/car.js
alanshaw: I did not realise you were allowed to add a message so I haven't been doing it so far. I agree a message is useful. does it just need:

```suggestion
/**
 * @param {string} str Data to encode into CAR file.
 */
```

packages/client/examples/node.js/storeCar.js
alanshaw: @vasco-santos can has `packToBlob` for Node.js using `@web-std/blob`?

packages/client/examples/node.js/storeCar.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const carParts = []
+  for await (const part of out) {
+    carParts.push(part)
+  }
+  const car = new Blob(carParts, { type: 'application/car' })
+
+  // send the CAR to nft.storage, setting isCar to true
+  const isCar = true
+  const cid = await storage.storeBlob(car, isCar)
```
alanshaw, line 28: Why setting the blob `type` as well as `isCar`? Can we just use blob `type`?

packages/client/src/lib.js
alanshaw: I think it's fine to add support for CAR in `storeBlob` - we can release this ASAP to enable client side CID generation and then when we have chunked uploads we can have a separate API and client method for `storeBigCar`. WDYT?

I agree the boolean param is horrible - can we just use the `type` property of the blob?

packages/client/test/importer.js
```diff
@@ -67,6 +68,18 @@ export const importBlob = async (content) => {
   throw new Error(`Import failed`)
 }
 
+/**
+ * @param {Uint8Array} content
+ */
+export const importCar = async (content) => {
+  const car = await CarCIDIterator.fromBytes(content)
```
alanshaw, line 75: Use `CarReader` for simplicity? I was trying to figure out why you're using `CarCIDIterator` and not `CarReader`? I don't think there's a reason, right?

packages/website/pages/new-file.js
```diff
@@ -67,6 +68,36 @@ export default function NewFile() {
                 required
               />
             </div>
+            <label>
+              <input id="is-car" name="is-car" type="checkbox" />
+              <span className="ml2">is CAR?</span>
+            </label>
+            <details className="db mt3 mb4">
+              <summary className="i">Tell me more, what is CAR?</summary>
```
alanshaw, line 76:
```suggestion
              <summary className="i">Tell me more, what is a CAR?</summary>
```

packages/website/public/schema.yml
```diff
@@ -30,6 +30,10 @@ paths:
             schema:
               type: string
               format: binary
+          application/car:
+            schema:
+              type: string
+              format: binary
```
alanshaw, line 36: Please can you add something to the description to say that CAR files are supported and why?

packages/client/examples/node.js/storeCar.js
alanshaw: Yes please for a browser example 😄

packages/website/pages/new-file.js
```diff
@@ -67,6 +68,36 @@ export default function NewFile() {
...
+                </a>{' '}
+                or via{' '}
+                <a
+                  className="black"
+                  href="https://car.on.fleek.co/"
+                  target="_blank"
+                  rel="noreferrer"
+                >
+                  https://car.on.fleek.co
+                </a>
```
alanshaw, line 98:
```suggestion
                <a
                  className="black"
                  href="https://car.ipfs.io/"
                  target="_blank"
                  rel="noreferrer"
                >
                  https://car.ipfs.io
                </a>
```

alanshaw (changes requested):
❤️🚀

Is the "Is CAR" checkbox in the UI just to surface the ability for users to upload CAR files? 

It would be nice to _not_ have to check the box and instead have a "CAR files supported! What's a CAR?". Is it possible to just recognise `.car` extension or `application/car`?

Please can you update the comments in the client to explain this new feature so it appears in our [client library docs](https://ipfs-shipyard.github.io/nft.storage/client/classes/lib.nftstorage.html#storeblob)? 🙏: https://github.com/ipfs-shipyard/nft.storage/blob/ccdbe968995e2aeeb5dbbf62771bd5dd302c875c/packages/client/src/lib.js#L224-L241

packages/client/src/lib.js
olizilla: i would rather just use the `type` param of the blob, but it would require the user to re-encode their file-blob to set the content type correctly if they are accepting `<input type='file'>`.

on discussion with @alanshaw we are going to go with @vasco-santos's suggestion here and change the client to add a `storeCar` method, which his work can then upgrade with large car support.

packages/client/examples/node.js/storeCar.js
vasco-santos: > @vasco-santos can has packToBlob for Node.js using @web-std/blob?

@olizilla and me already talked about it. Having Node.js using @web-std/blob would mean we should be consistent and use it for the Browser as well. But then, we need to direct people to use @web-std/blob when they already know of Blob in the browser. What do you think @alanshaw ?

packages/client/test/importer.js
olizilla: `CarReader` reads the whole CAR into mem. I just need `getRoots`, so this does less work, but agree, it's distracting from the purpose of this mock.

packages/website/pages/new-file.js
olizilla: this was deliberate

packages/website/pages/new-file.js
olizilla: it does not exist!

olizilla:
TODO:
- [ ] make car.ipfs.io a thing
- [ ] move to a `storeCar` method in client
- [ ] add browser example for client
- [ ] add to openAPI description to say that CAR files are supported and why?
- [ ] switch mock importer to use CarReader
- [ ] update the comments in the client to explain this new feature so it appears in our client library docs
- [ ] provide ts-ignore reasons

packages/website/pages/new-file.js
alanshaw: It should!

packages/client/examples/browser/storeCar.html
```diff
@@ -0,0 +1,42 @@
+<!DOCTYPE html>
+<head>
+  <meta charset="utf-8" />
+</head>
+<pre id="out"></pre>
+<script type="module">
+  import { NFTStorage } from 'https://cdn.skypack.dev/nft.storage'
+  import { packToBlob } from 'https://cdn.skypack.dev/ipfs-car/pack/blob'
```
olizilla, line 8: am seeing the this get skypack'd to this:
```

'use strict'

const { encoding: getCodec } = require('multibase')
const utf8Decoder = new TextDecoder('utf8')
```
https://github.com/achingbrain/uint8arrays/blob/744af839e7cb3010a2182155822d8b56448a98c4/to-string.js#L1-L16


```
import require$$0 from "/-/multibase@v4.0.4-uhuAWZAoQBdPwI78HUyE/dist=es2020,mode=imports/optimized/multibase.js";
import require$$1 from "/-/web-encoding@v1.1.4-BoaEvv5UU9EtSHTQgnOW/dist=es2020,mode=imports/optimized/web-encoding.js";
const {encoding: getCodec} = require$$0;
const {TextDecoder} = require$$1;
const utf8Decoder = new TextDecoder("utf8");
```
https://cdn.skypack.dev/-/uint8arrays@v2.1.3-MNVirG8yMHFVvhSeyUgN/dist=es2020,mode=imports/unoptimized/to-string.js

olizilla: ah ha! different verison! the actual source of 2.1.3 uses the problematic `web-encoding` package

https://github.com/achingbrain/uint8arrays/compare/v2.1.3...v2.1.5

packages/client/examples/node.js/storeCar.js
alanshaw: hmm, why do you need to use `@web-std/blob` in the browser as well?

The nft.storage client only uses `@web-std/*` in Node.js:

browser: https://github.com/ipfs-shipyard/nft.storage/blob/main/packages/client/src/platform.web.js
node: https://github.com/ipfs-shipyard/nft.storage/blob/main/packages/client/src/platform.js

packages/client/examples/browser/storeCar.html
olizilla: uint8arrays@2.1.3 appears to be the troublesome version at https://github.com/achingbrain/uint8arrays/blob/bf09df0c161ef4d5a951675f0f720ea01832aef5/to-string.js#L4

Gozala (approved):

packages/client/src/lib.js
```diff
@@ -70,18 +70,26 @@ class NFTStorage {
...
     if (blob.size === 0) {
       throw new Error('Content size is 0, make sure to provide some content')
     }
 
+    let headers = NFTStorage.auth(token)
+    if (isCar) {
+      headers = Object.assign(headers, {
+        'Content-Type': 'application/car',
+      })
+    }
```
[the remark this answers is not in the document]
olizilla: as `/upload` enpoint is using the `content-type` to determine how to process the file, it seemed safer to ensure `application/car` was used here. If the user creates a CAR somewhere else and then adds it via `<input type="file">` they need to be able to indicate it should be treated as a CAR, even if the blob's content type is not `application/car`. So we either require that callers re-encode their blobs to set the content-type, or re-create the blob here if it's type is not car, or we just force the header here. What do you think @Gozala?

packages/api/src/routes/nfts-upload.js
```diff
@@ -45,19 +45,24 @@ export async function upload(event, ctx) {
     }
     nftSize = size
   } else {
-    const blob = await event.request.blob()
+    let blob = await event.request.blob()
```
vasco-santos, line 48:
```suggestion
    const blob = await event.request.blob()
```

packages/client/examples/node.js/storeCar.js
vasco-santos: Here it is: https://github.com/vasco-santos/ipfs-car/pull/25

packages/client/examples/node.js/storeCar.js
vasco-santos: needs to be updated to use `storeCar`

packages/client/src/lib.js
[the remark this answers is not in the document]
olizilla: I've gone ahead and made it re-encode the blob to force the type to be `application/car`

olizilla:
making https://car.ipfs.io a thing over here https://github.com/protocol/infra/pull/716

packages/client/src/lib.js
```diff
@@ -92,6 +92,18 @@ class NFTStorage {
...
   }
+  /**
+   * @param {API.Service} service
+   * @param {Blob} blob
+   * @returns {Promise<API.CIDString>}
+   */
+  static async storeCar({ endpoint, token }, blob) {
+    const car =
+      blob.type !== 'application/car'
+        ? blob.slice(0, blob.size, 'application/car') // TODO: test this in safari.
```
alanshaw, line 103: Seems to work

Screenshot 2021-06-15 at 16 10 41

alanshaw (approved):

packages/client/examples/browser/car/README.md
```diff
@@ -0,0 +1,16 @@
...
+```console.
+npm install
+npm run dev
+
+# or
+yarn
+yarn dev
+```
+
+Then visit `http://localhost:3000?key=<your nft.upload API KEY here>`
```
alanshaw, line 16:
```suggestion
Then visit `http://localhost:3000?key=<your nft.storage API KEY here>`
```

packages/client/examples/browser/car/README.md
```diff
@@ -0,0 +1,16 @@
+# CAR upload demo - nft.storage
+
+A demo using ipfs-car in the browser to create a CAR file and pre-calculate the CID for an asset then storing it on nft.storage and confirming that it uses the exact same CID for the asset.
+
+## Getting started
+
+```console.
```
alanshaw, line 7:
```suggestion
```console
```

alanshaw (changes requested):

alanshaw (approved):

packages/client/src/lib.js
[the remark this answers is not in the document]
olizilla: i've removed the code to explicitly set the request header, but we now have a weirdness where if a user provides a web-std Blob instead of a Blob (by what terrible accident they do that is a seperate issue, but is the case for ipfs-car bundled with vite at the moment) then the browser does not feel obliged to set the request header to match the type of a Blob-like thing.

PR #180, opened 2021-06-14
refactor: move NFTS_IDX cron job to GitHub actions 

alanshaw:
Last job to move over.


alanshaw:
I've been running this job locally, but we won't need it when https://github.com/ipfs-shipyard/nft.storage/issues/331 lands so I'm going to close this PR as it's not needed.

PR #189, opened 2021-06-17
chore: use node 16 in CI

olizilla:
update all workflows to use:
- actions/setup-node@v2 with `node-version: 16`
- actions/checkout@v2
- bahmutov/npm-install@v1

License: MIT

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -17,6 +17,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
       - uses: bahmutov/npm-install@v1
       - name: Typecheck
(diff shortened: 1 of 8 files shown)


.github/workflows/client.yml
```diff
@@ -33,23 +36,15 @@ jobs:
...
-      - name: Setup node
-        uses: actions/setup-node@v1
+      - uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node }}
-
-      - name: Install dependencies
-        uses: bahmutov/npm-install@v1
-
+      - name: bahmutov/npm-install@v1
```
alanshaw, line 42: Should this be `uses:`?
```suggestion
      - uses: bahmutov/npm-install@v1
```

alanshaw (changes requested):

.github/workflows/client.yml
olizilla: YES!

alanshaw (approved):

PR #190, opened 2021-06-17
fix: misc docs issues

alanshaw:
Screenshot 2021-06-17 at 15 30 49


olizilla (approved):
LGTM. 

Sorry I totally missed that. I wasn't clear on how it worked, so just guessed. Should have flagged it in the review.

PR #198, opened 2021-06-18
feat: deal data selector

### LINKED ISSUE #192: Surface deal data selector in client
The deal data selector is now available so if folks actually want to retrieve data from Filecoin they'll soon be able to use this (🔜 currently needs custom lotus client https://github.com/filecoin-project/lotus/pull/6393).

image

alanshaw:
The deal data selector is now available so if folks actually want to retrieve data from Filecoin they'll soon be able to use this (🔜 currently needs custom lotus client filecoin-project/lotus#6393).

@ribasushi can you confirm this is just for "active" and "terminated" deals?

resolves #192

### DIFF
--- a/packages/client/src/lib/interface.ts
+++ b/packages/client/src/lib/interface.ts
@@ -145,7 +145,10 @@ export interface FinalizedDeal extends DealInfo {
    * Identifier for the deal stored on chain.
    */
   chainDealID: number
-
+  /**
+   * Selector for extracting stored data from the batch root.
+   */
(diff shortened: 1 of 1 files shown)


ribasushi (approved):

ribasushi:
> @ribasushi can you confirm this is just for "active" and "terminated" deals?
Just for "active" really, we need to re-evaluate how we do activity tracking based on what we actually have
Unlikely today due to meetings, but early next week

PR #199, opened 2021-06-18
feat: car chunking client

### LINKED ISSUE #173: Increase 100MB upload limit
The 100MB upload limit currently exists because nft.storage is hosted on Cloudflare and this is the upload limit for workers.

We plan to increase (or maybe even remove) this limit asap.

In order to enable parallel uploading and minting of NFTs we're first going to [enable CAR file uploads](https://github.com/ipfs-shipyard/nft.storage/issues/172) and then do the work to support chunked CAR uploads (we currently have a MVP [library for CAR chunking](https://github.com/alanshaw/carbites)).

vasco-santos:
Adds car chunking in nft.storage client do allow chunking cars to avoid the current 100mb upload limit

Closes #173


packages/client/examples/browser/large-car/README.md
```diff
@@ -0,0 +1,16 @@
+# CAR upload demo - nft.storage
```
vasco-santos, line 1: Temporary for tests

packages/client/src/lib.js
```diff
@@ -94,15 +104,40 @@ class NFTStorage {
...
+      const blob = carReader
+      carReader = await CarReader.fromIterable(toIterable(blob.stream()))
+    }
+
+    // const roots = await carReader.getRoots()
+    // TODO: Should we throw if multiple roots?
+
+    const targetSize = 960 * 960 * 100 // chunk to ~96MB CARs
+    // const targetSize = 960 * 960 * 4
+    const splitter = new TreewalkCarSplitter(carReader, targetSize)
```
alanshaw, line 122:
```suggestion
    const splitter = carReader instanceof Blob
      ? await TreewalkCarSplitter.fromBlob(carReader, targetSize)
      : new TreewalkCarSplitter(carReader, targetSize)
```

packages/client/src/lib.js
```diff
@@ -94,15 +104,40 @@ class NFTStorage {
...
-    return NFTStorage.storeBlob({ endpoint, token }, car)
+  static async storeCar({ endpoint, token }, carReader) {
+    // See if this is a blob
+    if (carReader instanceof Blob) {
+      const blob = carReader
+      carReader = await CarReader.fromIterable(toIterable(blob.stream()))
+    }
+
+    // const roots = await carReader.getRoots()
+    // TODO: Should we throw if multiple roots?
```
alanshaw, line 118: Carbites throws if > 1 root. https://github.com/alanshaw/carbites/blob/main/lib/treewalk/splitter.js#L31

packages/client/src/lib.js
```diff
@@ -94,15 +104,40 @@ class NFTStorage {
...
+   * @param {Blob|CarReader} carReader
+   * @returns {Promise<API.CIDString | undefined>}
    */
-  static async storeCar({ endpoint, token }, blob) {
-    const car =
-      blob.type !== 'application/car'
-        ? blob.slice(0, blob.size, 'application/car') // TODO: test this in safari.
-        : blob
-    return NFTStorage.storeBlob({ endpoint, token }, car)
+  static async storeCar({ endpoint, token }, carReader) {
```
alanshaw, line 110: Option for `onProgress`?

packages/client/src/lib.js
```diff
@@ -94,15 +104,40 @@ class NFTStorage {
...
+
+    // const roots = await carReader.getRoots()
+    // TODO: Should we throw if multiple roots?
+
+    const targetSize = 960 * 960 * 100 // chunk to ~96MB CARs
+    // const targetSize = 960 * 960 * 4
+    const splitter = new TreewalkCarSplitter(carReader, targetSize)
+
+    const upload = parallelMap(
+      3,
```
alanshaw, line 125: `MAX_CONCURRENT_UPLOADS`

packages/client/src/lib.js
```diff
@@ -94,15 +104,40 @@ class NFTStorage {
...
+      async (/** @type {AsyncIterable<Uint8Array>} */ car) => {
+        const carParts = await all(car)
+        const carFile = new Blob(carParts, {
+          type: 'application/car',
+        })
+        console.log('store blob')
+        const res = await NFTStorage.storeBlob({ endpoint, token }, carFile)
```
alanshaw, line 132: `p-retry` probably a good idea

packages/client/src/lib.js
```diff
@@ -94,15 +104,40 @@ class NFTStorage {
...
+        console.log('stored blob')
+        return res
+      }
+    )
+
+    const res = await all(upload(splitter.cars()))
+
+    return res && res[0]
```
alanshaw, line 140: Lets not import another module for this!

```suggestion
  let cid
  for await (const res of upload(splitter.cars())) {
    cid = res
  }
  return cid
```

vasco-santos: Just pointing out that we need to do this both here and for the async iterator from the spiller itself. I removed the module for now, but can put it back if you reconsider

packages/client/src/lib.js
```diff
@@ -13,6 +13,16 @@
...
  */
+
+// @ts-ignore module with no types
+import toIterable from 'stream-to-it'
+// @ts-ignore module with no types
+import { parallelMap } from 'streaming-iterables'
+
+import all from 'it-all'
+import { CarReader } from '@ipld/car'
+import { TreewalkCarSplitter } from 'carbites/treewalk'
```
alanshaw, line 24: I think these are going to have to be imported from platform.js to do the right thing in the right env.

packages/client/src/lib.js
vasco-santos: This would be nice, but I am not sure we have enough data here. We do not know how many chunks will exist on the fly, nor the total number of bytes

packages/client/src/lib.js
vasco-santos: I am not sure I get it. As far as I can tell, platform web is used to get stuff from globalThis, which is not the case here

alanshaw: We have to import from skypack (or other) in the browser don't we?

vasco-santos: humm, don't we ship a bundle of the client ourselves?

@ipld/car and @carbites already ship browser compatible ESM code. If we bundle, minimize and so on, I do not see the reason to use a tool like skypack? But, I am not sure about other potential benefits

alanshaw: Yeah ignore me I think I'm just talking rubbish, the js CDNs deal with it for us.

packages/client/src/lib.js
```diff
@@ -13,11 +13,25 @@
...
+
 import * as API from './lib/interface.js'
 import * as Token from './token.js'
 import { fetch, File, Blob, FormData } from './platform.js'
 import { toGatewayURL } from './gateway.js'
 
+const MAX_STORE_RETRIES = 3
+const MAX_CONCURRENT_UPLOADS = 3
+const MAX_CHUNK_SIZE = 960 * 960 * 100 // chunk to ~96MB CARs
```
jonyg80, line 33: Maybe it should be 
```suggestion
const MAX_CHUNK_SIZE = 1024 * 1024 * 96 // chunk to ~96MB CARs
```
because written code is chunking to 87.89063 MB

vasco-santos: thanks, but this was changes for testing purposes

packages/client/src/lib.js
```diff
@@ -94,15 +108,48 @@ class NFTStorage {
...
    */
-  static async storeCar({ endpoint, token }, blob) {
-    const car =
-      blob.type !== 'application/car'
-        ? blob.slice(0, blob.size, 'application/car') // TODO: test this in safari.
-        : blob
-    return NFTStorage.storeBlob({ endpoint, token }, car)
+  static async storeCar(
+    { endpoint, token },
+    carReader,
```
alanshaw, line 117: Since it can be a `Blob` **or** a `CarReader`...

```suggestion
    car,
```

packages/client/src/lib.js
```diff
@@ -94,15 +108,48 @@ class NFTStorage {
   }
   /**
    * @param {API.Service} service
-   * @param {Blob} blob
-   * @returns {Promise<API.CIDString>}
+   * @param {Blob|CarReader} carReader
+   * @param {{onStoredChunk?: (cid: API.CIDString) => void}} [options]
```
alanshaw, line 112: The CID will always be the same so not really useful...

```suggestion
   * @param {{onStoredChunk?: (size: number) => void}} [options]
```

packages/client/src/lib.js
```diff
@@ -94,15 +108,48 @@ class NFTStorage {
...
+   * @returns {Promise<API.CIDString | undefined>}
```
alanshaw, line 113: CID or throw

```suggestion
   * @returns {Promise<API.CIDString>}
```

packages/client/src/lib.js
```diff
@@ -94,15 +108,48 @@ class NFTStorage {
...
+      }
+    )
+
+    // Keep the first cid
+    let root
+    for await (const res of upload(splitter.cars())) {
+      if (!root) {
+        root = res
+      }
+    }
```
alanshaw, line 151: You can keep the first or the last or any. You can drop the conditional and we should throw if no iteration to keep TS happy.

```suggestion
    let root
    for await (const cid of upload(splitter.cars())) {
      root = cid
    }
    if (!root) {
      throw new Error('no CAR data uploaded')
    }
```

packages/client/src/lib.js
```diff
@@ -265,17 +312,31 @@ class NFTStorage {
...
+   * ```
+   * import { packToBlob } from 'ipfs-car/pack/blob'
    * const data = 'Hello world'
    * const { root, car } = await packToBlob({ input: [new TextEncoder().encode(data)] })
    * const expectedCid = root.toString()
    * const cid = await client.storeCar(car)
    * console.assert(cid === expectedCid)
    * ```
-   * @param {Blob} blob
+   * @param {Blob|CarReader} carReader
```
alanshaw, line 335:
```suggestion
   * @param {Blob|CarReader} car
```

packages/client/src/lib.js
```diff
@@ -265,17 +312,31 @@ class NFTStorage {
...
+   * @param {{onStoredChunk?: (cid: API.CIDString) => void}} [options]
    */
-  storeCar(blob) {
-    return NFTStorage.storeCar(this, blob)
+  storeCar(carReader, options) {
+    return NFTStorage.storeCar(this, carReader, options)
```
alanshaw, line 339:
```suggestion
  storeCar(car, options) {
    return NFTStorage.storeCar(this, car, options)
```

alanshaw (changes requested):
Looking good. Pretty much all nits at this point 😉

packages/client/examples/node.js/storeCar.js
```diff
@@ -9,15 +10,17 @@ async function main() {
   const storage = new NFTStorage({ endpoint, token })
 
   // locally chunk'n'hash the file to get the CID and pack the blocks in to a CAR
-  const { root, car } = await packToBlob({
+  const { root, out } = await pack({
```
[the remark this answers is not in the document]
vasco-santos: The generic `pack` is more efficient because we do not need to consume the entire content to produce the Blob. The Blob is useful for the browser, but I would really prefer to point people to use the `CarReader`, as (specially if they look into `packToFs`) will require way less memory.

What do you think?

packages/client/src/lib.js
```diff
@@ -13,11 +13,25 @@
  * ```
  * @module
  */
+
+// @ts-ignore module with no types
+import toIterable from 'stream-to-it'
```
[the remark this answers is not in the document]
vasco-santos: good catch, shouldn't the linter catch this?

packages/client/examples/node.js/storeCar.js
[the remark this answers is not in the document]
alanshaw: Ah gotcha - the [`CarReader` class](https://github.com/ipld/js-car#carreader) reads the entire thing into memory. You need to use the indexed reader instead.

packages/client/src/lib.js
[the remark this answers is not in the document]
alanshaw: I would have thought so...did you setup the [git hook](https://github.com/ipfs-shipyard/nft.storage#getting-started)?

packages/client/examples/node.js/storeCar.js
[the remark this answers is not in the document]
vasco-santos: oh, I did not know that about the CarReader. I will change it to the indexed one

packages/client/examples/node.js/storeCar.js
```diff
@@ -9,15 +10,17 @@ async function main() {
...
     input: fs.createReadStream('pinpie.jpg'),
   })
-
   const expectedCid = root.toString()
   console.log({ expectedCid })
 
-  // send the CAR to nft.storage, setting isCar to true
-  const cid = await storage.storeCar(car)
+  // Create the car reader
+  const carReader = await CarIndexedReader.fromIterable(out)
```
alanshaw, line 20: No `fromIterable` on `CarIndexedReader` - you need to use `fromFile`.

packages/client/src/lib.js
```diff
@@ -13,11 +13,23 @@
...
+
+import { CarReader } from '@ipld/car'
+import { TreewalkCarSplitter } from 'carbites'
+
 import * as API from './lib/interface.js'
 import * as Token from './token.js'
 import { fetch, File, Blob, FormData } from './platform.js'
 import { toGatewayURL } from './gateway.js'
 
+const MAX_STORE_RETRIES = 3
```
alanshaw, line 29: We really want this to succeed...

```suggestion
const MAX_STORE_RETRIES = 5
```

alanshaw (changes requested):

packages/client/examples/node.js/storeCar.js
vasco-santos: yes, I tested my local storeLargeCar instead of this file when pushed 🤦🏼‍♂️

alanshaw (approved):

PR #203, opened 2021-06-22
fix: chunking issues

alanshaw:
* Uses `local` option for Cluster when file is >2.5MB (size recommended by @hsanjuan). This causes cluster to add blocks to the local node only (instead of sending to all cluster nodes) and relies on bitswap to efficiently distibute the blocks asyncronously. It allows the request to complete sooner and speeds up uploads significantly.
* Uses `transform` from `streaming-iterables` instead of `parallelMap`. The `parallelMap` function is actually a parallel batch, so if 1 of the 3 concurrent requests stalls and the other 2 complete then we have to wait before starting a new batch of 3.
* Does not assume pinned status after file upload to Cluster - when creating the pin, Cluster is checked for status (resolves https://github.com/ipfs-shipyard/nft.storage/issues/191)
* Changes `MAX_CHUNK_SIZE` to 10MB - this effectively ensures we get feedback quicker and keeps our requests light enough that if a failure occurs it's not too disasterous to retry.

### DIFF
--- a/packages/api/src/routes/nfts-upload.js
+++ b/packages/api/src/routes/nfts-upload.js
@@ -9,6 +9,7 @@ import { validate } from '../utils/auth.js'
 import { debug } from '../utils/debug.js'
 
 const log = debug('nfts-upload')
+const LOCAL_ADD_THRESHOLD = 1024 * 1024 * 2.5
 
 /**
  * @typedef {import('../bindings').NFT} NFT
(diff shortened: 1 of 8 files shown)


packages/api/src/routes/nfts-upload.js
```diff
@@ -66,8 +71,9 @@ export async function upload(event, ctx) {
   }
 
   let pin = await pins.get(nft.cid)
-  if (!pin || pin.status !== 'pinned') {
-    pin = { cid: nft.cid, status: 'pinned', size: nftSize, created }
+  if (!pin) {
+    const status = cluster.toPSAStatus(await cluster.status(nft.cid))
```
vasco-santos, line 75: at some point, we need to make the status as pinned

vasco-santos (commented):
Overall changes look good for chunking.
Just a question about the pin logic changes

packages/api/src/routes/nfts-upload.js
alanshaw: There is a cron job! https://github.com/ipfs-shipyard/nft.storage/blob/main/packages/cron/src/jobs/pins.js

If you call `pins.set` with a pin that is not "pinned", it'll actually add an entry to the `FOLLOWUPS` KV. The cron job iterates over everything and checks the status with IPFS Cluster. If the status changed it saves the new value, and if it becomes pinned it asks cluster for it's DAG size and saves that too.

vasco-santos (approved):
LGTM
Thanks for researching and getting these improvements in

PR #205, opened 2021-06-23
fix: add storeCar to client interface.ts

olizilla:
I'm a ts-noob, but this feels like something we could test. Is there a way we can check the interface is complete in CI?

License: MIT

### DIFF
--- a/packages/client/src/lib/interface.ts
+++ b/packages/client/src/lib/interface.ts
@@ -49,7 +49,11 @@ export interface API {
   /**
    * Stores CAR file and returns a corresponding CID.
    */
-  storeCar(service: Service, content: Blob | File): Promise<CIDString>
+  storeCar(
+    service: Service,
+    content: Blob | File,
+    options?: { onStoredChunk?: (size: number) => void }
+  ): Promise<CIDString>
   /**
    * Stores a directory of files and returns a CID. Provided files **MUST**
    * be within a same directory, otherwise error is raised. E.g. `foo/bar.png`,
--- a/packages/client/src/lib/interface.ts
+++ b/packages/client/src/lib/interface.ts
@@ -47,6 +47,10 @@ export interface API {
    */
   storeBlob(service: Service, content: Blob | File): Promise<CIDString>
   /**
+   * Stores CAR file and returns a corresponding CID.
+   */
+  storeCar(service: Service, content: Blob | File): Promise<CIDString>
+  /**
    * Stores a directory of files and returns a CID. Provided files **MUST**
    * be within a same directory, otherwise error is raised. E.g. `foo/bar.png`,
    * `foo/bla/baz.json` is ok but `foo/bar.png`, `bla/baz.json` is not.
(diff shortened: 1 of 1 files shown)


packages/client/src/lib/interface.ts
```diff
@@ -46,6 +46,10 @@ export interface API {
    * Stores a single file and returns a corresponding CID.
    */
   storeBlob(service: Service, content: Blob | File): Promise<CIDString>
+  /**
+   * Stores CAR file and returns a corresponding CID.
+   */
+  storeCar(service: Service, content: Blob | File): Promise<CIDString>
```
alanshaw, line 52: I think this takes an options object too now.

vasco-santos:
>  Is there a way we can check the interface is complete in CI?

There is, but no clean approach. I was actually thinking about creating a module/action that could do this. Usually, we need to create a test project like https://github.com/multiformats/js-multiformats/tree/master/test/ts-use and run a CI job

packages/client/src/lib/interface.ts
olizilla: like dis?

```ts
  storeCar(
    service: Service,
    content: Blob | File,
    options?: { onStoredChunk?: (size: number) => void }
  ): Promise<CIDString>
```

packages/client/src/lib/interface.ts
```diff
@@ -46,6 +46,14 @@ export interface API {
    * Stores a single file and returns a corresponding CID.
    */
   storeBlob(service: Service, content: Blob | File): Promise<CIDString>
+  /**
+   * Stores CAR file and returns a corresponding CID.
+   */
+  storeCar(
+    service: Service,
+    content: Blob | File,
```
vasco-santos, line 54:
```suggestion
    content: Blob |CarReader,
```

and carReader needs to be imported

olizilla:
Curiously, the API impls conformance to the ts definition does get tested in the build step, but it does not complain about the missing options object. I'm missing a context about what tsc is actually verifying here.

packages/client/src/lib/interface.ts
olizilla: is?

content: Blob | File | CarReader,

in the case where you upload a CAR from a file input, you'd have a File. Though as File implements Blob, i'm not sure what benefit we get from stating that either is acceptable.

vasco-santos: yeah, just edited

vasco-santos: > in the case where you upload a CAR from a file input, you'd have a File. Though as File implements Blob, i'm not sure what benefit we get from stating that either is acceptable.

Yes, I think we can drop File

alanshaw: `File` helps _us_ if we want to access properties on content that only `File` implements. I think it doesn't matter for consumers.

olizilla: We don't use any File properties in the current impl, so it is removed.

alanshaw (approved):

PR #208, opened 2021-06-23
fix: remove 100mb limit from uploader

alanshaw:
This PR changes the uploader to always pack the file into a CAR so the upload limit is removed.

### DIFF
--- a/packages/website/next-env.d.ts
+++ b/packages/website/next-env.d.ts
@@ -1,2 +1,3 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
+/// <reference types="next/image-types/global" />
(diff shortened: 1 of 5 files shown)


vasco-santos (approved):
LGTM! 
Sad that File does not allow us to get the total size, otherwise we could add to the uploaded info the total size, but also the user can see it locally

alanshaw:
> LGTM!
> Sad that File does not allow us to get the total size, otherwise we could add to the uploaded info the total size, but also the user can see it locally

We could use the `size` property no? https://developer.mozilla.org/en-US/docs/Web/API/File/size

What would we show - percentage complete?

vasco-santos:
> We could use the size property no? https://developer.mozilla.org/en-US/docs/Web/API/File/size

Oh, I looked at https://github.com/web-std/file/blob/default/src/lib.js and did not see size in there. But if we have in the browser, yes let's to it

> What would we show - percentage complete?

I think that is more user friendly than x mb / y mb, so yes percentage ❤️

vasco-santos (approved):
LGTM 💯

PR #214, opened 2021-06-24
fix: the CarReader type is a collection of interfaces

alanshaw:
This resolves type errors for folks wishing to pass a `CarIndexedReader` to `storeCar`.

`CarIndexedReader` does not extend the `CarReader` class but they both implement `RootsReader` and `BlockReader`...and these are the interfaces carbites needs.

https://github.com/ipld/js-car/blob/da397a8b3659a3e3ead9411746962885537ec1ac/api.ts#L49

### DIFF
--- a/packages/client/src/lib/interface.ts
+++ b/packages/client/src/lib/interface.ts
@@ -1,7 +1,8 @@
 import type { CID } from 'multiformats'
 export type { CID }
 
-import type { CarReader } from '@ipld/car'
+import type { RootsReader, BlockReader } from '@ipld/car/api'
+type CarReader = RootsReader & BlockReader
 export type { CarReader }
 
 /**
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -39,7 +39,7 @@
     "@web-std/fetch": "^2.0.1",
     "@web-std/file": "^1.1.0",
     "@web-std/form-data": "^2.1.0",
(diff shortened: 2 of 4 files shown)


vasco-santos (approved):
LGTM! Good catch, probably @ipld/car should also export such type, but let's not block this on that

packages/client/src/lib/interface.ts
```diff
@@ -1,7 +1,8 @@
 import type { CID } from 'multiformats'
 export type { CID }
 
-import type { CarReader } from '@ipld/car'
+import type { RootsReader, BlockReader } from '@ipld/car/api'
+type CarReader = RootsReader & BlockReader
```
olizilla, line 5: Paging @rvagg. Perhaps for convenience @ipld/car should export an interface for CarReaderish things

rvagg:
I'd love some TS eyes on https://github.com/ipld/js-car/pull/32 please. I'm pretty sure that it means you can both `import { CarReader } from '@ipld/car/api'` to get an interface _and_ `import { CarReader } from '@ipld/car/reader'` to get a class that implements that interface in TS, although you'd have to do a rename there to avoid a conflict if you use both. My hesitation is in the mixing of namespaces and reusing the symbols, but I think as long as they are separate modules then it should be fine since we have rename affordances everywhere that it matters.

PR #218, opened 2021-06-26
Update FAQ

momack2:
Updated the question about data storage to attempt to mitigate some of the flack we've gotten on Twitter - and added an FAQ about storage limits now that we've removed the cloudflare worker limit!

### DIFF
--- a/packages/website/pages/index.js
+++ b/packages/website/pages/index.js
@@ -407,15 +407,26 @@ function FAQ() {
         </p>
         <h2 className="chicagoflf white">
           <HashLink id="how-long-will-data-be-stored-on-nftstorage">
-            How long will data be stored on nft.storage?
+            How long will data be stored on NFT.storage?
           </HashLink>
         </h2>
         <p className="lh-copy white">
-          Data will be available in IPFS indefinitely. See the{' '}
+          Data will be available in IPFS indefinitely as well as stored in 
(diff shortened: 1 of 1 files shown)


packages/website/pages/index.js
```diff
@@ -407,16 +407,27 @@ function FAQ() {
...
         </p>
+        <h2 className="chicagoflf white">
+          <HashLink id="are-there-any-size-restrictions-for-stored-nfts">
+            Are there any size restrictions for stored NFTs?
+          </HashLink>
+        </h2>
+        <p className="lh-copy white">
+          NFT.storage can store NFTs up to <strong>32GB </strong>
+          in size! (There was previously a 100MB limit due to 
+          Cloudflare workers that has since been removed!🎉)
```
alanshaw, line 429: There is still a 100MB per-request limit, but we now accept chunked uploads so no single NFT is limited by the per-request limit.

```suggestion
          NFT.storage can store NFTs up to <strong>32GB </strong>
          in size! (There was previously a 100MB limit due to 
          Cloudflare workers but NFT.storage now supports chunked
          uploads, allowing files bigger than 100MB to be uploaded! 🎉)
```

alanshaw (changes requested):

alanshaw (approved):

PR #234, opened 2021-07-06
feat: scheama update for nft.storage

Gozala:
This is an attempt to bring schema changes from [web3.storage](https://raw.githubusercontent.com/web3-storage/web3.storage/main/packages/db/fauna/schema.graphql) and align it with requirements of niftysave and nft.storage.

This probably needs more work, but at least this gives us place to iterate / discuss.


Gozala:
Major difference from prior iterations here is that instead of trying to fold all uploads types into one this draft uses separate upload types for each. I am not sure this is ideal, however I find it better then trying to normalize into same struct and trying to derive based on it's properties.

hugomrdias:
LGTM i just would like to understand the split into several Upload types, what use case took you to do this?

Gozala:
> LGTM i just would like to understand the split into several Upload types, what use case took you to do this?

mostly it felt like we were forcing square peg through a circle. Furthermore if we were to add new kind of upload type in the future thing won’t fit again.

not forcing things into same shape makes things simpler today and won’t create problems with future types either

hugomrdias:
but it creates multiple collections for nfts uploads which makes the queries more complex, maybe we can find a better solution

hugomrdias:
closing this because we are moving to postgres

PR #287, opened 2021-08-06
revert: local add everything

alanshaw:


ribasushi (approved):
@alanshaw we could deploy it, but I'd feel safer doing it later
LGTM either way

alanshaw:
When I merge it'll deploy to staging. We deploy to production later.

PR #350, opened 2021-08-30
fix: pin ipfs-car verion to avoid empty upload issue

olizilla:
uploading via website fails with ipfs-car>=0.5 (you get an empty dir with cid `bafybeiczsscdsbs7ffqz55asqdf3smv6klcw3gofszvwlyarci47bgf354` for all uploads). Rolling back as a quick fix.

License: (Apache-2.0 AND MIT)


olizilla:
self merging to expdite fix for prod issue.

PR #358, opened 2021-08-31
WIP pgsql provisioning script, dbml diagram, migrations via prisma

the-simian:
adds the work in progress to a WIP branch to show other team members  for review

diagram:
image


the-simian:
This is just a WIP branch and is open to share some cod in motion with others. If you're seeing this and want to make more sense of it start with the readme I composed to help get going. 

https://github.com/ipfs-shipyard/nft.storage/pull/358/files#diff-6e81b2b575ce775687386a711695ddcb868c1ca22214ebccab84d385435b9ff8

alanshaw (commented):
Just a few comments:

* I'm not sure I understand why we've accumulated multiple SQL migrations here. We don't have _any_ postgres data yet do we?
* We've used snake case for postgres table/column/etc names in dagcargo and nft.storage and web3.storage schemas that are being worked on. I believe because that's the convention in postgres. Can we switch here? It would be nice to have consistency.
* I'd like to understand what is generated and what is source. I see at least one file here is generated...but I'm not sure about the rest?

the-simian:
@alanshaw did you read the readme I included?

Gozala:
Hey Folks,

Lets work as a one team please 🥺 Me acting as a middleman is not going to be great use of anyones time. I'll comment on couple of points made inline below:

- Multiple migrations
  @alanshaw after you dropped from the call other day @hugomrdias @the-simian @mikeal and myself agreed that it would be best if @hugomrdias just moved ahead with the schema for nft.storage and let the @the-simian setup a schema management system for teams consideration. Action item for this is https://github.com/ipfs-shipyard/nft.storage/issues/379
  
  However later after talking to @the-simian we've agreed that it would be best to leave schema management system alone for now and deal with that after we've all migrated.
  
  I think code here predates that decision. @the-simian given that we've decided not to do schema management/migration now can we please leave that out for now.

- table naming, casing etc...
  @the-simian please don't put me in this position (nor can I override @alanshaw's request here). We are a team and I hope we can collectively make decisions by reasoning through pros and cons of available choices. If there is a good reason in support for using different naming convention please elaborate so that they could be evaluated.

the-simian:
I'm going to close this pull request since it predates key decisions, and I think the confusions about that is resulting in miscommuncation

PR #394, opened 2021-09-03
Wip pg provision script and diagram

the-simian:
image


Gozala:
Generated diagram from https://github.com/web3-storage/web3-schema/pull/5 to be able to so for easy comparison https://dbdiagram.io/d/61329357825b5b0146f27eb7

Gozala:
Generated diagram from .dbml file here https://dbdiagram.io/d/613293c9825b5b0146f27f32

__WIP_pg_schema_review/readme.md
```diff
@@ -0,0 +1,47 @@
+# If you only read one thing, read the `provision.sql` file
+
+# SQL Migrataion scratchpad Repo
+
+This is a repo to test out some ideas for migrating and defining the sql schemas in the postgres database that will be the target for the migration away from fauna
+
+At this moment (8/30/2021) there's just a few folders
+
+- niftysave- Literally just a copy of the niftysave repo so its nearby and convenient to look at
```
Gozala, line 9: This is gone now

```suggestion
```

__WIP_pg_schema_review/readme.md
```diff
@@ -0,0 +1,47 @@
...
+- prisma-and-gql-test - right now its empty but I plan to test a nexus endpoint on top of the generated prisma client to see how difficult it is
+- prisma-test - this is where I've worked most recently and its got the schemas and migrations generated from prisma
```
Gozala, line 11: I think above two bullet points our outdated now.

__WIP_pg_schema_review/readme.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Quick Start
+
+1. Make sure you have docker installed and working, im running a postgres image there atm
+2. cd into the `prisma-test` and `docker-compose up` should start the image. I've exposted the port on 5439 so it doesn't clash with other things on my system but you're welcome to change this.
+3. Be aware the connection string is in the .env file. This is just a test to its `jondoe`, `randompassword` for now.
+4. `npm install`. The Prisma directory has 3 things
+   - **dbml**. Every time we run `npx prisma generate` a new .dbml file is made. You can paste this at `https://dbdiagram.io/d` and check out the visualizer
+   - **migrations**. This is where every named migration lives. Once we've made a good schema we can put thes in a .old folder and jsut make the one init migration but in the meantime you can observe the sort of sql prisma emits here
```
Gozala, line 20: Seems outdated

```suggestion
```

__WIP_pg_schema_review/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "prisma-test",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
Gozala, line 7:
```suggestion
    "test": "echo \"Error: no test specified\" && exit 1",
    "gen:dbml": "prisma generate"
```

__WIP_pg_schema_review/readme.md
```diff
@@ -0,0 +1,47 @@
...
+- prisma-and-gql-test - right now its empty but I plan to test a nexus endpoint on top of the generated prisma client to see how difficult it is
+- prisma-test - this is where I've worked most recently and its got the schemas and migrations generated from prisma
+
+## Quick Start
+
+1. Make sure you have docker installed and working, im running a postgres image there atm
+2. cd into the `prisma-test` and `docker-compose up` should start the image. I've exposted the port on 5439 so it doesn't clash with other things on my system but you're welcome to change this.
+3. Be aware the connection string is in the .env file. This is just a test to its `jondoe`, `randompassword` for now.
+4. `npm install`. The Prisma directory has 3 things
+   - **dbml**. Every time we run `npx prisma generate` a new .dbml file is made. You can paste this at `https://dbdiagram.io/d` and check out the visualizer
```
Gozala, line 19:
```suggestion
   - **dbml**. Every time we run `npm gen:dbml` a new `.dbml` file is made. You can paste this at `https://dbdiagram.io/d` and check out the visualizer
```

__WIP_pg_schema_review/readme.md
```diff
@@ -0,0 +1,47 @@
...
+   - **migrations**. This is where every named migration lives. Once we've made a good schema we can put thes in a .old folder and jsut make the one init migration but in the meantime you can observe the sort of sql prisma emits here
+   - **schema.prisma**. The actual Prisma schema!
+
+The commands you care about most are:
+
+1. `npx prisma migrate dev` : will create a named migration, and you can go look at the sql
```
Gozala, line 25: I think we're leaving it out for now

```suggestion
1. `npx prisma migrate dev` : will create a named migration, and you can go look at the sql
```

__WIP_pg_schema_review/readme.md
```diff
@@ -0,0 +1,47 @@
+# If you only read one thing, read the `provision.sql` file
```
Gozala, line 1: I would make a note that it's generated from .prima file

Gozala (commented):
@the-simian I understand you're trying to make it clear this is a draft, but can you still please move `__WIP_pg_schema_review` into `packages/database_v2` ?

__WIP_pg_schema_review/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules
```
Gozala, line 1: This is obsolete `.gitignore` in the root already has all this.

__WIP_pg_schema_review/prisma/dbml/schema.dbml
```diff
@@ -0,0 +1,147 @@
...
+  uri String [pk, unique]
+  ipfs_url String [unique, not null]
+  problem String [not null]
+  cid String
+  content content
+}
+
+Table content {
+  cid String [pk, unique]
+  resources resource [not null]
```
Gozala, line 60: Is there way to visualize fields that aren't actually in the table somehow differently ? Looking at .sql I see there is no `resources` table, but looking at the diagram they look like actual fields https://dbdiagram.io/d/613293c9825b5b0146f27f32

__WIP_pg_schema_review/readme.md
the-simian: sorry this should be deleted, I forgot to remove it on my new branch

the-simian: yes I'll move it into the databases section

__WIP_pg_schema_review/readme.md
the-simian: yes I just copied it at the root, I'll actually make a convenience script for this

__WIP_pg_schema_review/prisma/dbml/schema.dbml
the-simian: I'll look into this

the-simian:
These will be important, I'm keeping them updated as I go:

This is a visualization of the generated SQL file imported and converted to dbml
- https://dbdiagram.io/d/613596c1825b5b0146f415eb

This is a visualization of the prisma's schema file to dbml
- https://dbdiagram.io/d/6135955a825b5b0146f41465

Dropping them here for reference

__WIP_pg_schema_review/prisma/dbml/schema.dbml
the-simian: this might be what we need in this case:  https://github.com/ipfs-shipyard/nft.storage/pull/394#issuecomment-913334271

You can see whats in the actual table vs the 'prisma api'.  for the short term we just care about the actual tables, I've converted all the joins into explicit join tables, so I think it will be a lot easier to read now.

the-simian:
questions about this (I think @alanshaw you might know what is best)
https://github.com/web3-storage/web3-schema/blob/815f825ac7469fcfe131db7a725d4b94dea66225/schema.sql#L46
On pin status, it looks like I have some different status than you, should both of these be like:

```
  CREATE TYPE pin_status AS ENUM (
    'ClusterError',
    
    'PinError',  
    'PinQueued',
    'Pinned',   
    'Pinning',  
    
    'Remote',  --is remote and sharded a form of error, or something else?
    'Sharded', 
    
    'Undefined', -- what causes this to occur?
    
    'UnpinError',
    'UnpinQueued',
    'Unpinned',
    'Unpinning',
  )
  ```

here's where there are some comments related to ours, I am assuming that 'PinFailed' and 'PinError' are just the same thing
https://github.com/ipfs-shipyard/nft.storage/pull/394/files#diff-d60b876e8930f8c1495c714bc2258bca1cb80a6eac2d9b5bc59615befbc4475eR35

**Comment by the-simian** [comment]
My other questions are about the pin table, I put my questions then a copy of both side by side here so its easy to read


1. I can see that we both have id as the primary key (although I chose text because I wasn't sure if the id would only be numbers or something else but I can change that of course). Should I change mine to BIGSERIAL, is it always going to be a number?

2. right now `pin_location_id` and `location_id` are the same. To me, it makes sense to omit `pin` because you're in the pin table, so `location_id` is the preferable property. same with `pin_status`, mine should probably change (from pin) to `status` and also yours too, since we're in the pin table, and it 'owns' its status. I just ported that directly from the fauna models to get started, I don't think 'pin'  in my own table makes sense at least. what do you think?

3. I've also chosen `created` and `updated` across the board as the convention in the schema I have. should I change all of these to `inserted_at` and `updated_at` ? I mainly just took this right out of fauna. If my own preference matters I tend to go with `created_at` and `updated_at` and I put it in nearly every table. I am ambivalent about what we do but should probably be identical everywhere.

4. what's left is cid, it looks like there's now multiple `cid`s should I change mine to `content_cid` since that is the kind of `cid` it references (to match your schema)? Is it ambiguous at just `cid`?


for reference: 

```
  CREATE TABLE pin (
    id BIGSERIAL PRIMARY KEY,
    content_cid TEXT NOT NULL,
    pin_location_id bigint NOT NULL,
    status pin_status NOT NULL,
    inserted_at timestamp with time zone DEFAULT timezone('utc'::text, now()) NOT NULL,
    updated_at timestamp with time zone DEFAULT timezone('utc'::text, now())
  )
  ```
  
  ```
  CREATE TABLE "pin" (
    "id" TEXT NOT NULL,
    "cid" TEXT NOT NULL,
    "location_id" TEXT NOT NULL,
    "pin" "pin_status" NOT NULL,
    "status_text" TEXT,
    "updated" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "created" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,

    PRIMARY KEY ("id")
);
```

I might have questions about other tables but the pin status and pin table looks like something in my own needs to change at minimum based on what I'm reading. I'd love your feedback cc @Gozala

**Comment by the-simian** [comment]
I just took some time reading through your dotStorage DB migration information, really helpful @alanshaw . I think our pin_location tables are basically identical and in line with the doc, which is good. I did notice - should I change every `id` to big serial?

I initially picked `TEXT` just in case whatever is generating was using uuids or if there was a chance there was historical data with non-numbers. Is it totally safe to just `BIGSERIAL` every last one? I can make that change if that's the case! :)

**Comment by alanshaw** [comment]
Super quickly:

nft.storage does not currently have pin location, but web3.storage does. **However** it's currently under review so nft.storage initially will _not_ have pin locations for the migration (the pin table is essentially the same as what it currently has in the cloudflare KV).

I was under the impression that the current niftysave pin table was the same/similar to the proposed nft.storage one. 

re: `BIGSERIAL` - need to check on individual cases and how the migration would work but I think yes, you should be using that as I believe we don't generate IDs in Fauna.

re: timestamps - yes suffix with "_at" to be consistent with nft.storage and web3.storage.

I'll circle back to this tomorrow, but just incase you're blocked hopefully it should help a little?

**Comment by alanshaw** [comment]
Note: ideally, you'll not include `content` or `pin` tables in the schema here but reference them as `nft_storage.content` etc.

**Comment by alanshaw** [comment]
@Gozala there doesn't seem to be any data in the pin or pin location collections in the nft.storage fauna db - can you clarify what's going on here?

**Comment by Gozala** [comment]
List of action items after comparing [nft.storage schema](https://dbdiagram.io/d/6137ad47825b5b0146f766cf) as per #395 and [niftysave schema](https://dbdiagram.io/d/613596c1825b5b0146f415eb) as per link above.

- [ ] Rename `created` field in `content` table to `inserted_at`.
- [ ] Add `updated_at` field to `content` table.
- [ ] Rename `cid` field in `pin` table to `content_cid`.
- [ ] Rename `pin` field to `status` in `pin` table.
- [ ] Make `pin_status` enum have the same variants as in `nft.storage`.
- [ ] Make `id` in `pin` table a `BIGSERIAL`.
- [ ] Rename `created` field to `inserted_at` in `pin` table.
- [ ] Rename `updated` field to `updated_at` in `pin` table.
- [ ] Add `service` field to `pin` table and corresponding enum type.
- [ ] Remove `location_id` field from `pin` table.
- [ ] Remove `pin_location` table (@alanshaw is that a plan for nft.storage and web3.storage right now ?)
- [ ] Rename `cid` field in `resource` table to `content_cid` to follow naming same naming convention as in `*.storage`

**Comment by the-simian** [comment]
@Gozala I'm on it, this looks super clear, thank you.

**Comment by Gozala** [comment]
> questions about this (I think @alanshaw you might know what is best)
> https://github.com/web3-storage/web3-schema/blob/815f825ac7469fcfe131db7a725d4b94dea66225/schema.sql#L46
> On pin status, it looks like I have some different status than you, should both of these be like:
> 
> ```
>   CREATE TYPE pin_status AS ENUM (
>     'ClusterError',
>     
>     'PinError',  
>     'PinQueued',
>     'Pinned',   
>     'Pinning',  
>     
>     'Remote',  --is remote and sharded a form of error, or something else?
>     'Sharded', 
>     
>     'Undefined', -- what causes this to occur?
>     
>     'UnpinError',
>     'UnpinQueued',
>     'Unpinned',
>     'Unpinning',
>   )
> ```
> 
> here's where there are some comments related to ours, I am assuming that 'PinFailed' and 'PinError' are just the same thing
> https://github.com/ipfs-shipyard/nft.storage/pull/394/files#diff-d60b876e8930f8c1495c714bc2258bca1cb80a6eac2d9b5bc59615befbc4475eR35

This is related to discussion that happened here https://github.com/web3-storage/web3.storage/issues/31. At a time I went ahead with what made sense to me and have not got around to syncing up with @alanshaw on this.

I still maintain position that it's best to enforce data invariant at insertion time as opposed to query time, that way client always gets data that makes sense and we're forced to think how to handle instances where invariants aren't upheld. Doing it at query unfortunately in practice means user has to deal with states that are confusing or make no sense because we have not got around to dealing with it.

**Comment by Gozala** [comment]
Just to be completely clear, it is @alanshaw's call to decide what the enum should be given all the cons and pros. In the niftysave we should just use same enum as *.storage.

**Comment by Gozala** [comment]
>     1. I can see that we both have id as the primary key (although I chose text because I wasn't sure if the id would only be numbers or something else but I can change that of course). Should I change mine to BIGSERIAL, is it always going to be a number?

If I recall correctly fauna team indicated that their generated IDs are basically timestamps so there should be no problem there. As of stuff coming off chain, we should just use types as they come off chain. For the reference here is the graph we query to get data off chain https://thegraph.com/legacy-explorer/subgraph/nftstorage/eip721-subgraph and here is the definition of it https://github.com/nftstorage/eip721-subgraph

**Comment by Gozala** [comment]
> 1. ...
> 1. right now pin_location_id and location_id are the same. To me, it makes sense to omit pin because you're in the pin table, so location_id is the preferable property. same with pin_status, mine should probably change (from pin) to status and also yours too, since we're in the pin table, and it 'owns' its status. I just ported that directly from the fauna models to get started, I don't think 'pin' in my own table makes sense at least. what do you think?
> 1. I've also chosen created and updated across the board as the convention in the schema I have. should I change all of these to inserted_at and updated_at ? I mainly just took this right out of fauna. If my own preference matters I tend to go with created_at and updated_at and I put it in nearly every table. I am ambivalent about what we do but should probably be identical everywhere.
> 1. what's left is cid, it looks like there's now multiple cids should I change mine to content_cid since that is the kind of cid it references (to match your schema)? Is it ambiguous at just cid?

I believe those are addressed by the checklist. Give that nft.storage / web3.storage are aligned and further ahead lets just follow the lead.

**Comment by Gozala** [comment]
> Note: ideally, you'll not include `content` or `pin` tables in the schema here but reference them as `nft_storage.content` etc.

We had a little chat with @alanshaw today about whether it would make sense to try and converge nft.storage and niftysave and I believe we came to conclusion that in initial migration it would be best to reduce coordination and there for not try to have same db etc...

@alanshaw does that comment still hold given that decision ? I suppose we could define `nft_storage` schema and prefix our tables as such, although I'm not sure of there are any tradeoffs with that.

**Comment by Gozala** [comment]
> @Gozala there doesn't seem to be any data in the pin or pin location collections in the nft.storage fauna db - can you clarify what's going on here?

That is correct. Plan had always been that nft.storage will update those once it migrates but that never got completed so we don't currently end up with data there. 

I thin that plans still makes sense once nft.storage and niftysave share the schema, however depending where individual migrations will be we might do it before nft.storage.

File: packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql [hunk 10]
```diff
@@ -0,0 +1,222 @@
...
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "resource" (
+    "status" "resource_status" NOT NULL,
+    "status_text" TEXT NOT NULL,
+    "uri" TEXT NOT NULL,
+    "ipfs_url" TEXT NOT NULL,
+    "problem" TEXT NOT NULL,
```
Gozala, line 70: This seems like leftover from older schema it no longer exists

https://github.com/ipfs-shipyard/nft.storage/blob/343b835dc902454b91e702e2845e9f0ec6cae471/packages/database/fauna/resources/schema.graphql#L250-L308

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
```diff
@@ -0,0 +1,222 @@
...
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "resource" (
+    "status" "resource_status" NOT NULL,
+    "status_text" TEXT NOT NULL,
+    "uri" TEXT NOT NULL,
+    "ipfs_url" TEXT NOT NULL,
```
Gozala, line 69: This actually can be null. I think name might be confusing here and I'm open to suggestion to make it more obvious.

As [per comment in the schema](https://github.com/ipfs-shipyard/nft.storage/blob/343b835dc902454b91e702e2845e9f0ec6cae471/packages/database/fauna/resources/schema.graphql#L287-L295) what happens here is that we look at `uri` field and attempt to derive corresponding `ipfs://` URL from it:

https://github.com/ipfs-shipyard/nft.storage/blob/343b835dc902454b91e702e2845e9f0ec6cae471/packages/niftysave/src/ipfs-url.js#L29-L35

If it original `uri` was malformed `ipfs://` URL it will be fixed up if it was IPFS gateway url e.g. `https://ipfs.io/ipfs/Qm...hash` it will  be reformatted as `ipfs://` but if it was some arbitrary url e.g. `https://flickr.com/...` that it will have no `ipfs_url` at all.

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
```diff
@@ -0,0 +1,222 @@
...
+    "cid" TEXT,
+    "created" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updated" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    PRIMARY KEY ("uri")
+);
+
+-- CreateTable
+CREATE TABLE "resources_on_metadata" (
+    "metadata_id" TEXT NOT NULL,
```
Gozala, line 80: As alluded to earlier we should just use cid of the metadata here instead.

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
```diff
@@ -0,0 +1,222 @@
...
+    "peer_name" TEXT NOT NULL,
+    "region" TEXT NOT NULL,
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "token_contract" (
+    "id" TEXT NOT NULL,
+    "name" TEXT NOT NULL,
```
Gozala, line 122: I think it may be null at least graphql schema for the graph has this as optional.

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
```diff
@@ -0,0 +1,222 @@
...
+    "symbol" TEXT NOT NULL,
```
Gozala, line 123: I think in practice this is mandatory but according to the graph we get it from it is optional.

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
```diff
@@ -0,0 +1,222 @@
...
+    "id" TEXT NOT NULL,
+    "next_id" TEXT NOT NULL,
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "erc721_import_result_to_token" (
+    "erc721_import_result_id" TEXT NOT NULL,
+    "token_id" TEXT NOT NULL,
```
Gozala, line 140: 😬😱 Having both `token.id` and `token_id` is getting out of hand. maybe we should rename `token` table to `nft` to make things less confusing.

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
```diff
@@ -0,0 +1,222 @@
...
+    "id" TEXT NOT NULL,
+    "name" TEXT NOT NULL,
+    "symbol" TEXT NOT NULL,
+    "supports_eip721_metadata" BOOLEAN NOT NULL,
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "erc721_import_result" (
```
Gozala, line 130: I think we could probably rename this to `erc721_import`, I had result there to workaround with the fact that you have input types and output types in GraphQL and reflecting on this now all other output types end up without `result` suffix.

packages/database_v2/__WIP_pg_schema_review/artifacts/schema.graphql
```diff
@@ -0,0 +1,651 @@
+type Block {
```
Gozala, line 1: Can we leave this out for now, I think it out of scope here and just adds noise.

packages/database_v2/__WIP_pg_schema_review/prisma/migrations/migration_lock.toml
```diff
@@ -0,0 +1,3 @@
+# Please do not edit this file manually
```
Gozala, line 1: Can we instead leave out whole migrations dir in this pr

Gozala (changes requested):

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
```diff
@@ -0,0 +1,222 @@
+-- Generated by prisma gainst the schema.prisma file
```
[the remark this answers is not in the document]
the-simian: I'm going to look into this probably first tomorrow. I think it is I agree that would be really useful in our situation

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
the-simian: yeah, that's a much better name. I think we're at the point where we're pas the the 1:1 from fauna and the schema is being iterated on and we can pick ideal names so I'm going to be making a pass thinking of that as well, looking for things like this

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
the-simian: I agree that's a better name, it also doesn't cause confusion with user tokens/session tokens. Token just means too many things in programming already.

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
```diff
@@ -0,0 +1,222 @@
...
+    "token_uri" TEXT NOT NULL,
+    "ipnft" TEXT NOT NULL,
+    "problem" TEXT,
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "metadata" (
+    "id" TEXT NOT NULL,
```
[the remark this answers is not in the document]
the-simian: changing this to resources, based on the discussion which actually lines up with the table name and, it is the 'R in URL so it makes a lot of sense here.

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
```diff
@@ -0,0 +1,222 @@
...
+    "cid" TEXT,
+    "source_id" TEXT NOT NULL,
```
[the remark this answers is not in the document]
the-simian: this is currently replaced (and linked) by `token_uri` on the nft_asset table, which I think was where we landed in our discussion, so I'm going to resolve this

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
```diff
@@ -0,0 +1,222 @@
...
+    "image_uri" TEXT NOT NULL,
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "resource" (
+    "status" "resource_status" NOT NULL,
+    "status_text" TEXT NOT NULL,
+    "uri" TEXT NOT NULL,
```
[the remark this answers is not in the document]
the-simian: to keep it I ended up making the join table `resource_source_uri`, which.. I am not sure if that's awkward or not? I'll resolve this I made a note to check this out in our review in the AM

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
the-simian: ok yes, it looks like status and status_text are the new entities, and I've updated the enum as well.

packages/database_v2/__WIP_pg_schema_review/artifacts/provision.sql
the-simian: I remember coming across this when I was working on the search app. I've marked this as optional. I feel like the name is ok, at least in the context of when I was making a consumer.

packages/database_v2/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "prisma-test",
```
Gozala, line 2:
```suggestion
  "name": "@nftstorage/niftysave-db",
```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+
+    PRIMARY KEY ("hash")
+);
+
+-- CreateTable
+CREATE TABLE "nft" (
+    "id" TEXT NOT NULL,
+    "token_id" TEXT NOT NULL,
+    "mint_time" TIMESTAMP(3) NOT NULL,
+    "nft_asset_id" TEXT NOT NULL,
```
Gozala, line 26:
```suggestion
    "token_uri" TEXT NOT NULL,
```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+CREATE UNIQUE INDEX "pin.id_unique" ON "pin"("id");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "contract.id_unique" ON "contract"("id");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "erc721_import.id_unique" ON "erc721_import"("id");
+
+-- AddForeignKey
+ALTER TABLE "nft" ADD FOREIGN KEY ("nft_asset_id") REFERENCES "nft_asset"("token_uri") ON DELETE CASCADE ON UPDATE CASCADE;
```
Gozala, line 183:
```suggestion
ALTER TABLE "nft" ADD FOREIGN KEY ("token_uri") REFERENCES "nft_asset"("token_uri") ON DELETE CASCADE ON UPDATE CASCADE;
```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+    "nft_asset_id" TEXT NOT NULL,
+    "contract_id" TEXT NOT NULL,
+    "owner_id" TEXT NOT NULL,
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "owner" (
+    "id" TEXT NOT NULL,
```
Gozala, line 35:
```suggestion
    "id" TEXT NOT NULL,
    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+);
+
+-- CreateTable
+CREATE TABLE "nft" (
+    "id" TEXT NOT NULL,
+    "token_id" TEXT NOT NULL,
+    "mint_time" TIMESTAMP(3) NOT NULL,
+    "nft_asset_id" TEXT NOT NULL,
+    "contract_id" TEXT NOT NULL,
+    "owner_id" TEXT NOT NULL,
```
Gozala, line 28:
```suggestion
    "owner_id" TEXT NOT NULL,
    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,

```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+-- CreateEnum
+CREATE TYPE "pin_status" AS ENUM ('ClusterError', 'PinError', 'PinQueued', 'Pinned', 'Pinning', 'Remote', 'Sharded', 'Undefined', 'UnpinError', 'UnpinQueued', 'Unpinned', 'Unpinning');
+
+-- CreateEnum
+CREATE TYPE "pin_service" AS ENUM ('Pinata', 'IpfsCluster');
+
+-- CreateTable
+CREATE TABLE "block" (
+    "hash" TEXT NOT NULL,
+    "number" INTEGER NOT NULL,
```
Gozala, line 16: In fauna capturing timestamps made little sense as db itself tracked it natively and you could query things in time. In postgres world we'd want to do it manually.

```suggestion
    "number" INTEGER NOT NULL,
    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+    "id" TEXT NOT NULL,
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "nfts_by_blocks" (
+    "block_hash" TEXT NOT NULL,
+    "nft_id" TEXT NOT NULL,
+    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
Gozala, line 44:
```suggestion
    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+    PRIMARY KEY ("token_uri")
+);
+
+-- CreateTable
+CREATE TABLE "metadata" (
+    "content_cid" TEXT NOT NULL,
+    "token_uri" TEXT NOT NULL,
+    "name" TEXT,
+    "description" TEXT,
+    "image_uri" TEXT NOT NULL,
```
Gozala, line 67:
```suggestion
    "image_uri" TEXT NOT NULL,
    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,

```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    PRIMARY KEY ("source_uri")
+);
+
+-- CreateTable
+CREATE TABLE "resources_by_metadata" (
+    "metadata_cid" TEXT NOT NULL,
+    "resource_source_uri" TEXT NOT NULL,
+    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
Gozala, line 89:
```suggestion
    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+
+    PRIMARY KEY ("metadata_cid","resource_source_uri")
+);
+
+-- CreateTable
+CREATE TABLE "content" (
+    "cid" TEXT NOT NULL,
+    "dag_size" INTEGER,
+    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
Gozala, line 98:
```suggestion
    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "contract" (
+    "id" TEXT NOT NULL,
+    "name" TEXT,
+    "symbol" TEXT,
+    "supports_eip721_metadata" BOOLEAN NOT NULL,
```
Gozala, line 121:
```suggestion
    "supports_eip721_metadata" BOOLEAN NOT NULL,
    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "erc721_import" (
+    "id" TEXT NOT NULL,
+    "next_id" TEXT NOT NULL,
```
Gozala, line 129:
```suggestion
    "next_id" TEXT NOT NULL,
    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

packages/database_v2/artifacts/provision.sql
```diff
@@ -0,0 +1,216 @@
...
+
+    PRIMARY KEY ("id")
+);
+
+-- CreateTable
+CREATE TABLE "erc721_import_by_nft" (
+    "erc721_import_id" TEXT NOT NULL,
+    "nft_id" TEXT NOT NULL,
+    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
Gozala, line 138:
```suggestion
    "inserted_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,    
```

Gozala:
Closing this, as it was superseded by #407

PR #395, opened 2021-09-06
feat: postgres schema

alanshaw:
I've taken the schema from https://github.com/ipfs-shipyard/nft.storage/pull/263 with the following changes:

* Primary keys are called `id` except where we're using domain data as the primary key (currently only `cid` in content).
* Use shortened primary key definition to use `BIGSERIAL` for enhanced readability.
* ~~Added `nft_storage` schema.~~
* Renamed `account` -> `user`.
* Make `issuer` NOT the primary key on `user`, added `id` column for primary and referenced it from other tables.
* Change `user.issuer` to `user.magic_link_id` and `user.sub` to `user.github_id`.
* Prefixed foreign keys with (complete) referenced table name where appropriate.
* Renamed `cid` in the `upload` table to `source_cid` so more obvious it is the given/original/provided/source CID.
* Added the `deleted_at` column to `upload` (and had to remove the `UNIQUE` constraint because of it).
* Consistent enums capitalization to match web3.storage and niftysave.
* Added `upload_type` enum.
* Added `mime_type` text field to `upload`.
* Renamed `content.size` to `content.dag_size` for consistency with web3.storage.

### DIFF
--- a/packages/db/tables.sql
+++ b/packages/db/tables.sql
@@ -29,8 +29,8 @@ CREATE TABLE IF NOT EXISTS nft_storage.user (
   github TEXT,
   email TEXT NOT NULL,
   public_address TEXT NOT NULL UNIQUE,
-  inserted_at timestamp with time zone DEFAULT timezone('utc'::text, now()) NOT NULL,
-  updated_at timestamp with time zone DEFAULT timezone('utc'::text, now()) NOT NULL
+  inserted_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
+  updated_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
 );
 
 CREATE TABLE IF NOT EXISTS nft_storage.auth_key (
@@ -38,16 +38,16 @@ CREATE TABLE IF NOT EXISTS nft_storage.auth_key (
   name TEXT NOT NULL,
   secret TEXT NOT NULL,
   user_id TEXT NOT NULL REFERENCES user ( id ),
-  inserted_at timestamp with time zone DEFAULT timezone('utc'::text, now()) NOT NULL,
-  updated_at timestamp with time zone DEFAULT timezone('utc'::text, now()) NOT NULL
+  inserted_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
+  updated_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
 );
 
 CREATE TABLE IF NOT EXISTS nft_storage.content (
   -- normalized base32 v1
   cid TEXT PRIMARY KEY,
   dag_size BIGINT,
-  inserted_at timestamp with time zone DEFAULT timezone('utc'::text, now()) NOT NULL,
-  updated_at timestamp with time zone DEFAULT timezone('utc'::text, now()) NOT NULL
+  inserted_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
+  updated_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
 );
 
 CREATE TABLE IF NOT EXISTS nft_storage.pin (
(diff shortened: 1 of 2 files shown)


packages/db/tables.sql
```diff
@@ -0,0 +1,75 @@
...
+);
+
+CREATE TYPE nftstorage.upload_type AS ENUM (
+  'Car',
+  'Blob',
+  'Multipart'
+);
+
+CREATE TABLE IF NOT EXISTS nftstorage.user (
+  id BIGSERIAL PRIMARY KEY,
```
hugomrdias, line 22: regarding the new `id`, i still think that we dont need it. 
Even if we change the auth again whatever identifier we use can go into the `issuer`, this id would never be used for selects. We always either get the issuer from tokens or the auth session and those will never be this new auto generated `id`.

Right now we have issuer and sub because we (i did :/) did a mistake and the tokens and kvs may have one of those, and we can't control that so we check both. In the migration im gonna normalize the kv data to always use `issuer` but we can still have tokens with `sub` inside so we need to keep checking for both.

If we start using a new id, we will probably end up checking for `issuer`, `sub` and `id`.

hugomrdias (changes requested):

packages/db/tables.sql
```diff
@@ -0,0 +1,75 @@
...
+CREATE TYPE nftstorage.service AS ENUM (
+  'Pinata',
+  'IpfsCluster'
+);
+
+CREATE TYPE nftstorage.upload_type AS ENUM (
+  'Car',
+  'Blob',
+  'Multipart'
+);
```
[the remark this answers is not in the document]
alanshaw: It's a different type though, maybe we should store as `mime_type`?

packages/db/tables.sql
alanshaw: > regarding the new `id`, i still think that we dont need it.
> Even if we change the auth again whatever identifier we use can go into the `issuer`, this id would never be used for selects. We always either get the issuer from tokens or the auth session and those will never be this new auto generated `id`.

Can it go into `issuer`? What do we do with existing users?

I just think we should identify users _ourselves_ and _store_ identifiers we get from auth providers along with the user...so that switching auth provider is a simple case of check the right field. This way there's no migrations or lazy upgrades and no confusions to be had about whether the field stores one type of ID or another. We can even support multiple providers.

The other problem I have is that `issuer` does not sound like an ID and I think it will be confusing to have an `issuer`/`user_issuer` in another table.

hugomrdias: ok, lets use the auto gen id and maybe change issuer to `magic_link_id` and sub to `github_id` sounds good ?

hugomrdias (changes requested):

packages/db/tables.sql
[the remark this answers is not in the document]
hugomrdias: so we do 
```sql
CREATE TYPE nftstorage.upload_type AS ENUM (
  'Car',
  'Blob',
  'Multipart',
  'Remote',
  'NFT'
);
```

and add another column for mime type ?

Gozala:
Generated diagram from here https://dbdiagram.io/d/6137ad47825b5b0146f766cf

packages/db/tables.sql
```diff
@@ -0,0 +1,79 @@
...
+  name TEXT NOT NULL,
+  secret TEXT NOT NULL,
+  user_id TEXT NOT NULL REFERENCES user ( id ),
+  inserted_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
+  updated_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS nft_storage.content (
+  -- normalized base32 v1
+  cid TEXT PRIMARY KEY,
```
the-simian, line 47: is this NOT NULL?

packages/db/tables.sql
```diff
@@ -0,0 +1,77 @@
+CREATE TYPE public.pin_status AS ENUM (
```
Gozala, line 1: In niftysave we end up with a [following enum](https://github.com/ipfs-shipyard/nft.storage/pull/407/files#diff-4996c80e080afc9e4bf0eb737f86b756a6cf5b536012a8d6607e49a36c8792f3R118-R123 ) quoting below:

```sql
CREATE TYPE pin_status AS ENUM (
    'PinError',
    'PinQueued',
    'Pinned',
    'Pinning'
);
```

I would suggest using same as above here since:

1. It [represents a subset used by web3.storage](https://github.com/web3-storage/web3-schema/blob/7331b863b77f3bc0a71d1810a2fd14e813cfee9c/schema.sql#L74-L101)
2. Adding more variants to enum should not require record updates, while it would if we would need to rename values.

alanshaw: @hugomrdias are you ok with this? It's fine with me.

hugomrdias: fine with me

alanshaw: Now tracked in https://github.com/ipfs-shipyard/nft.storage/issues/331

alanshaw:
Closing this as the changes have been manually merged into https://github.com/ipfs-shipyard/nft.storage/pull/263

PR #401, opened 2021-09-07
feat: add navbar and fix mobile navigation

rafaelramalho19:
This PR adds a navbar usage optimized for mobile with the help of an hamburger menu. It was inspired from web3.storage

## Demo:

image

image

image

## Notes:

Blue background may need adjustments to the logo (but this applies for both desktop and mobile, let's create a different issue with this?). But I styled it black when it's a bright background

image

Closes https://github.com/ipfs-shipyard/nft.storage/issues/361


packages/website/components/layout.js
```diff
@@ -31,7 +31,7 @@ export default function Layout({
   const shouldWaitForUser = needsUser && status === 'loading'
 
   return (
-    <div className="sans-serif">
+    <div className="sans-serif flex flex-column min-vh-100">
```
jdiogopeixoto, line 34: Now that we have this can you remove `min-vh-100` from the other pages?

jdiogopeixoto (approved):
LGTM! Just left one small comment.

packages/website/components/layout.js
rafaelramalho19: Yes, good call 👍

rafaelramalho19: Actually, if we remove the min-vh-100 from the secondary container in all the pages, they don't grow to fill the remaining space...

jdiogopeixoto: You're right, it needs additional work. We can handle that later 👍

PR #407, opened 2021-09-09
Feat/niftysave postgres

Gozala:
Changes in this pull request supposed to replace `packages/database` with `packages/database_v2` which is postgres setup for niftysave. Here is the [link to the schema](https://dbdiagram.io/d/613fb365825b5b0146004a76) (generated from init_tables.sql and init_foreign_keys.sql) and below is high level overview of changes:

1. `packages/database_v2/artifacts/init_tables.sql` file contains source code for initial database schema for niftysave.
2. Schema aligns `content` and `pins` tables with .storage schema's.
3. Initial schema intentionally does not contain foreign keys. This would simplify data migration from fauna to postgres as we will not have to deal with insert order across tables. Once migration is complete, key constraints will be added.
4. `init_foreign_keys.sql` file contains source code for adding primary keys after.
4. `packages/database_v2/docker-compose.yml ` defines a container that starts db and hasura graphql interface for running locally.
5. Bunch of other files are included that enable us to manage hasura (feel free to disregard)
    - `./config.yaml ` is [hasura configuration file](https://hasura.io/docs/latest/graphql/core/hasura-cli/config-reference.html#configuration-file)
    - `./metadata/*` is [hasura metadata](https://hasura.io/docs/latest/graphql/core/migrations/manage-metadata.html) for managing GraphQL API mapping & triggers.
    - `./migrations` directory is a setup for [handling database migrations & rollbacks](https://hasura.io/docs/latest/graphql/core/migrations/migrations-setup.html) with hasura. It's generated by hasura toolchain when you make changes to the db schema & can be applied to a remote / local db.


packages/database_v2/artifacts/init.sql
```diff
@@ -0,0 +1,168 @@
...
+    updated_at TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    inserted_at TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    PRIMARY KEY (id)
+);
+
+CREATE TABLE IF NOT EXISTS nft_asset (
+    token_uri TEXT NOT NULL,
+    ipfs_URL TEXT,
+    status nft_asset_status NOT NULL,
+    status_text TEXT NOT NULL,
```
Gozala, line 78: We have a problem here:

`nft` maps to `nft_asset` by `token_uri` but it can not map to `nft_metadat` by `token_uri` because different `token_uri`s may resolve to the same `content_cid`.

To address this we should have nullable `content_cid` in `nft_asset` table and we should remove `token_uri` from the `nft_metadata` table.

Gozala: Thinking more about it I am wondering if we should remove `nft_asset` table all together because:

1. It is (almost) identical to `resource` table and was mostly different in fauna so they would map to different GraphQL types.
2. Number of times I wanted to drop `nft_asset` in favor of `resource` type however in GraphQL it did not align because `TokenAsset` had ref link to `Metadata` corresponding to it, while `Resource` did not had one (at least semantically it meant different thing).

Gozala: Also worth noting that `status` field for `nft_asset` is larger enum because we do parse it while in case of resource we do not.

packages/database_v2/migrations/default/1631657122677_init/up.sql
```diff
@@ -0,0 +1,316 @@
...
+    resource_uri TEXT NOT NULL,
+
+    updated_at TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    inserted_at TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    PRIMARY KEY (content_cid, resource_uri)
+);
+
+CREATE TABLE IF NOT EXISTS content (
+    cid TEXT NOT NULL,
+    dag_size INTEGER,
```
alanshaw, line 266:
```suggestion
    dag_size BIGINT,
```

`INTEGER` would only allow us to store ~2GB max.

packages/database_v2/migrations/default/1631657122677_init/up.sql
```diff
@@ -0,0 +1,316 @@
...
+
+    updated_at TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
alanshaw, line 268: Does this differ from ours?:

```suggestion
    updated_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
```

packages/database_v2/artifacts/init-tables.sql
```diff
@@ -0,0 +1,316 @@
...
+DROP TYPE IF EXISTS pin_service CASCADE;
+CREATE TYPE pin_service AS ENUM (
+    'Pinata',
+    'IpfsCluster'
+);
+
+-- A blochain block identified by a it's hash.
+CREATE TABLE IF NOT EXISTS blockchain_block (
+    hash TEXT NOT NULL,
+    number INTEGER NOT NULL,
```
alanshaw, line 134: I imagine this needs to be `BIGINT` too?

packages/database_v2/migrations/default/1631657122677_init/up.sql
```diff
@@ -0,0 +1,316 @@
...
+    PRIMARY KEY (cid)
+);
+
+
+CREATE TABLE IF NOT EXISTS pin (
+    id BIGSERIAL NOT NULL,
+    content_cid TEXT NOT NULL,
+    service pin_service NOT NULL,
+    status pin_status NOT NULL,
+    status_text TEXT,
```
alanshaw, line 279: We don't have this field...

alanshaw (changes requested):

packages/database_v2/artifacts/init-tables.sql
Gozala: I'm not sure where exactly this type is defined by the thegraph.com but this table corresponds to following type

Screenshot 2021-09-17 at 1 55 46 PM

So I assumed it would be correspond to `INTEGER` here

Gozala: And according to [subgraph docs](https://thegraph.com/docs/developer/create-subgraph-hosted#graphql-supported-scalars) (quoting here)

> | Type | Description |
> |-- | --|
> |Bytes | Byte array, represented as a hexadecimal string. Commonly used for Ethereum hashes and addresses.|
> |ID | Stored as a string.|
> |String | Scalar for string values. Null characters are not supported and are automatically removed.|
> |Boolean | Scalar for boolean values.|
> |Int | The GraphQL spec defines Int to have size of 32 bytes.|
> |BitInt | Large integers. Used for Ethereum's uint32, int64, uint64, ..., uint256 types. Note: Everything below uint32, such as int32, uint24 or int8 is represented as i32.|
> |BigDecimal | BigDecimal  High precision decimals represented as a signficand and an exponent.  The exponent range is from −6143 to +6144. Rounded to 34 significant  digits.|

`Int` is 32 bytes which is wheres according to postgres docs `INTEGER` is only 4 bytes and `BIGINT` is 8 bytes. So it seems that even `BIGINT` may not be adequate here.

I think we can roll with `BIGINT` and migrate to another scalar if we do indeed reach the limit.

packages/database_v2/migrations/default/1631657122677_init/up.sql
Gozala: I can remove this, but I found it useful to have `status` and ` status_text` in all tables where records change their state. That way `status` is a variant from enum and `status_text` can contain more details like stack traces or whatever have you

packages/database_v2/migrations/default/1631657122677_init/up.sql
the-simian: I've converted every date to explicitly specify 'utc' so there's not any issues with timezones, schemas will match now

packages/database_v2/migrations/default/1631657122677_init/up.sql
the-simian: I just removed this with the INT->BIGINT and utc changes

packages/database_v2/artifacts/init-tables.sql
the-simian: Agree on the BIGINT, just changed everything to reflect this

PR #412, opened 2021-09-10
feat: heroku infra

alanshaw:
* 2x Postgres for prod/staging
* 1x replica for prod
* 2x PostgREST for prod/staging
* DB credentials for PostgREST (`web_anon`, `authenticator`, `nft_storage`)
* RO DB credentials for `dagcargo`
* Custom domain for DB: `db.nft.storage` and `db-staging.nft.storage`
* SSL for both domains


packages/infra/heroku/grant-dagcargo.sql
```diff
@@ -0,0 +1,2 @@
+GRANT USAGE ON SCHEMA public TO dagcargo;
+GRANT SELECT ON public.aggregate, public.aggregate_entry, public.auth_key, public.content, public.deal, public.pin, public.upload, public.user TO dagcargo;
```
olizilla, line 2: Could simplify this by granting read on all tables in the public schema, and for all new table too
```suggestion
GRANT SELECT ON ALL TABLES IN SCHEMA public TO dagcargo;
ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO dagcargo;
```

olizilla: this is what i used for managing perms on fil-flow

packages/infra/heroku/grant-postgrest.sql
```diff
@@ -0,0 +1,6 @@
+-- Authenticator login
+GRANT web_anon TO authenticator;
+-- NFT.Storage user
+GRANT nft_storage TO authenticator;
+GRANT USAGE ON SCHEMA public TO nft_storage;
+GRANT SELECT, UPDATE, DELETE ON public.aggregate, public.aggregate_entry, public.auth_key, public.content, public.deal, public.pin, public.upload, public.user TO nft_storage;
```
olizilla, line 6:
```suggestion
GRANT SELECT, UPDATE, DELETE ON ON ALL TABLES IN SCHEMA public TO nft_storage;
ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, UPDATE, DELETE ON TABLES TO nft_storage;
```

olizilla: also, could consider not allowing DELETE on any tables from apps users, to ensure all deleting is done as an update to add a deleted flag.

olizilla (approved):

packages/infra/heroku/grant-postgrest.sql
alanshaw: Can you explain what this is doing?

+1 on not allowing delete.

olizilla: My understanding is:
- `GRANT x ON ALL TABLES` means "on all tables in the schema now".
- altering the default privileges for the user means they get access by default to new tables created in the future as well.

the docs are not helpful on the first point, but the this is useful for ALTER DEFAULT PRIVILEGES https://www.postgresql.org/docs/current/sql-alterdefaultprivileges.html

...have added comments in the suggestion. I've not run this script, so this is more of an FYI / reminding you of things we have wrestled in the past. see: https://github.com/filecoin-shipyard/fil-flow/issues/19#issuecomment-722491223

alanshaw: Tweaked to remove syntax error and `DELETE` privilege:

```suggestion
-- allow access to all tables in the public schema currently
GRANT SELECT, UPDATE ON ALL TABLES IN SCHEMA public TO nft_storage;
-- allow access to new tables that are created in the public schema in the future
ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT, UPDATE ON TABLES TO nft_storage;
```

hugomrdias (approved):

PR #418, opened 2021-09-13
feat: migration deals schema

hugomrdias:
This PR adds cargo fdws and deals schema.


packages/api/db/cargo.sql
```diff
@@ -0,0 +1,36 @@
...
+
+-- Create materialized view from cargo "aggregate_entries" table
+CREATE MATERIALIZED VIEW public.aggregate_entry 
+    AS SELECT * FROM cargo.aggregate_entries;
+
+-- Indexes for "aggregate_entries" mat view
+CREATE UNIQUE INDEX aggregate_entry_unique_cidv1_aggregate_cid
+    ON public.aggregate_entry (cid_v1, aggregate_cid);
+CREATE INDEX aggregate_entry_cid_v1
+    ON public.aggregate_entry (cid_v1);
```
ribasushi, line 17: Don't do this:
index (`A`, `B`)
index (`A`)

Instead do
index (`B`,`A`)
index (`A`)

This way searching for both `A` and `B` is properly indexed for you

packages/api/scripts/cli.js
```diff
@@ -80,4 +84,59 @@ prog
...
+
+CREATE SERVER dag_cargo_server
+  FOREIGN DATA WRAPPER postgres_fdw
+  OPTIONS (host '${process.env.DAG_CARGO_HOST}', dbname '${process.env.DAG_CARGO_DATABASE}', fetch_size '200000');
+
+CREATE USER MAPPING FOR current_user
+  SERVER dag_cargo_server
+  OPTIONS (user '${process.env.DAG_CARGO_USER}', password '${process.env.DAG_CARGO_PASSWORD}');
+    
+    `
```
alanshaw, line 136: Please can this just be part of the `cargo.sql` file and just do some simple variable substitution (e.g. `$DAG_CARGO_HOST`)? If we need to change this I'd like to not have to hunt around for this part of the SQL. We also get syntax highlighting and linting in the editor if it's in a `.sql` file.

packages/api/test/nfts-pins-v1.spec.js
```diff
@@ -0,0 +1,75 @@
...
+      body: JSON.stringify({
+        cid: 'bafybeia3enxye7slrio4ode3kdzq74o6u7wdyytjcw5cwawwkhjt2l5ks4',
+        name: 'test-file11',
+      }),
+    })
+    assert(res, 'Server responded')
+    assert(res.ok, 'Server response ok')
+    const { ok, value } = await res.json()
+  })
+})
```
alanshaw, line 75: Is this relevant to this PR?

alanshaw (changes requested):

packages/api/scripts/cli.js
hugomrdias: imo its useful to do it like this, because it keeps the sql files runnable. 
i would do this change later when we have a proper sql migration system that support psql variable like sqitch.

alanshaw (approved):
LGTM

PR #433, opened 2021-09-16
feat: handle errors in the files page

### LINKED ISSUE #404: Incomplete uploads in Web UI should give user an error
Similar to on the Web3.Storage side, there is no error message displayed in the web app UI if an upload does not finish and an incomplete DAG is uploaded to Cluster. We should put in an error message.

rafaelramalho19:
## Alert Component 👀 
 - ➕ Added an "Alert" component that allows messages to be shown on the top/bottom of the page.
 - ⏲️ This component has a built-in timer that animates it when appearing/leaving.
 - 💬 This component design details is being discussed in the public discussion (Discord/Slack at the #nft-storage channel)

Closes #404

## Demo 👀 

https://user-images.githubusercontent.com/24696635/133646543-e11098c0-6bcc-42b4-a9c3-6f36de10fb8e.mov

### DIFF
--- a/packages/website/components/alert.js
+++ b/packages/website/components/alert.js
@@ -1,0 +1,58 @@
+import React, { useState } from 'react'
+import clsx from 'clsx'
+
+/**
+ * @typedef {Object} AlertProps
+ * @prop {string} [className]
+ * @prop {('top'|'bottom')} [position]
+ * @prop {('error'|'success'|'info')} [type]
+ * @prop {Number} [timer] How much seconds before onTimerEnd should be called
+ * @prop {Function} [onTimerEnd]
+ * @prop {JSX.Element} [children]
+ */
+
+const ANIMATION_DURATION = 400
+
+/**
+ *
+ * @param {AlertProps} props
+ * @returns
+ */
+export default function Alert({
+  className,
+  position = 'bottom',
+  type = 'info',
+  timer,
+  onTimerEnd,
+  children,
+}) {
+  const [hasTimerEnded, setTimerHasEnded] = useState(false)
+
+  if (timer && onTimerEnd) {
+    setTimeout(() => {
+      setTimerHasEnded(true)
+      setTimeout(onTimerEnd, ANIMATION_DURATION)
+    }, timer * 1000)
+  }
+
+  return (
+    <div
+      className={clsx(
+        className,
+        'flex justify-center items-center',
+        'fixed left-0 right-0',
+        position === 'bottom' &&
+          hasTimerEnded &&
+          'bottom-0 disappear-to-bottom',
+        position === 'bottom' &&
+          !hasTimerEnded &&
+          'bottom-0 appear-from-bottom',
+        position === 'top' && 'top-0',
+        type === 'error' && 'bg-red',
+        type === 'success' && 'bg-green'
+      )}
+    >
+      {children}
+    </div>
+  )
+}
--- a/packages/website/components/alert.js
+++ b/packages/website/components/alert.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import clsx from 'clsx'
 
 /**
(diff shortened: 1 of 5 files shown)


packages/website/components/alert.js
```diff
@@ -0,0 +1,58 @@
...
+  onTimerEnd,
+  children,
+}) {
+  const [hasTimerEnded, setTimerHasEnded] = useState(false)
+
+  if (timer && onTimerEnd) {
+    setTimeout(() => {
+      setTimerHasEnded(true)
+      setTimeout(onTimerEnd, ANIMATION_DURATION)
+    }, timer * 1000)
```
alanshaw, line 35: Shouldn't this be done in `useEffect` or something similar?

packages/website/pages/new-file.js
```diff
@@ -67,12 +70,17 @@ export default function NewFile() {
...
           },
         })
-      } catch (err) {
+
+        router.push({ pathname: '/files', query: version ? { version } : null })
+        setError('')
+      } catch (/** @type any */ err) {
         console.error(err)
+        setError(
+          `Error uploading (${err.message}). Check the console for more details`
```
alanshaw, line 79: There's no more info that's useful for the user:

```suggestion
          `Error uploading: ${err.message}`
```

packages/website/components/alert.js
```diff
@@ -0,0 +1,58 @@
...
+const ANIMATION_DURATION = 400
+
+/**
+ *
+ * @param {AlertProps} props
+ * @returns
+ */
+export default function Alert({
+  className,
+  position = 'bottom',
```
alanshaw, line 23: Can it be from the top?...cover the nav and make dismissable.

alanshaw (changes requested):

packages/website/components/alert.js
rafaelramalho19: Not necessarily, since this doesn't affect any variables inside the component or affect the render in any way. But I'll change it to use `useEffect` to make it more reliable to future iterations :)

packages/website/components/alert.js
rafaelramalho19: Sure, it's not changed here, this is the component general definition, but I'll make the specific error we're adding here from the top :)

rafaelramalho19: image
image

Looks good @alanshaw ?

packages/website/pages/new-file.js
```diff
@@ -176,6 +182,25 @@ export default function NewFile() {
             </div>
           </form>
         </Box>
+        <When condition={error !== ''}>
+          <Alert
+            className="pa4 white"
+            position="top"
+            type="error"
+            timer={4}
+            onTimerEnd={() => setError('')}
```
alanshaw, line 191: Can this one not auto dismiss please? I want users to retry the upload or explicitly accept the issue and then move on. For large uploads a user may leave the page and then come back and have missed the error message.

packages/website/pages/new-file.js
```diff
@@ -67,12 +70,15 @@ export default function NewFile() {
             setPercentComplete(Math.round((totalBytesSent / car.size) * 100))
           },
         })
-      } catch (err) {
+
+        router.push({ pathname: '/files', query: version ? { version } : null })
+        setError('')
```
alanshaw, line 75: Can we clear the error before we start uploading, so that on retry we don't have a stale error message?

packages/website/pages/new-file.js
```diff
@@ -176,6 +182,25 @@ export default function NewFile() {
...
+            timer={4}
+            onTimerEnd={() => setError('')}
+          >
+            <>
+              {error}{' '}
+              <button
+                className="border ml2 br-100 b--transparent pointer"
+                onClick={() => setError('')}
+              >
+                X
```
alanshaw, line 199: Would be good if this were a proper cross graphic and not an X...🙏

alanshaw (changes requested):

packages/website/pages/new-file.js
rafaelramalho19: Sure, it now makes sense since we don't autoclear it :)

packages/website/pages/new-file.js
rafaelramalho19: We already had a cross icon 😅

alanshaw (approved):

PR #436, opened 2021-09-16
Fix/dedupe-website

hugomrdias:
- Dedupe website
- Remove `version` from client


packages/api/src/bindings.d.ts
```diff
@@ -37,6 +37,10 @@ export interface RouteContext {
   params: Record<string, string>
 }
 
+export interface CronContext {
+  sentry: Toucan
+}
```
alanshaw, line 42: We don't have any crons in CF anymore right...can we just remove?

alanshaw (approved):

PR #442, opened 2021-09-17
niftysave postgres migration

### LINKED ISSUE #381: Import niftysave into a new db
Once #378 is ready and deployed we'll need to import data from current DB. Which could be accomplished with a same strategy as migration from schema with 'status' fields  to schema with queues as implemented in the following script:

https://github.com/ipfs-shipyard/nft.storage/blob/98bf9c726b90001fe959f141b0f0e66f878b8a31/packages/database/src/migrations/2021-08-18-token-asset-queue.js#L43-L62

Gozala:
This pull request builds upon #407 that does following:

1. Adds `niftysave_migration` table for maintaining cursors into individual fauna collections.
   - That way migration can resume where it fall of in case of early termination
2. Adds high level functions to `migrate.js` that does the most heavy lifting in terms of
    - Resuming (or starting) migration from the  cursor state
    - Delegates to individual collection specific function that performs migration
    - Provides `writeMigrationState` that performs all inserts & updates cursor.
3. Adds `migration/content.js` file that utilizes `migrate.js` to migrate `Content` collection
4. Adds `yarn migrate-content` script meant to be run form CRON.

Idea is to create collection / table specific `migration/${name}.js` modules that will take batch of records from fauna and invoke `writeMigrationState` with a corresponding (sub)mutation to insert those into postgres thorugh hasura.

P.S: So far I have been running all this on local hasuara / postgres setup.

List of migration scripts:

- [x] Content -> content
- [x] Block -> blockchain_blocks @the-simian 
- [x] block_tokens -> nfts_by_blockchain_blocks
- [x] Cursor -> ???
- [x] ERC721ImportResult -> erc721_import
- [x] eRC721ImportResult_tokens -> erc721_import_by_nft
- [x] ~~Analyzed -> null~~
- [x] ~~FailedAnalyze -> null~~
- [x] ~~log -> null~~
- [x] Metadata -> nft_metadata
- [x] metadata_assets -> other_nft_resources
- [x] Metric -> niftysave_migration
- [x] ~~migrations -> null~~
- [x] Owner -> nft_owner
- [x] Pin -> pin
- [x] PinLocation -> pin_location
- [x] Resource -> resource
- [x] ~~ScheduledAnalyze -> null~~
- [x] Token -> nft
- [x] TokenAsset -> nft_asset
- [x] TokenContract -> blockchain_contract

Fixes #381


packages/niftysave/src/migrate/nft-asset.js
```diff
@@ -0,0 +1,83 @@
...
+ *   ipfs_url: string | null
+ *   status: import('../../gen/db/schema').TokenAssetStatus,
+ *   status_text: string
+ *   updated_at: import('faunadb').values.FaunaTime
+ * }} Asset
+ *
+ * @param {Migration.Document<Asset>} document
+ */
+const insert = ({
+  ts,
```
Gozala, line 49:
```suggestion
```

packages/niftysave/src/migrate/content.js
```diff
@@ -0,0 +1,35 @@
...
+ *    created?: import('faunadb').values.FaunaTime
+ *    dagSize?: number
+ * }} Content
+ *
+ * @param {Migration.Document<Content>} doc
+ */
+const insert = ({ data }) => ({
+  cid: data.cid,
+  dag_size: data.dagSize || null,
+  inserted_at: (data.created || { date: new Date() }).date.toUTCString(),
```
Gozala, line 15:
```suggestion
  inserted_at: (data.created || { date: new Date() }).date,
```

packages/niftysave/src/migrate/erc721-import.js
```diff
@@ -0,0 +1,35 @@
...
+})
+
+/**
+ * @param {Migration.Document<ERC721ImportResult>[]} documents
+ * @returns {Migration.Mutation}
+ */
+export const mutation = (documents) => ({
+  insert_erc721_import: [
+    {
+      objects: documents.map(insert),
```
Gozala, line 24: @the-simian should we let it crash or should we ignore conflicts here ?

the-simian: let it crash because we shouldn't ever go back in time/cursor, and log the exception imo

the-simian (approved):
This looks good, I think the next step is to get the migrations scripts actually running now that we've tested them all independently in docker

packages/niftysave/src/migrate/blockchain-block.js
```diff
@@ -26,8 +26,7 @@ export const mutation = (blocks) => ({
       objects: blocks.map(insert),
       on_constraint: {
         on_constraint:
```
the-simian: needs to be `constraint`

PR #456, opened 2021-09-22
fix: change conventional build script 

the-simian:
to build-all to prevent CI conflict on service slike heroku


package.json
```diff
@@ -8,7 +8,7 @@
     "test:client": "yarn --cwd packages/client test",
     "test:api": "yarn --cwd packages/api test",
     "test:website": "yarn --cwd packages/website test",
-    "build": "run-s build:*",
+    "build-all": "run-s build:*",
```
Gozala, line 11:
```suggestion
```

PR #462, opened 2021-09-23
feat: add exponentian backoff to mitigate ratelimits 

Gozala:


the-simian (approved):
This all looks good to me. should keep things alive and self-repairing as well!

PR #463, opened 2021-09-23
fix: remove usage_model from top level

alanshaw:
* `usage_model` cannot be overridden if set at top level and for dev we shouldn't need to setup a paid account.
* Added `DATABASE_URL` in the right places in the config (vars are not inherited).
* Split out vars in mine and @gozala's sections to be consistent with the staging/prod config.
* Removed `type = "webpack"` from mine and @gozala's sections so we inherit `type = "javascript"`.

Useful info on config: https://developers.cloudflare.com/workers/cli-wrangler/configuration

### DIFF
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -79,24 +79,29 @@ Open `wrangler.toml` and add an env for yourself (replacing "USER" with your nam
 
 ```toml
 [env.USER]
-type = "webpack"
 name = "nft-storage-USER"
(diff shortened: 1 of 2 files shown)


zebateira:
this is great!
maybe also update the docs? https://github.com/ipfs-shipyard/nft.storage/tree/main/packages/api#development-setup

hugomrdias (approved):

PR #465, opened 2021-09-23
Fix/459

### LINKED ISSUE #459: change status type to enum
'PinError', 'PinQueued', 'Pinned', 'Pinning'
make view to return the response enum

hugomrdias:
- changes status type to new enum
- extract cli cmds
- move db types to cli
- generate types from local docker

based on #454

closes #459


packages/api/db/tables.sql
```diff
@@ -63,6 +63,26 @@ CREATE TABLE IF NOT EXISTS pin
...
+           WHEN status = 'Pinning'
+               THEN 'pinning'
+           WHEN status = 'PinError'
+               THEN 'failed'
+           END status,
+       content_cid,
+       service,
+       inserted_at,
+       updated_at
+FROM pin;
```
alanshaw, line 84: What is the rationale for this being done in the database? We have `db-transforms.js`, which is doing lots of transformations, why not keep them all grouped together there? It doesn't feel right to me to put this particular transform in the database but have the rest in `db-transforms.js`.

packages/api/scripts/cli.js
```diff
@@ -6,13 +6,11 @@ import { fileURLToPath } from 'url'
...
-import pg from 'pg'
-import fs from 'fs'
-import execa from 'execa'
-
-const { Client } = pg
 // @ts-ignore
 import git from 'git-rev-sync'
+import { dbCmd } from './cmds/db.js'
+import { dbSqlCmd } from './cmds/db-sql.js'
+import { dbTypesCmd } from './cmds/db-types.js'
```
alanshaw, line 13: 👌

alanshaw (changes requested):

packages/api/db/tables.sql
hugomrdias: This is an experiment to see how well this approach fits with everything else, i was hoping to understand if we could use a schema isolation setup like the one described here https://postgrest.org/en/v8.0/schema_structure.html#schema-isolation. 
This would allow for more flexibility in the internal tables without breaking the api code because the API would only use psql functions and views, we can be more granular with permissions for views and tables etc. Basically it would give us more options for future changes in the schema with very low overhead.

I would like to move more stuff into postgres and simplify the api on anything related to the data going in and out of postgres isolating it all inside postgres. Some of this stuff would be easier and safer to do inside postgres than in the API because of the type safety, consistency features already builtin. 

Another example of this, that will be in another PR im working on, is triggers to bump `update_at` columns instead of having that logic in multiple places in the API code we can just have one trigger per table that always does that.

alanshaw: Ok gotcha. I'm wary of leaning too much into postgres. Our team skillset lies in javascript predominantly so we should play to that strength. The only difference between FQL and SQL is sligltly more familiarity within the team. For example, right now, I cannot remember the syntax to write a function in SQL and I imagine that to be true of other members. We don't want to slow down our development pace again like we did with FQL. There are certainly instances where it is beneficial to move code into the DB, but IMHO it should be the exception not the rule.

The secondary consideration here is testing, I'm unclear on the testing and coverage story for code in the database. I would wager that we, as a team are better at writing and testing JS code than SQL and that the tools we have in JS are better at helping us pinpoint problems or understand where we're not testing code paths.

I think in this instance it would be better to keep it simple by grouping related concerns together i.e. map the enums in `db-transforms.js`. I would like to get version 1 of this DB migration out the door and if we feel we need to complicate it later by adding an abstraction layer in the database then we can do that, but we should discuss the pros and cons before.

Re: triggers, my learned wisdom has been to **not** use triggers, since they are invisble side effects. You'll know about them when you write them but then they're transparent, and easily forgotten. They seem like a great idea until you have one special case where you _do_ want to update a document but _don't_ want to set the `updated_at` column.

hugomrdias: ok i will revert it to the js side!

alanshaw (approved):

PR #466, opened 2021-09-23
feat: maintenance mode

alanshaw:
This PR adds middleware to prevent writes or prevent reads _and_ writes to the API when maintenance mode is enabled.

To enable, you simply set an environment variable `MAINTENANCE_MODE` to one of the following values

- `--` = no reading or writing
- `r-` = read only mode
- `rw` = read and write (normal operation)

The `/version` endpoint has been updated to return the current maintenance mode so that a banner can be shown in the UI.

It allows us to perform **migrations** where we need to prevent writes to the database/cluster for the duration and as well as giving us a way of preventing writes to cluster/the db when **outages** occur.

Screenshot 2021-09-28 at 13 52 15


packages/api/src/middleware/maintenance.js
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * -- = no reading or writing
+ * r- = read only mode
+ * rw = read and write (normal operation)
+ *
+ * @param {() => Mode} getter
+ */
+export function setMaintenanceModeGetter(getter) {
+  getMaintenanceMode = getter
+}
```
hugomrdias, line 51: this can't mutate DEFAULT_MODE directly ?

hugomrdias (changes requested):

packages/api/src/middleware/maintenance.js
alanshaw: I don't understand the question...

hugomrdias (approved):

PR #471, opened 2021-09-27
fix: db config documentation and test fixes

alanshaw:


packages/api/pw-test.config.cjs
```diff
@@ -56,8 +56,7 @@ module.exports = {
       stdout.toString().includes('Server started on: http://localhost:9094')
     ) {
       console.log('⚡️ Mock cluster started.')
-      await execa(cli, ['db', '--start', '--project', project])
-      await delay(2000)
+      await execa(cli, ['db', '--init', '--start', '--project', project])
```
alanshaw, line 59: Added `--init` here - not sure how things were working in CI before...

hugomrdias: start seems to be enough run everything need, did it fail for you locally ?

hugomrdias: start seems to run everything need, did it fail for you locally ?

hugomrdias (approved):

packages/api/pw-test.config.cjs
alanshaw: Ah seems to work now if I remove it so I'll revert.

PR #476, opened 2021-09-27
Fix/niftysave schema

Gozala:
As per conversation on Friday we have encountered problem in data migration because we encountered token records that store large SVG file source into `tokenURI` field. In addition modern day browsers impose no size limit on data urls, so we may encounter same problems with `resoruce` table because metadata file may contain data url links that are too big.

This pull request addresses this by changing schema such that

- `nft` table no longer stores `token_uri` but stores `token_uri_hash` instead. migration also populates table by deriving hashes from `token_uri` column.
- `nft_asset` table is added generated `token_uri_hash` column which derives value from `token_uri`.
   > Note `token_uri_hash` will be a foreign key in `nft` table but that is intentionally left out now as we'll add foreign keys after migration is complete.
- `resource` table was added new generated `uri_hash` column which derives value from `uri`.
- `other_nft_resources` switched from `resource_uri` to `resource_uri_hash`.
- `nft_metadata` switched from `image_uri` to `image_uri_hash`

Updated diagram can be viewed at https://dbdiagram.io/d/613fb365825b5b0146004a76


packages/database_v2/migrations/default/1632668838948_uri-hash/up.sql
```diff
@@ -0,0 +1,103 @@
...
+
+--- NFT
+
+-- Add token_uri_hash field
+ALTER TABLE nft
+    ADD COLUMN token_uri_hash bytea;
+
+-- Populate column with hashes
+UPDATE nft
+    SET token_uri_hash = sha256(token_uri::bytea);
```
Gozala, line 29: @ribasushi this seems to fail with

```
sql_error_code = 22P02 ERROR:  invalid input syntax for type bytea
```

Which if this is correct https://stackoverflow.com/questions/36063777/postgresql-9-5-syntax-error is due to the fact that some `token_uri` contains invalid escape which I'd guess is one of those svgs that have `<\image>`.


I clearly don't seem to know the right way to turn text into bytea to hash it, mind educating me on this ?

ribasushi: Fml, I never had to do this myself.
Instead of `sometext::BYTEA` you apparently need `CONVERT_TO( sometext, 'SQL_ASCII' )`
If it doesn't like the chosen encoding - go with something making sense in your context ( `LATIN1` or similar )

Gozala (commented):
fix: change niftysave schema to use uri hashes as primary keys

PR #482, opened 2021-09-29
Abstraction over uri_hash

Gozala:
This pull request attempts to provide a functionality to abstract over `*_uri_hash` usage by:

1. Exposing postgres [updatable views](https://www.postgresql.org/docs/9.5/sql-createview.html#SQL-CREATEVIEW-UPDATABLE-VIEWS), which works around https://github.com/hasura/graphql-engine/issues/7602 and https://github.com/hasura/graphql-engine/issues/4633
2. Defines postgres stored procedures exposed as GraphQL mutations to insert into tables that have `*_uri_hash` but do not have `*_uri`.

image

P.S.: I would like to update migrations scripts to make use of this before merging it.


packages/database_v2/migrations/default/1632897573948_ingest-functions/up.sql
```diff
@@ -0,0 +1,150 @@
...
+SELECT
+    *
+FROM
+    other_nft_resources
+WHERE
+    other_nft_resources.content_cid = inserted_cid;
+
+END;
+
+$ function $;
```
the-simian, line 109: Heads up, while checking this out, I had to change this to `$$` (and also the others) to actually run for me

packages/database_v2/migrations/default/1632897573948_ingest-functions/up.sql
```diff
@@ -0,0 +1,150 @@
...
+
+CREATE FUNCTION add_nft_metadata (
```
the-simian, line 111: `CREATE OR REPLACE FUNCTION`

packages/database_v2/migrations/default/1632897573948_ingest-functions/up.sql
the-simian: `$function$`

...it was the spaces.

the-simian: Actually now I see how they got introduced, it is in the function definition screen, weird but NBD

the-simian: Ok tested and works great, still need to work out how to do a batch of them though

packages/database_v2/migrations/default/1632897573948_ingest-functions/up.sql
Gozala: > `CREATE OR REPLACE FUNCTION`

I don't like `create or replace` in migrations, if you had a function that you're trying to replace and you did not know migration should fail.

Gozala:
@the-simian I think it would be a really good idea to change our stored procedures so that input parameters refer to table columns as opposed to named types. As per https://www.postgresql.org/docs/9.1/sql-createfunction.html 

>   The data type(s) of the function's arguments (optionally schema-qualified), if any. The argument types can be base, composite, or domain types, or can reference the type of a table column.
>
>    Depending on the implementation language it might also be allowed to specify "pseudotypes" such as cstring. Pseudotypes indicate that the actual argument type is either incompletely specified, or outside the set of ordinary SQL data types.
>
>    The type of a column is referenced by writing **table_name.column_name%TYPE**. Using this feature can sometimes help make a function independent of changes to the definition of a table.
default_expr

the-simian:
That's a good idea here's the relevant docs for v 13 https://www.postgresql.org/docs/current/sql-createfunction.html

the-simian:
Sound good to me, here's the most recent docs, there are some minor changes: https://www.postgresql.org/docs/current/sql-createfunction.html

PR #485, opened 2021-09-29
refactor: rename account table to user

### LINKED ISSUE #374: Rename `accounts` table to `user`

alanshaw:
resolves #374

"user" needs to be quoted or namespaced in SQL because user is a reserved word in the language. Justification for doing this anyway:

1. We refer to "user" everywhere else in the JS code and we go to the trouble of ["renaming" fields from "account" to "user" in our queries](https://github.com/ipfs-shipyard/nft.storage/pull/485/files#diff-d2f63c0f47f000a546d34f098e45e9981f20eca7b8fccc3cfb35c95ab24fba47L81) and we have to [construct queries like `account_id: userId`](https://github.com/ipfs-shipyard/nft.storage/pull/485/files#diff-d2f63c0f47f000a546d34f098e45e9981f20eca7b8fccc3cfb35c95ab24fba47L81) because of it (confusing).
1. We actually don't need to quote or namespace prefix "user" when querying via postgREST and we don't often query directly in SQL (only in 1 stored procedure).
1. In my head I also find it difficult to square off one user having multiple "accounts". We don't support this right now but "account" suggests that we do.
1. IMO it is worth a small inconvenience on something we don't often do for an overall more consistent and less confusing code base.


packages/website/components/navbar.js
```diff
@@ -75,11 +83,14 @@ export default function Navbar({ bgColor = 'bg-nsorange', user }) {
...
+  const onMobileLinkClick = useCallback(
+    (event) => {
+      onLinkClick(event)
 
-    toggleMenu()
-  }, [onLinkClick, toggleMenu])
+      toggleMenu()
+    },
+    [onLinkClick, toggleMenu]
+  )
```
alanshaw, line 93: Prettier probably made this change...

alanshaw (commented):
@hugomrdias `db-types.d.ts` was **not** re-generated, the change was done manually. I get a _very_ different file if I use the CLI tool to generate.

hugomrdias (approved):

PR #491, opened 2021-09-29
Feat/461-473

hugomrdias:


alanshaw (commented):
* How are we dealing with new users/nfts that are created after a sync?
* How do we sync a deleted NFT?
* How do we deal with the ordering of inserts to postgres and our foreign key constraints? Is it this?:
    * Insert users
    * Insert auth keys
    * Insert content (normalized from `PINS`)
    * Insert pins (derived from `PINS`)
    * Insert uploads (`NFTS`)
* What are the tasks that have to be done during the downtime? i.e. after the initial sync has been done, but more data has arrived
    * What time estimate do we have for this?
* How are we syncing the `PINS` table? I see no code for this...
* Can you confirm these are no longer needed:
    * `NFTS_IDX` can just query `upload where user_id = X order by inserted_at DESC` etc.
    * `FOLLOWUPS` we can just query `pins where service = 'IpfsCluster' AND status <> Pinned`
    * `DEALS` is FDW & materialized views
    * `METRICS` query postgres directly for now?
    * `PINATA_QUEUE`  we can just query `pins where service = 'Pinata' AND status <> Pinned`

packages/tools/crons/pinsv1.js
```diff
@@ -0,0 +1,86 @@
...
+    return console.log(upsertError)
+  }
+
+  console.log(`
+Overall there's ${count} pins to be processed.
+In this run, ${pins.length} were processed and ${updatedPins.length} updated to new status.
+`)
+}
+
+run()
```
alanshaw, line 86: This can be removed. I ported it to https://github.com/ipfs-shipyard/nft.storage/pull/496

alanshaw (approved):

PR #503, opened 2021-10-01
feat: implement ingest_erc721_token endpoint

Gozala:
This change introduces `ingest_erc721_token` SQL function and corresponding GraphQL mutation as a replacement of former [`importERC721`](https://github.com/ipfs-shipyard/nft.storage/blob/47634bf6e49710846fff1fc4235b7ec7f14918c8/packages/database/fauna/resources/Function/importERC721.fql). Unlike former it does not work with batches instead it ingests one nft at a time.

This change deprecate `erc721_import` and `erc721_import_by_nft` tables. Going forward [`ingest.js`](https://github.com/ipfs-shipyard/nft.storage/blob/47634bf6e49710846fff1fc4235b7ec7f14918c8/packages/niftysave/src/ingest.js) task will:

1. Obtain current cursor by taking `nft.id` with largest `inserted_at`.
2. Will enqueue pulled `nfts` from the subgraph to writable end of transform stream.
3. Have concurrent task that pulls each nft record from the readable end of transform stream and write it to db via `ingest_erc721_token`.

### DIFF
--- a/packages/database_v2/metadata/databases/default/functions/functions.yaml
+++ b/packages/database_v2/metadata/databases/default/functions/functions.yaml
@@ -1,3 +1,4 @@
 - "!include public_add_nft.yaml"
 - "!include public_add_nft_metadata.yaml"
 - "!include public_add_other_nft_resource.yaml"
+- "!include public_ingest_erc721_token.yaml"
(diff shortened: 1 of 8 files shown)


the-simian:
Currently: been working off this on #505 and its working well

PR #517, opened 2021-10-04
Feat/niftysave analyzer rewire

Gozala:


the-simian:
Thanks for pushing this so I can see the TransformStream implementation

PR #522, opened 2021-10-05
fix: fix db error and avoid duplicate cids

hugomrdias:
Moves db error into the client to make it reusable
Avoids duplicate cidv0/v1 in the push pipeline for content and pins

### DIFF
--- a/packages/tools/cf-sync/push-to-db.js
+++ b/packages/tools/cf-sync/push-to-db.js
@@ -114,7 +114,7 @@ async function nftByUser(issuer, sub, ctx, task, keys, dbUser) {
  * @param {Context} ctx
  * @param {any[]} values
  * @param {definitions['auth_key'][]} keys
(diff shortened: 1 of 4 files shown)


packages/tools/cf-sync/push-to-db.js
```diff
@@ -126,32 +126,41 @@ async function addNFTs(values, keys, ctx, dbUser) {
...
   for (const { data, size, pinStatus } of values) {
-    contentArray.push({
-      cid: data.cid,
-      dag_size: size,
-      inserted_at: data.created,
-      updated_at: data.created,
-    })
+    const parsedCID = parseCid(data.cid)
+    // avoid duplicate cidv1 in content and pin
+    if (!cids.includes(parsedCID.contentCid)) {
```
alanshaw, line 137: You could use a `Set`!

alanshaw (approved):
LGTM

PR #542, opened 2021-10-07
feat: add events KV to temporarily record user and NFT create/delete events

alanshaw:
This adds a KV and a few lines to code to record an event when user:create, nft:create or nft:delete happen.

It'll allow us to "catch up" with the current state of the world much quicker than trawling everything again.

Events are keyed by ISO string and event data is stored in KV metadata. This is so that when listing keys they arrive in date order ASC and we don't have to query each key separately for the event data.


alanshaw:
Superseded by https://github.com/ipfs-shipyard/nft.storage/pull/554

PR #545, opened 2021-10-07
fix: add CORS headers to /version endpoint

alanshaw:
Adds CORS headers to `/version` so that it can be called from the browser.


alanshaw:
@jdiogopeixoto can you pull this into https://github.com/ipfs-shipyard/nft.storage/pull/544 and verify it works? 🙏

PR #505, opened 2021-10-01
WIP: Feat/update erc721 ingestor

the-simian:
ERC721 Import spike: 
still in progress, not really ready for review. I is currently in a working state (it reads, it writes) However, it still needs:
- Change out the "sink and store" recursive functions for a proper Typed Buffer
- typing
- refactor the config into the project's pattern so it reads `yargs`, etc.
- pull in the new stuff on #503 to test this.

PR #564, opened 2021-10-09
Remove fauna

Gozala:
This pull request depends on #505 as it is only remaining module that uses fauna, which is why ts will complain and that is expected. This PR can land after that one is merged.


ribasushi:
> Remove fauna

image

PR #570, opened 2021-10-09
feat: add service worker

rafaelramalho19:
Creating this PR as a part of the solution for https://github.com/ipfs-shipyard/nft.storage/issues/474

## Question:

The current cache solution is [First network, then fallback to cache](https://developers.google.com/web/tools/workbox/modules/workbox-strategies?hl=en#network_first_network_falling_back_to_cache) 

image

Should we change it?


olizilla:
That sounds good to me!

dchoi27:
Closing for now, can pick this up later if we want

Issue #355, opened 2021-08-31
Port fix for rejection of empty CAR files


alanshaw:
https://github.com/web3-storage/web3.storage/pull/344

JeffLowe:
related to #390 per Alan

dchoi27:
Yeah more precisely #645 will close this

PR #588, opened 2021-10-12
feat!: automatic client side CAR chunking for large data

alanshaw:
`storeBlob`, `storeDirectory` and `store` now construct CAR files and then call `storeCar` (which `POST`s to `/upload`). This automatically gives them CAR chunking capability.

It adds the following **static** functions:

```ts
NFTStorage.encodeBlob(blob: Blob): Promise<{ cid: CID, car: CarReader }>

NFTStorage.encodeDirectory(files: Blob[]): Promise<{ cid: CID, car: CarReader }>

NFTStorage.encodeNFT<T extends TokenInput>(input: T): Promise<{ cid: CID, token: Token<T>, car: CarReader }>
```

After encoding your CAR and obtaining the root CID you can call:

```js
await client.storeCar(car)
```

🚨 There are trade offs here:

1. We're always sending CAR files so our `type` field is going to always be `Car` for uploads from the JS client.

    To mitigate this we could simply inspect the root node of the CAR, from this we can assertain the type of the data. We should talk about our `type` field and what it means. Right now we've mapped it to the method of upload...lets resolve this and submit a separate PR to fix.

    FYI, I'm going to be implementing [#355](https://github.com/ipfs-shipyard/nft.storage/issues/355) soon so will be inspecting the root node of the CAR anyway for validation purposes.

2. The `files` property is not set for the `Multipart` or `Nft` type uploads (since it's being uploaded as a CAR).

    I actually can't see any requests to the `/status/:cid` API (literally 0 - everyone uses `/check/:cid`) in the cloudflare logs which is the only place this data is exposed to users. I don't think folks will miss it. However if we inspect the root node we can get a shallow directory listing to put here.

    For `Nft` types we can't really set it anymore. There's also a regression right now where we're _not_ setting it anyway. As an aside I'm not sure how much value it has since the data is just a list of file names, without paths within the object that is created...

resolves https://github.com/ipfs-shipyard/nft.storage/issues/220


packages/client/src/lib.js
```diff
@@ -222,7 +222,7 @@ class NFTStorage {
     validateERC1155(input)
     const blockstore = new Blockstore()
     try {
-      const token = await Token.encode(input, blockstore)
+      const token = await Token.Token.fromTokenInput(input, { blockstore })
       onRootCidReady && onRootCidReady(token.ipnft)
```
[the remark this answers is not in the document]
alanshaw: Ok so I'm clear, you're saying `storeBlob`, `storeDirectory` and `store` are **high level** and `storeCar` is **low level** - right?

I'm happy to _not_ expose `onRootCidReady, onStoredChunk, maxRetries` from those high level APIs but retain `onStoredChunk, maxRetries` as options to `storeCar`.

In this world we'd have the following _static_ functions to encode data as CARs that all return `{ cid: CID, car: BlockstoreCarReader }`:

* `encodeNFT` - encodes an object to a CAR
* `encodeBlob` - encodes a blob to a CAR
* `encodeDirectory` - encodes a directory of files to a CAR

By default these static functions would use an in memory blockstore, but would have an optional `{ blockstore: Blockstore }` option to allow the user to pass a filesystem/levelDB backed blockstore for BIG data.

So then you could use any one of those static functions to do the two step:

```js
const { cid, car } = await NFTStorage.encodeNFT({ ... })
await NFTStorage.storeCar(service, car, options)
```

The existing `storeBlob`, `storeDirectory` and `store` methods will call the appropriate static functions.

Does that sound reasonable? I hope so - I'm going to update the PR accordingly.

packages/client/src/lib.js
```diff
@@ -266,6 +238,63 @@ class NFTStorage {
...
+   * @param {Blob} blob
+   * @returns {Promise<{ cid: CID, car: import('./lib/interface.js').CarReader }>}
+   */
+  static async encodeBlob(blob) {
+    if (blob.size === 0) {
+      throw new Error('Content size is 0, make sure to provide some content')
+    }
+    const blockstore = new Blockstore()
+    const { root: cid } = await pack({
+      // @ts-ignore
```
Gozala, line 260: Nit: please add some comment on why we have to ignore

packages/client/src/lib.js
```diff
@@ -266,6 +238,63 @@ class NFTStorage {
...
+      size += file.size
+    }
+
+    if (size === 0) {
+      throw new Error(
+        'Total size of files should exceed 0, make sure to provide some content'
+      )
+    }
+
+    const blockstore = new Blockstore()
```
Gozala, line 287: NIt: It's second time I see following 7 lines of code only diff is `wrapWithDirectory` flag. I think it would be a good idea to factor it out to a function that just gives back `cid` and `car`.

packages/client/src/lib.js
```diff
@@ -266,6 +238,63 @@ class NFTStorage {
     }
   }
 
+  /**
+   * @template {API.TokenInput} T
+   * @param {API.Encoded<T, [[Blob, Blob]]>} input
```
Gozala, line 243: I think this is an error (which probably confused the type checker) `input` here supposed to be `T`, that is basically anything as long as it's compatible with `API.TokenInput`. 

Type `API.Encoded<T, [[Blob, Blob]]>` supposed to represent `T` in which all `Blobs` are substituted with `Blobs` which in theory should be the same `T`.

Gozala (changes requested):
Hey sorry for the delay for some reason it was not showing up in github notifications so I had to remember to do this. 

I think this revision looks great, only problem is bunch of functions switched from `T` input for metadata to  `API.Encoded<T, [[Blob, Blob]]>` which I think is the mistake & will also cause API docs to become confusing as it will no longer be clear what users needs to pass.

If this change was motivated by trouble with type checker, please instead do localized `ts-ignore` and feel free to assign me a followup issue to resolve it.

I marked this as "request changes" but I think once types are changed back this can land without another review around.

Thanks again for taking over this

alanshaw:
@alanshaw reminder to update https://github.com/nftstorage/nft.storage/issues/730 when this merges.

PR #638, opened 2021-10-20
refactor: get DB client from route context

alanshaw:
This PR is a refactor that changes the way routes obtain the DB client. Instead of it being returned from the `validate` function in `auth-v1.js` it is attached to the `RouteContext` so every route has access to it so they don't need to instantiate their own if they don't use `validate`.

The `Router` is now agnostic to the route context type, mandating only that it extends a `BasicRouteContext` which contains the URL parameters. It's constructor now takes a function that creates the context object as a parameter.

There is concrete `RouteContext` and `Handler<RouteContext>` types in `bindings.d.ts` which are used by our routes. The handler conforms to the `Handler<T extends BasicRouteContext>` type required by the router.

I've done this for consistency - I will add an s3 client to the `RouteContext` next.

### DIFF
--- a/packages/api/src/bindings.d.ts
+++ b/packages/api/src/bindings.d.ts
@@ -2,6 +2,7 @@ export {}
 
 import Toucan from 'toucan-js'
 import { Mode } from './middleware/maintenance.js'
+import { DBClient } from './utils/db-client.js'
 
 declare global {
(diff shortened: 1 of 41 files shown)


hugomrdias (approved):
LGTM just need a rebase

PR #693, opened 2021-10-26
chore: add /store endpoint to API schema

### LINKED ISSUE #490: Specify what file types are allowed in HTTP API docs
In the JS client docs, we specify that files referenced in the metadata have to adhere to ERC-1155:

However, we don't have anything in the HTTP API docs specifying this:

We should get more specific in the API docs!

EDIT: Realizing now that following ERC-1155 might just be a JS client only thing. However, the comment below on MP4s still holds.

yusefnapora:
I started to add a note about content types to the API docs, but it seems that the `/store` endpoint wasn't in the schema, just `/upload`. 

So this adds docs for the `/store` endpoint based on what I could glean from the code. Someone should probably look it over to make sure it's correct.

May need some formatting tweaks and so on - I wasn't able to run the website build locally, since my env isn't set up, so I haven't seen it rendered yet.

Also closes #490 - we actually allow all file types now, but I added a recommendation to only use images for the `image` field.

### DIFF
--- a/packages/website/public/schema.yml
+++ b/packages/website/public/schema.yml
@@ -9,6 +9,64 @@ servers:
 tags:
   - name: NFT Storage
 paths:
+  /store:
+    post:
+      tags:
+        - NFT Storage
+      summary: Store an ERC-1155 compatible NFT
+      description: |
+        Store an [ERC-1155](https://eips.ethereum.org/EIPS/eip-1155)-compatible NFT as 
+        a collection of content-addressed objects connected by IPFS CID links.
+
+        The POST request accepts `multipart/form-data` content, which must contain a form field named `meta`.
+
+        The `meta` field must contain a JSON string that conforms to the [ERC-1155 metadata schema](https://eips.ethereum.org/EIPS/eip-1155#metadata).
+
+        Any field inside the `meta` object whose value is set to `null` will be treated as a file reference, 
+        and you must supply a form data field containing the file content as a binary string.
+
+        The name of the form data field containing the file content should be the "path" of the JSON field, using `.` to traverse nested
+        objects.
+
+        For example, with a `meta` object of the form:
+
+          ```json
+          {
+            "name": "Hello",
+            "image": null,
+            "properties": {
+              "videoClip": null
+            }
+          }
+          ```
+
+        You must include form fields named `image` and `properties.videoClip` in your request body, with the content of the image and video files as the form field values.
+
+        All form fields other than `meta` must contain binary file data, and the field name must match a key in the `meta` JSON object whose value is `null`.
+
+        ### Mime Type Recommendations
+
+        The ERC-1155 metadata standard specifies that the `image` metadata field should reference a file with a content type of `image/*`.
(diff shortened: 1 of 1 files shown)


packages/website/public/schema.yml
```diff
@@ -9,6 +9,64 @@ servers:
...
+      summary: Store an ERC-1155 compatible NFT
+      description: |
+        Store an [ERC-1155](https://eips.ethereum.org/EIPS/eip-1155)-compatible NFT as 
+        a collection of content-addressed objects connected by IPFS CID links.
+
+        The POST request accepts `multipart/form-data` content, which must contain a form field named `meta`.
+
+        The `meta` field must contain a JSON string that conforms to the [ERC-1155 metadata schema](https://eips.ethereum.org/EIPS/eip-1155#metadata).
+
+        Any field inside the `meta` object whose value is set to `null` will be treated as a file reference,
```
Gozala, line 25: I think this is bit misleading,  and I would phrase it more like this

```suggestion
        Any field(s) inside the `meta` object can be substituted with an ipfs URL to a file(s), by providing a form data field with a name matching a (`.` delimited) property path and value containing the file content (in binary string or plain text depending on file format).
```

packages/website/public/schema.yml
```diff
@@ -9,6 +9,64 @@ servers:
...
+            "image": null,
+            "properties": {
+              "videoClip": null
+            }
+          }
+          ```
+
+        You must include form fields named `image` and `properties.videoClip` in your request body, with the content of the image and video files as the form field values.
+
+        All form fields other than `meta` must contain binary file data, and the field name must match a key in the `meta` JSON object whose value is `null`.
```
Gozala, line 45: I don't think we have `null` requirement, or if we do we probably should drop that. Maybe instead we should say if `value` been substituted is non falsy request will be rejected with error.

Gozala (approved):
Few suggestions to simplify things a bit, otherwise looks great! Thanks for doing this.

packages/website/public/schema.yml
```diff
@@ -9,6 +9,64 @@ servers:
...
+        objects.
+
+        For example, with a `meta` object of the form:
+
+          ```json
+          {
+            "name": "Hello",
+            "image": null,
+            "properties": {
+              "videoClip": null
```
alanshaw, line 38:
```suggestion
            "image": undefined,
            "properties": {
              "videoClip": undefined
```

Maybe use `undefined` as the value here, so that `JSON.stringify()` does not include it in the output e.g.:

```js
JSON.stringify({ foo: undefined })
"{}"
JSON.stringify({ foo: undefined })
"{\"foo\":null}"
```

packages/website/public/schema.yml
alanshaw: Yeah it's the multipart parts that determine the _files_ that are included in the metadata not the `meta` object.

packages/website/public/schema.yml
alanshaw: > I don't think we have `null` requirement

I don't believe we do either.

alanshaw (approved):
Please include @Gozala's changes before merging. Otherwise this is great.

PR #696, opened 2021-10-27
feat: add doc comments & include interface.ts in typedoc

yusefnapora:
I had a go at improving the typedoc output a bit. Now it includes the type definitions from `src/lib/interface.ts`, so you can click through to things like the API interface.

Speaking of which, the `@implements` tag on the `NFTStorage` class was `API.Service`, but it should be `API.API`, since `API.Service` is the `{ token, endpoint }` context object. When I switched it to `@implements API.API`, I got a bunch of type errors since it got confused by the static methods having the same name as the instance methods.

So this changes the `API.API` interface definition and splits it into two: `InstanceAPI` has the instance methods of the `NFTStorage` class while `StaticAPI` has the static methods. Now you can do `@implements API.InstanceAPI` in the jsdoc and tsc is happy.

### DIFF
--- a/packages/client/src/lib/interface.ts
+++ b/packages/client/src/lib/interface.ts
@@ -1,10 +1,24 @@
+/**
+ * @module types
+ */
+
 import type { CID } from 'multiformats'
+/**
+ * A Content Identifier for some binary content.
+ *
+ * See https://github.com/multiformats/js-multiformats for details, or
+ * the [IPFS docs](https://docs.ipfs.io/concepts/content-addressing/) for a
+ * conceptual overview of CIDs.
+ */
 export type { CID }
 
 import type { BlockDecoder } from 'multiformats/block'
+/** An interface for decoding blocks of binary data. See https://github.com/multiformats/js-multiformats for details. */
 export type { BlockDecoder }
 
 import type { CarReader } from '@ipld/car/api'
+
+/** An interface for reading CAR data. See https://github.com/ipld/js-car for details. */
 export type { CarReader }
 
 /**
@@ -12,12 +26,22 @@ export type { CarReader }
  */
 export type Tagged<T, Tag> = T & { tag?: Tag }
 
+/**
+ * Service context information used for operations requiring authorization.
+ */
 export interface Service {
+  /** Root URL of service API. */
   endpoint: URL
+
+  /** An authentication token provided by the service at the given endpoint. */
   token: string
 }
 
+/**
+ * Service context information for operations that do not require authorization.
+ */
 export interface PublicService {
+  /** Root URL of service API. */
   endpoint: URL
 }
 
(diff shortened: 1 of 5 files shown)


Gozala:
> Speaking of which, the @implements tag on the NFTStorage class was API.Service, but it should be API.API, since API.Service is the { token, endpoint } context object. 

This was intentional, idea is that `Client` instance is just glorified (with bunch of methods) service endpoint, yet you could use it as an argument to static methods. It probably needs to implement both interfaces.

packages/client/src/lib/interface.ts
```diff
@@ -26,7 +50,13 @@ export interface PublicService {
  */
 export type CIDString = Tagged<string, CID>
 
-export interface API {
+/**
+ * The public interface for the static methods of the NFT.Storage client API. See {@link NFTStorage} for the
+ * implementation.
+ */
+export interface StaticAPI {
+  new (options: { token: string; endpoint?: URL }): InstanceAPI
```
Gozala, line 58:
```suggestion
```

Can we not have this ? Idea behind API (now StaticAPI) is that other component is that:

1. Alternative compatible API can be implemented by other party. Requiring that it also be class is unnecessary complication.
2. Component can request API implementation to be passed to it so it can exercise it. Requirement of it being a class that constructs non static instance is unnecessary burden to satisfy. If components withes non static API it should probably just require that as an argument instead no matter how it was instantiated.

packages/client/src/lib/interface.ts
```diff
@@ -96,12 +112,93 @@ export interface API {
   check(service: PublicService, cid: string): Promise<CheckResult>
 }
 
+/**
+ * The public API for the instance methods of the NFT.Storage client. See {@link NFTStorage} for the implementation.
+ */
+export interface InstanceAPI {
```
Gozala, line 118: Maybe we should just call it `NFTStorageClient` instead ?

packages/client/src/lib/interface.ts
Gozala: If you want ensure our class is compatible in that case I would just break out this into separate `InstanceFactory` interface and do:

```js
/**
 * Just to verify API compatibility.
 * @type {API.StaticAPI & API.InstanceFactory}
 */
const api = NFTStorage
```

Gozala (approved):
This is awesome! Thanks and can't wait to see improved API docs on our site!

P.S.: Made some suggestions, but feel free to act or disregard as you see fit.

hugomrdias:
i did a bunch of changes here https://github.com/nftstorage/nft.storage/pull/714 around this topic we should sync on this because it seems we have 2 different approachs

dchoi27:
@yusefnapora please merge!

dchoi27:
@yusefnapora @hugomrdias did you have a chance to sync? we should try to close soon :)

dchoi27:
Closing per @yusefnapora's request! Major things got taken care of

PR #714, opened 2021-10-31
feat: remove kv code, update deps and improve client docs

hugomrdias:


alanshaw (changes requested):
Hard to review as it looks like so much change. I think this may have been easier to review as a commit to delete and then a commit to rename. It would be more accurate history but actually it'll probably be easier in the future to view the history as you have done it.

Tests are passing so I assume all ok. Great to remove all the old code.

Please can you update the `README.md` in `api` to remove the KV store instructions?

Other things I have on my list that need to be done (if not already):

* Delete KV stores in CF
* Remove the temporary worker we created for the migration
* Remove `migration_event` DB table (from SQL schema and staging/production DB)

hugomrdias:
Delete KV stores in CF
done

PR #715, opened 2021-11-01
Fixed log in store.html example

davidbrai:
The log function accepts just one parameter, so the data wasn't actually printed.
This commit fixes it


alanshaw (approved):

alanshaw:
Thanks @davidbrai 😀

PR #723, opened 2021-11-04
fix: survey button not opening modal

### LINKED ISSUE #707: Survey button not opening modal

Codigo-Fuentes:
Production only bug; could not reproduce locally;
Pr tests removing <When> wrapper around typeform button
if this works, it will close #707 

**update: after build fix this is working on preview deploy here: https://18bee2e3.nft-storage-1at.pages.dev/files/

Per Ze's comment on #707 , the only difference I can see between working code here: https://github.com/web3-storage/web3.storage/pull/574 and the code in nft. is the <When> wrapper. Let's try removing it.

If this doesn't work we have other options on deck to try:

1. Dynamically loading some elements / components using Next dynamic imports or useEffect
2. Using the React Embed Library from typeform https://developer.typeform.com/embed/react/

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -39,8 +39,9 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node }}
-      - uses: bahmutov/npm-install@v1
+      - run: bahmutov/npm-install@v1
         with:
+          useLockFile: false
           working-directory: packages/client
       - name: Test (ES)
         run: yarn --cwd packages/client test:es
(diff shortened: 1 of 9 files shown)


JeffLowe:
Based on P0 status of this, we assigned review to you @alanshaw 
But let us know if you prefer any other teammember do the review.

the-simian:
@Codigo-Fuentes I gave this a quick review, it looks good to me

the-simian (approved):

the-simian:
To supplement my "thumbs up" here's a gif of it working correctly when I checked it
it works

Issue #689, opened 2021-10-26
Niftysave: Massively parallelize ingestion and and analysis, binning procs by time


the-simian:
This is to resolve the fact that after migrating off the previous data store we marked failed items as Queued so we'd retry them. There's currently several million items in Queue and we need to cover more ground very quickly. In addition some domains are unreliable, which slows analysis. Here's some short-term goals:

- [x]  1.  create several analysis procs
- [x] 2. bin these procs by time (probably by month) to cover missing ground
- [x] 5. have procs move at different rates. Some need to try once or twice and move on to not get jammed by problematic domains, others need to try more frequently. The goalis some scanners need to be moving faster and covering more ground, others more slowly and more aggressively trying failed/ Queued.

- [x] 7. create more ingest.js procs (these would be similar to the existing ingest script but start at specific time intervals

Ultimately after discussion, this is a short-term solution to quickly update the datastore. After we've caught up we might implement a more sophisticated and self-repiring pattern moving forward that doesn't require manual intervention

The following are ticketed elsewhere and added t the overall epic.  They will be explored as needed.
- [ ] 3. possibly have analysis action moving in reverse; that is from present time backwards. https://github.com/nftstorage/nft.storage/issues/747
- [ ] 4. target states to retry, this might include Failures https://github.com/nftstorage/nft.storage/issues/748
- [ ] 6. Disallow certain domains via config (this will eventurally be moved into the status JSON b (more detail in #521)

PR #729, opened 2021-11-05
Feat/time slicing analyzer

the-simian:
This has a working time-slicer on the analyzer, so we can run multiple analyzers in parallel (#689 )


the-simian:
the fix for https://github.com/nftstorage/nft.storage/issues/732 is included in this PR

PR #731, opened 2021-11-05
feat: add FAQ page

### LINKED ISSUE #555: Dedicated FAQ page
We should make a dedicated FAQ page on NFT.Storage where we can list an exhaustive list of FAQs. We can link from the top menu bar, as well as at the FAQs at the bottom of the main page.

- [x] add FAQs to top nav and bottom footer on main page
- [x] pull in FAQ content from main page and use as FAQ content https://nft.storage/
- [x] pull in additional FAQ content confirmed by @dchoi27 and/or @alanshaw

Codigo-Fuentes:
Closes #555 

- Adds dedicated FAQ page at /faq
- Redundantly retains faq on /
- Adds FAQ content per discussion @ https://github.com/nftstorage/nft.storage/issues/555#issuecomment-939074129
- Adds links to header and footer
- Fixes incorrect alignment of items in footer
- Adds InlineCode component used throughout new FAQ items

faq-page
inline-code
economics

Fixed 👇 
correctly aligned

previous 👇 
misaligned


dchoi27 (changes requested):
- Can we kill the image for the free question
- Capitalization consistency in using "NFT.Storage"
- Can we advertise the size restriction as 31GiB (and change this in the homepage's FAQ section as well)
- Might have missed in the code, but are we adding a link to the full FAQ section in the homepage's FAQ section?

Thanks!!!

Codigo-Fuentes:
@dchoi27 I made the first three changes. At present the faq on the home page is a redundant, identical copy of the content on the dedicated faq page at /faq (the same component is used for both)

There are links to that page in the header and footer. Do you still want another link? Where would you want that link? 

The heading of the faq section on homepage is currently a hash link. As we now have a dedicated page, we could remove that and have that heading link to the faq page (although if there are hash links like this: https://nft.storage/#faq in the wild we would probably want to handle them by redirecting to the /faq page)

Also, something like "read more..." doesn't make much sense at the moment since the content is identical on both pages.

Thoughts?

dchoi27:
Oh - didn't realize the content shown on the homepage is in its own file. Could we split it so the content we're showing on the homepage is just a subset of what's in the dedicated FAQ page? And then we could add something like "More FAQs" on the homepage.

One new thing - could add a paragraph to the answer to `I tried using an HTTP gateway to retrieve my content from IPFS but am receiving an HTTP error. Does this mean my content was not stored successfully on NFT.Storage?`:
```
Additionally, if the data was not stored on NFT.Storage, then there might be issues with the IPFS node(s) with a copy of the data providing that data to the gateway. Using NFT.Storage makes sure that the content stored is broadcasted to the network using best practices!
```

Thank you!

Codigo-Fuentes:
✅ copy addition in.
✅ abbreviated faq on homepage, links to full faq at bottom (pictured below)

more-faqs

dchoi27 (approved):
could we increase the size of the font for "More FAQs" at the bottom or differentiate in some other way? Otherwise looks good!

PR #740, opened 2021-11-08
feat: temp metaplex upload route

alanshaw:
WIP - not tested but maybe works?

PR #743, opened 2021-11-08
feat: validate metaplex jwt signature

yusefnapora:
Hey @alanshaw, here's my metaplex JWT validation code as a PR to your branch to make it easier to follow. 

I don't think we should merge it yet though - apparently CloudFlare workers have added [support for Ed25519 signatures to their WebCrypto implementation](https://developers.cloudflare.com/workers/runtime-apis/web-crypto#footnote%201). That's likely a lot faster than using the tweetnacl dependency I added to this commit, which is all JS and might blow the CPU budget in a worker.

I'll play with it a bit tomorrow and see if I can get it working with the CloudFlare flavored WebCrypto.

cc @hugomrdias


yusefnapora:
@alanshaw I rebased this onto `main` and changed the PR target to main, so we can merge this one directly instead of pulling it into #740. Seems simpler this way :)

packages/api/src/routes/metaplex-upload.js
```diff
@@ -0,0 +1,100 @@
...
+  const upload = await db.createUpload({
+    mime_type: blob.type,
+    type: 'Car',
+    content_cid: contentCid,
+    source_cid: sourceCid,
+    dag_size: dagSize,
+    user_id: user.id,
+    files: [],
+    key_id: key.id,
+    meta,
```
alanshaw, line 74: Please can we pull out the pieces we want from the parsed JWT so that folks cannot construct a token to put arbitrary data here?

alanshaw (changes requested):
Looks good so far.

packages/api/src/utils/jwt.js
```diff
@@ -6,6 +8,13 @@ const algorithms = {
       name: 'SHA-256',
     },
   },
+
+  // NODE-ED25519 exists only in the CloudFlare workers runtime API
+  // see: https://developers.cloudflare.com/workers/runtime-apis/web-crypto#footnote%201
+  'NODE-ED25519': {
```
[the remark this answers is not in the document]
yusefnapora: good call, thanks

packages/api/src/utils/jwt.js
```diff
@@ -100,6 +109,119 @@ export async function verifyJWT(token, secret, alg = 'HS256') {
...
+  if (!payload.iss) {
+    return false
+  }
+
+  const pubkey = await keyFromDID(payload.iss)
+  const sig = Base64URL.parse(tokenParts[2])
+  const headerPayload = utf8ToUint8Array(tokenParts[0] + '.' + tokenParts[1])
+
+  if (!crypto || !crypto.subtle) {
+    return verifyEd25519SignatureWithJSCrypto(headerPayload, sig, pubkey)
```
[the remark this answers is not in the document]
yusefnapora: Yes, exactly. I'll add some comments to make that more obvious.

yusefnapora:
Just rebased onto main to fix up the yarn.lock conflict.

packages/api/src/index.js
```diff
@@ -106,6 +108,9 @@ r.add('post', '/api/pins', psa(pinsAdd, RW), [postCors])
 r.add('post', '/api/pins/:requestid', psa(pinsReplace, RW), [postCors])
 r.add('delete', '/api/pins/:requestid', psa(pinsDelete, RW), [postCors])
 
+// Temporary Metaplex upload route, mapped to metaplex user account.
+r.add('post', '/metaplex/upload', withMode(metaplexUpload, RW), [postCors])
+
```
alanshaw, line 113: Can we move this to just above the `/internal` routes? It's in the middle of the legacy `/api/*` routes right now and I don't want it to accidentally be removed.

packages/api/src/routes/metaplex-upload.js
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * When >2.5MB, use local add, because waiting for blocks to be sent to
+ * other cluster nodes can take a long time. Replication to other nodes
+ * will be done async by bitswap instead.
+ */
+const LOCAL_ADD_THRESHOLD = 1024 * 1024 * 2.5
+
+// TODO: remove when https://github.com/nftstorage/nft.storage/pull/638 merged
+const db = new DBClient(database.url, secrets.database)
```
alanshaw, line 24: https://github.com/nftstorage/nft.storage/pull/638 was merged, you can obtain this instance from the `ctx` object.

packages/api/src/routes/metaplex-upload.js
```diff
@@ -0,0 +1,158 @@
...
+  const upload = await db.createUpload({
+    mime_type: blob.type,
+    type: 'Car',
+    content_cid: contentCid,
+    source_cid: sourceCid,
+    dag_size: dagSize,
+    user_id: user.id,
+    files: [],
+    key_id: key.id,
+    meta: { iss, req: JSON.stringify(req) },
```
alanshaw, line 92: Why stringify here? We cannot query the contents from Postgres if it's a string.

```suggestion
    meta: { iss, req: { put: { rootCID: req.put.rootCID } } },
```

...also just need to ensure we don't put anything else in the DB that we don't want there. Do you want to include `req.put.tags`? If yes can we validate it in any way?

packages/api/src/utils/ed25519.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/**
+ * Extracts an Ed25519 public key from a `did:key` decentralized identifier string.
+ *
+ * @param {string} did - a "did:key" formatted DID string
+ * @returns {Promise<Uint8Array>} - the decoded public key
+ * @throws if DID is invalid or does not contain a valid Ed25519 public key
+ */
+export async function keyFromDID(did) {
```
alanshaw, line 86: Function is not async.

packages/api/src/routes/metaplex-upload.js
```diff
@@ -0,0 +1,158 @@
...
+  const local = blob.size > LOCAL_ADD_THRESHOLD
+  const { cid, bytes: dagSize } = await cluster.addCar(blob, { local })
+  const { sourceCid, contentCid } = parseCid(cid)
+
+  // confirm that the content cid matches the cid from the signed metadata
+  if (!req.put || typeof req.put.rootCID !== 'string') {
+    throw new HTTPError('invalid token - no root cid in payload', 401)
+  }
+  const signedCid = req.put.rootCID
+  if (signedCid !== sourceCid && signedCid !== contentCid) {
```
alanshaw, line 79: `contentCid` is the normalised CID. Are you saying that `req.put.rootCID` will _always_ be a v1 base32 encoded CID? There's no way people can put `Qm...` CID's in the CAR?

packages/api/src/routes/metaplex-upload.js
```diff
@@ -0,0 +1,158 @@
...
+  return { user: { id: data.user.id }, key: { id: data.id } }
+}
+
+/**
+ * Verifies a self-signed JWT for a metaplex upload.
+ *
+ * @param {string} token - an encoded JWT token from an x-web3auth header.
+ * @returns {Promise<boolean>} - true if the token contains a valid solana public key and a valid signature from the key.
+ */
+async function verifyMetaplexJWT(token) {
```
alanshaw, line 128: I'd maybe rename this `parseMetaplexJWT` and have it return the `payload`. It also should do the verification and _always_ throw when something is wrong.

You can also ensure that `req.put.rootCID` exists in this same function.

Then you'd not need to parse the token twice and I think it'll make things easier to follow.

packages/api/test/metaplex-upload-v1.spec.js
```diff
@@ -0,0 +1,142 @@
+import assert from 'assert'
+import {
+  createTestUserWithFixedToken,
+  DBTestClient,
+  rawClient,
+} from './scripts/helpers.js'
+import { fixtures } from './scripts/fixtures.js'
+import { createCar } from './scripts/car.js'
+
+describe(' V1 - Metaplex Upload ', () => {
```
alanshaw, line 10:
```suggestion
describe('Metaplex Upload ', () => {
```

Also please remove "v1" from the filename. The v1 stuff will all be removed at some point.

alanshaw (changes requested):
Some minor changes but I think this is almost good to merge.

packages/api/src/routes/metaplex-upload.js
yusefnapora: hmm... good point. I should probably use `CID.parse` and `CID.equals` instead.

packages/api/src/routes/metaplex-upload.js
yusefnapora: good call, that seems much nicer

packages/api/src/routes/metaplex-upload.js
yusefnapora: I realized I can just use the same `parseCid` helper and compare `contentCid`s instead of `CID.parse`.

packages/api/src/routes/metaplex-upload.js
yusefnapora: I changed things just now to extract the `solanaCluster` from the tags and put it into metadata. That's the only bit we might need in the future, apart from the issuer & cid, I think. Now `meta` just has those three fields and ignores everything else.

yusefnapora:
hmm... looks like I broke something with the refactor. will fix now

yusefnapora:
A few stupid bugs crept in during the refactor - squashed now.

alanshaw (approved):
🚢

PR #745, opened 2021-11-09
In Progress: Feat/erc721 write-table

the-simian:
(This is still in progress, not ready for a review)
- adds a migration in hasura: a table with no fk relationships and no logging and the ability to use actual batch inserts for optimized write speed
- regenerates hasura zeus-graphql
- pulls the zeus script into a scripts dir
- small changes to the prettierrc so it doesn't try to format .sh/.sql files

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,6 +7,7 @@ worker
 site/public
 docs
 packages/niftysave/gen
+*.sql
 CHANGELOG.md
 .next
 .supabase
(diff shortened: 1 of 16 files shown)


the-simian:
references: https://github.com/nftstorage/nft.storage/issues/727

the-simian:
I am going to merge this so I can verify that it works in prod as it has worked locally. This does not change the ingest-script, this table still works independently of the existing table. Changes should be nonbreaking

Issue #362, opened 2021-09-01
Clean up styles for NFT.Storage API docs


dchoi27:
@alanshaw for these docs? https://ipfs-shipyard.github.io/nft.storage/client/

alanshaw:
No these ones: https://nft.storage/api-docs/

This is low priority but it would be good to stort things like text sizes, fonts and colours to be more in keeping with the rest of the site.

dchoi27:
@Codigo-Fuentes if you have a chance feel free to pick this up, these docs are super ugly

PR #709, opened 2021-10-29
Adds global styling to api doc section

msecret:
This adds a CSS file (I couldn't find a better way to customize swagger ui) to the api docs section so it uses the fonts, colors and some general styling from the rest of the site. This doesn't include all styling; for instance the double buttons require changing the code structure of swagger-ui, which is possible, but a more complex solution. It does bring it closer to looking like the rest of site.

To correctly set the global CSS so it overrides swagger's existing CSS I'm using double classes for each rule: `.swagger-ui.swagger-ui`. This ensures the styles will be overridden without !important specificity. This could be slightly improved with sass in the project because I could wrap all the rules in this once with it's nesting power.

I attempted to write this api-docs css file so it composes or extends the Tachyons classes vs writing actual CSS but no solution proved easy to implement with either CSS Modules or Sass.

Fixes #362.

PR #759, opened 2021-11-11
Feat/doc styles

the-simian:
styles the docs to look like the site, an accessible nft.storage swagger 4.x theme

1. This uses the material from (https://github.com/nftstorage/nft.storage/pull/709) , which I pulled into a branch from @msecret 's fork
2. Resolves https://github.com/nftstorage/nft.storage/issues/362
3. This replaces the swagger 4.x theme with an nft.storage theme, which is loaded in the _app.js, like any non-module CSS. (it must be done this way. Since this is our theme, there's no need to use any CSS hacks to override swagger, we just load this one file now
4. There are additional vars added to the `global.css` file in order to hydrate the swagger theme such as additional grayscale values, default border properties
5. This is more responsive than the previous theme, but I did not exhaustively check every known device (see screenshots below)
6. I added the class `nft-storage` at the root of the app. I ended up not using this as an CSS overriding strategy but I think its worth leaving.
7. General cleanup: Some spicier animations, and various sizing/spacing fixes (eg. previous docs screen wouldn't go the whole page when items were collapsed). Too many to list, so I included several before/after screenshots below


the-simian:
image

the-simian:
image

the-simian:
image

the-simian:
spicier

the-simian:
Before/After (items collapsed)
image

the-simian:
Before/After (modal)
image

the-simian:
Before/After (inside the opblocks)
image

the-simian:
Before/ After (Schemas)
image

(probably the least has changed here other than border/spacing, but mainly checking there's no regressions)

alanshaw (approved):
Thanks @msecret and @the-simian ❤️

PR #764, opened 2021-11-12
feat: update ToS

dchoi27:


packages/website/pages/terms.js
```diff
@@ -23,75 +23,158 @@ export default function TermsOfService() {
...
+        </p>
+        <p className="lh-copy">
+          All data stored on {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} is stored in long-term (18 month) deals 
+          with independent Filecoin Storage Providers. Filecoin storage deals can 
+          be renewed by anyone, are non-custodial, and anyone can access information 
+          about them (content address, duration and expiration, price, etc.) and 
+          verify the data is legitimately stored via the Filecoin blockchain.
```
Codigo-Fuentes, line 53: - suggestion: "...and verify **that** the data is legitimately..."

packages/website/pages/terms.js
```diff
@@ -23,75 +23,158 @@ export default function TermsOfService() {
...
           The following terms and conditions govern all use of the{' '}
           <a className="black" href="https://nft.storage">
-            nft.storage
+            NFT.Storage
           </a>{' '}
           website (the “Website”) and all content, services and products
-          available at or through the Website. The Website is offered subject to
-          your acceptance without modification of all of the terms and
-          conditions contained herein. As all data uploaded to nft.storage will
+          available at or through the Website. As all data uploaded to nft.storage will
```
Codigo-Fuentes, line 29: NFT.storage ?

packages/website/pages/terms.js
```diff
@@ -23,75 +23,158 @@ export default function TermsOfService() {
         <p className="lh-copy">
           The following terms and conditions govern all use of the{' '}
           <a className="black" href="https://nft.storage">
-            nft.storage
+            NFT.Storage
           </a>{' '}
           website (the “Website”) and all content, services and products
```
Codigo-Fuentes, line 28: - suggestion: "...and all content, services, and products" (consistent usage of oxford comma)

packages/website/pages/terms.js
```diff
@@ -23,75 +23,158 @@ export default function TermsOfService() {
...
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '}. {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} 
+          will do the following to make the transition for its users as seamless as possible: 
+          (1) give users an export of all data associated with their account, including 
+          the content addresses of this data and information about the Filecoin deals 
+          containing this data, (2) free renewal of Filecoin deals until an automated,
```
Codigo-Fuentes, line 135: - suggestion: ... (2) ... **provide** free renewal ... (parallel series)

packages/website/pages/terms.js
```diff
@@ -23,75 +23,158 @@ export default function TermsOfService() {
...
+          </a>{' '} is committed to ensuring that all NFTs on Filecoin have a 
+          perpetual storage solution. All Filecoin storage deals facilitated by 
+          {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} are automatically renewed in perpetuity by {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} before 
+          expiration. In late 2022, we plan to build tooling to enable the
```
alanshaw, line 75: Sounds a bit too too much like a "rough roadmap".

How about something like:

> NFT.Storage will switch to non-custodial, perpetual renewal of storage contracts independent of NFT.Storage to secure the perpetual persistence of all NFTs without dependence on any single entity when the tooling becomes available in the Filecoin network.

packages/website/pages/terms.js
```diff
@@ -23,75 +23,158 @@ export default function TermsOfService() {
...
+          using the correct CID. Users should not store any private or sensitive 
+          information in an unencrypted form using {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '}. Further, deleting 
+          files from {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} will remove them from the file listing for a user's 
+          account, but nodes on the IPFS network may retain copies of the data indefinitely.
```
alanshaw, line 112: ...and data aggregated into Filecoin deals will persist for the duration of the deal.

packages/website/pages/terms.js
```diff
@@ -23,75 +23,158 @@ export default function TermsOfService() {
...
+          </a>{' '} needs to be terminated, the service 
+          will provide at least 180 days notice to users. Per the Filecoin protocol, all 
+          existing Filecoin deals will continue, independent of {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '}. {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} 
+          will do the following to make the transition for its users as seamless as possible:
```
alanshaw, line 132: What "transition" is being referred to here? If the service needs to be terminated?

packages/website/pages/terms.js
```diff
@@ -23,75 +23,158 @@ export default function TermsOfService() {
...
-        <p className="lh-copy">
-          Deals may be made with Filecoin nodes operating on test network(s)
-          (i.e. not mainnet). These deals will be flagged as not being a mainnet
-          deal and will be lost on network resets. Node’s that are stored data
-          are not guaranteed to be accessible and may only be available via
-          private networks for testing purposes.
+            NFT.Storage
+          </a>{' '} intends to provide free, perpetual storage for NFTs. All data already 
+          uploaded to the service for free will continue to be persisted on the Filecoin 
+          network at no cost to the user until we have built a tool for perpetual,
```
alanshaw, line 167: Should not refer to "we" here IMHO.

> until a facility for perpetual, non-custodial storage is available in the Filecoin network.

Although "until" here suggests that it will not be "no cost" at this point.

alanshaw (changes requested):

packages/website/pages/terms.js
dchoi27: yeah agreed - it was a direct ask from Pooja to include a date, but will make it more subtle

packages/website/pages/terms.js
dchoi27: yeah. added a clause to clarify that this is referring to the hypothetical scenario in which the service had to be terminated

packages/website/pages/terms.js
dchoi27: yeah i was struggling with that exact point. took another stab - does that sound better?

packages/website/pages/terms.js
```diff
@@ -23,75 +23,161 @@ export default function TermsOfService() {
...
+          and guaranteed to be publicly available and verifiable via its unique 
+          content identifier.
+        </p>
+        <p className="lh-copy">
+          All data stored on {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} is stored in long-term (18 month) deals 
+          with independent Filecoin Storage Providers. Filecoin storage deals can 
+          be renewed by anyone and are non-custodial. Anyone can access information
```
alanshaw, line 54: > Filecoin storage deals can be renewed by anyone

Is this right? I know you can create a deal with another storage provider for the same data, but I wasn't aware that renewal in the sense we're alluding to here was possible yet?

packages/website/pages/terms.js
```diff
@@ -23,75 +23,161 @@ export default function TermsOfService() {
...
+        </p>
+        <p className="lh-copy">
+          All data stored on {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} is stored in long-term (18 month) deals 
+          with independent Filecoin Storage Providers. Filecoin storage deals can 
+          be renewed by anyone and are non-custodial. Anyone can access information 
+          about them (content address, duration, expiration, price, etc.) and 
+          verify that the data is legitimately stored via the Filecoin blockchain.
```
alanshaw, line 56:
```suggestion
          verify that the data is provably stored by the Filecoin blockchain. 
```

packages/website/pages/terms.js
```diff
@@ -23,75 +23,161 @@ export default function TermsOfService() {
...
+        <p className="lh-copy">
+          All data stored on {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} is stored in long-term (18 month) deals 
+          with independent Filecoin Storage Providers. Filecoin storage deals can 
+          be renewed by anyone and are non-custodial. Anyone can access information 
+          about them (content address, duration, expiration, price, etc.) and 
+          verify that the data is legitimately stored via the Filecoin blockchain. 
+          All Filecoin deals are{' '}
```
alanshaw, line 57:
```suggestion
          All Filecoin deals made by NFT.STorage are{' '}
```

packages/website/pages/terms.js
```diff
@@ -23,75 +23,161 @@ export default function TermsOfService() {
...
+          <a className="black" href="https://github.com/filecoin-project/filecoin-plus-client-onboarding">
+            Verified Deals
+          </a>{' '}
+          , which Filecoin Storage Providers are incentivized to accept at a lower 
+          price vs. other deals because they have a relatively higher increase in 
+          their chance to win block rewards.
+        </p>
+        <p className="lh-copy">
+          {' '}
```
alanshaw, line 66:
```suggestion
```

packages/website/pages/terms.js
```diff
@@ -23,75 +23,161 @@ export default function TermsOfService() {
...
+            NFT.Storage
+          </a>{' '} are automatically renewed in perpetuity by {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} before 
+          expiration. {' '}
+          <a className="black" href="https://nft.storage">
+            NFT.Storage
+          </a>{' '} will switch to non-custodial, perpetual renewal 
+          of storage contracts independent of the service to secure the perpetual
```
alanshaw, line 82: Double "perpetual"

alanshaw (changes requested):

packages/website/pages/terms.js
dchoi27: Making a new deal is the only way to renew whether you were the original dealmaker or not. So it's a matter of whether we want to use the term "renew" at all - IMHO the mechanism qualifies as renewal but open to discussion!

ribasushi: > Making a new deal is the only way to renew ...

Note: the above is true **today**. Unclear where we will be 1 year from now ( most deals would not have expired yet ).

dchoi27:
Closed with #822 taking precedence

PR #770, opened 2021-11-12
fix: cluster local add threshold

### LINKED ISSUE #660: Uploading a CAR file fails with "Error: 524: Origin Time-out"
I have been having trouble uploading a CAR file built from a folder that contains 1000 JSON files. This is not a big file (458KB) and I don't know why it never goes through.

Just to be clear, the code works for smaller CAR files (with like 5 items in a folder), but just stops working at certain point as I increase the number of files in the folder.

here's the code to replicate:

```javascript
require('dotenv').config()
const { CarIndexedReader } = require('@ipld/car')
const { NFTStorage, File } = require('nft.storage')
const token = process.env.NFT_STORAGE_KEY
const storage = new NFTStorage({ token })
const upload = async (folderPath) => {
  const carReader = await CarIndexedReader.fromFile(`${process.cwd()}/output.car`)
  const cid = await storage.storeCar(carReader)
  console.log("CID", cid)
}
(async () => {
  await upload("metadata2")
  process.exit()
})();
```

The `storage.storeCar()` method just hangs. I wait for about 10 minutes and it crashes with the following error:

```
Uploading...
/Users/a/ipfscar/node_modules/nft.storage/dist/src/lib.cjs:111
      throw new Error(result.error.message)
            ^

Error: 524: Origin Time-out
    at Function.storeBlob (/Users/a/ipfscar/node_modules/nft.storage/dist/src/lib.cjs:111:13)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async RetryOperation._fn (/Users/a/ipfscar/node_modules/p-retry/index.js:50:12) {
  attemptNumber: 6,
  retriesLeft: 0
}
```

I've also attached the car file I'm using (unzip the file before using):

[output.car.zip](https://github.com/nftstorage/nft.storage/files/7400800/output.car.zip)

Gozala:
Fixes #660

### DIFF
--- a/packages/api/src/routes-v1/nfts-upload.js
+++ b/packages/api/src/routes-v1/nfts-upload.js
@@ -5,9 +5,10 @@ import { validate } from '../utils/auth-v1.js'
 import { debug } from '../utils/debug.js'
 import { toNFTResponse } from '../utils/db-transforms.js'
 import { parseCid } from '../utils/utils.js'
+import * as constants from '../constants.js'
 
 const log = debug('nfts-upload')
-const LOCAL_ADD_THRESHOLD = 1024 * 1024 * 2.5
+const LOCAL_ADD_THRESHOLD = constants.cluster.localAddThreshold
 
 /**
  * @typedef {import('../bindings').NFT} NFT


Gozala:
@alanshaw it does not looks like it helped I still get the same outcome running this:

```sh
curl -i -X POST https://nft-storage-gozala.protocol-labs.workers.dev/upload --data-binary @/Users/gozala/Downloads/output.car -H "Content-Type: application/car"
HTTP/2 500 
date: Fri, 12 Nov 2021 20:20:31 GMT
content-type: application/json;charset=UTF-8
content-length: 70
access-control-allow-origin: *
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
report-to: {"endpoints":[{"url":"https:\/\/a.nel.cloudflare.com\/report\/v3?s=A3uXtdjAn32vcyHTRXJTlTEk3Q79822popr1ZBxVWnzhLGptB59PSURZKUil5ToWfaq8TPBFdlskeRPEmtlyUDFf6JKrmPs0jaR6lYKEkqUGee57OKFh4j%2FtcwdcYni1xPL5aJcO9mLbujN8HV%2BuhWDxj91K66c2PJ6ZlQp5HQ%3D%3D"}],"group":"cf-nel","max_age":604800}
nel: {"success_fraction":0,"report_to":"cf-nel","max_age":604800}
server: cloudflare
cf-ray: 6ad274a729ac609b-SEA
alt-svc: h3=":443"; ma=86400, h3-29=":443"; ma=86400, h3-28=":443"; ma=86400, h3-27=":443"; ma=86400

{"ok":false,"error":{"code":"Error","message":"504: Gateway Timeout"}}
```

And from wrangler tail I see that `local` is passed:

```
 | [Info] route https://nft-storage-gozala.protocol-labs.workers.dev/upload/ 
 | [Info] route to handler 
 | [Info] upload v1 
 | [Info] fake auth 
 | [Info] receiving blob 
 | [Info] adding to cluster as a car 
 | [Info] >> adding car to cluster with auth Basic c3RvcmFnZS1ib3Q6RnVVdjJEcFMyRGNUZXlaYg== & options {
  "local": true
} 
 | [Info] route failed Error: 504: Gateway Timeout 
 | [Debug] router Error: 504: Gateway Timeout
    at request (worker.js:38968:27)
    at async add (worker.js:38847:22)
    at async addCar (worker.js:39072:42)
    at async uploadV1 (worker.js:42879:52)
    at async Router.route (worker.js:35385:17) +0ms 
```

alanshaw:
Shall we merge and test against staging just incase?

alanshaw (approved):

Gozala:
> Shall we merge and test against staging just incase?

@alanshaw give me a moment, I'm trying to make sure it's not another regression in ipfs-client that fails to pass some params.

Gozala:
Looks like this is where request is going to `https://gozala-cluster-api-nft-storage.loca.lt/add?meta-size=457675&local=true&stream-channels=false&format=car&raw-leaves=true&cid-version=1`

Gozala:
aah it `lt` crashed which is what was going on 😖

Gozala:
Ok wrangler is complaining about

```
Error: Something went wrong with the request to Cloudflare...
No KV namespace found with ID 7e441603d1bc4d5a87f6cecb959018e4 [API code: 10041]
```

Which I think is because @hugomrdias was removing some namespaces, and may have removed this one as well

https://github.com/nftstorage/nft.storage/blob/82560e39a083b60d258a0115d9aa5b15a0bfb73f/packages/api/wrangler.toml#L12

Gozala:
I see https://github.com/nftstorage/nft.storage/pull/714 removed all the namespace refs and I'm guessing they were deleted on CF as well, so it may be better to land this after the other one.

Gozala:
I have merged this, because testing with my CF namespace proved it addresses the problem. Unfortunately deployment to staging also failed because `8964654a0a1d4f7586b59db90ef97594` namespace no longer seems to exist. I suspect this will be resolved once #714 lands. I also do not want to spend time figuring out namespaces since they will be removed by mentioned change anyway.

Issue #472, opened 2021-09-27
Improvements to initial setup for developers


alanshaw:
Fallout from https://github.com/ipfs-shipyard/nft.storage/pull/471

* Clean up docker containers properly. When my configuration was not setup properly containers weren't being cleaned up. This needs to be fixed.
  * In `pw-test.config.cjs`, if we throw in `beforeTests` then `afterTests` is not run
* Document somewhere that docker needs 8GB RAM for the postgres and postgrest docker containers to run in. I got "error 137" starting postgrest - this is because I only allowed 2GB of RAM for docker (I think maybe the default).
* Rename env vars (suggestions shown...). These are currently very generic and there was no documentation for the values they should contain:
  * `DATABASE_URL` => `POSTGREST_API_URL`
  * `DATABASE_TOKEN` => `POSTGREST_BASIC_AUTH_TOKEN`
  * `DATABASE_CONNECTION` => `POSTGRES_CONN_STRING`
* Add an npm script to setup a localtunnel for using local postgrest - the app needs to access a locally running database (or perhaps we should recommend setting up a free supabase account for this?).

PR #776, opened 2021-11-14
feat: add docker-compose and devcontainer.json for local dev env

yusefnapora:
I think I did it!

This adds a `.devcontainer` folder with a bunch of stuff to enable running all the dependent services using docker-compose.

If you're using VSCode, you should see something like this when you checkout this branch:
Screen Shot 2021-11-13 at 7 23 48 PM

Clicking "reopen in container" should hopefully spin up all the services defined in `.devcontainer/docker-compose.yml`, along with the DB services in `packages/api/docker/docker-compose.yml`, which is symlinked to `.devcontainer/docker-compose.db.yml`.

There's also a `.devcontainer/docker-compose.extend.yml` file that's a placeholder for things you might need to override for your local env. I'm using it because my main dev machine is an M1 mac and dealing with the arch issues is annoying, so I usually run docker stuff using a [docker context](https://docs.docker.com/engine/context/working-with-contexts/) that points to a nearby x86 machine. Remote contexts wont work with bind mounts from the local machine, so my solution is to clone the repo on the remote machine, then put this in my override compose file:

```yaml
services:
  dev:
    volumes:
      - /home/yusef/work/repos/nft.storage:/workspace
```

That way the checked out repo on the remote machine gets bind-mounted into the container, instead of trying to use the local checkout on my mac. This is slightly less convenient, since the remote copy doesn't automatically sync with the local one, but that could be solved with sshfs or something.

I also added a new command to `tools/cli.js` called `local-env-check` that tries to sanity check your `.env` file, make sure you're logged into `wrangler`, and check that all the required wrangler secrets are defined. It doesn't check everything, but it seems like a good start.

The local-env-check should run automatically when you attach to the container in vscode. If you don't have all the right vars, it should complain at you.

If everything's all good, it will tell you to run `node ./packages/api/scripts/cli.js db-sql --cargo --testing` to setup the local db schema.

If you're not using VSCode, you can run `.devcontainer/compose.sh`, which is a wrapper around `docker-compose` that targets the combined compose files. It accepts all the same args as `docker-compose`, so e.g.: `./.devcontainer/compose.sh up`.

Please kick the tires on this and let me know how it works for you!

### DIFF
--- a/packages/tools/utils/localenv.js
+++ b/packages/tools/utils/localenv.js
@@ -1,0 +1,256 @@
+import dotenv from 'dotenv'
+import path from 'path'
+import fs from 'fs/promises'
+import childProcess from 'child_process'
+import toml from 'toml'
+import { fileURLToPath } from 'url'
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url))
+const ENV_PATH = path.join(__dirname, '..', '..', '..', '.env')
+const API_DIR = path.join(__dirname, '..', '..', 'api')
+
+const REQUIRED_VARS = [
+  {
+    name: 'CF_ACCOUNT_ID',
+    desc:
+      'CloudFlare account id. Get from `wrangler whoami` after logging in with `wrangler login`',
+  },
+  {
+    name: 'CF_TOKEN',
+    desc:
+      'CloudFlare API token. Go to https://dash.cloudflare.com/profile/api-tokens and use CF Workers template.',
+  },
+  {
+    name: 'DATABASE_URL',
+    desc: 'URL of postgrest DB api for local development.',
+    defaultValue: `https://${process.env.DEV_USERNAME ||
+      process.env.USER}-postgrest-nft-storage.loca.lt`,
+  },
+  {
+    name: 'DATABASE_CONNECTION',
+    desc: 'Connection URI for local db.',
+    defaultValue: 'postgresql://postgres:postgres@localhost:5432/postgres',
+  },
+  {
+    name: 'DATABASE_TOKEN',
+    desc:
+      "Signed JWT for local db API. Use the default if you're using the devcontainer / docker-compose setup.",
+    defaultValue:
+      'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJyb2xlIjoicG9zdGdyZXMifQ.oM0SXF31Vs1nfwCaDxjlczE237KcNKhTpKEYxMX-jEU',
+  },
+  {
+    name: 'PINATA_JWT',
+    desc:
+      'Pinata API token. Go to  https://app.pinata.cloud/keys and make a key with all Pinning API permissions.',
+  },
+  {
+    name: 'DAG_CARGO_HOST',
+    desc:
+      'hostname / ip for dag-cargo database (read-only replica). Get from team 1Password if working on PL-managed instance.',
+  },
+  {
+    name: 'DAG_CARGO_DATABASE',
+    desc:
+      'hostname / ip for dag-cargo database (read-only replica). Get from team 1Password if working on PL-managed instance.',
+  },
+  {
+    name: 'DAG_CARGO_USER',
+    desc:
+      'username for dag-cargo database (read-only replica). Get from team 1Password if working on PL-managed instance.',
+  },
+  {
+    name: 'DAG_CARGO_PASSWORD',
+    desc:
+      'password for dag-cargo database (read-only replica). Get from team 1Password if working on PL-managed instance.',
+  },
+]
+
+const REQUIRED_WRANGLER_SECRETS = [
+  {
+    name: 'DATABASE_TOKEN',
+    desc:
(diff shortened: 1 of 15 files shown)


yusefnapora:
Just a heads up that I squashed everything down to one commit and removed my user config from `wrangler.toml` that slipped into this branch by mistake.

yusefnapora:
This is still having some problems with the cluster setup. I've realized that using a single cluster node isn't enough - it will respond to `/id`, but it needs to bootstrap to at least one other node before it enables the add endpoint.

@alanshaw you mentioned something about being able to run a single node cluster if you build from source. do you have a link to how that works handy? Or we can just spin up a second peer.

With a second peer, I can reach the add endpoint, but then I get errors when uploading CARs, either via the website or curl. I sprinkled some console logs around to get the error message from the cluster request, which looks like a cbor decoding thing:

```
invalid header: Invalid majorByte: 0xef
```

I eventually gave up yesterday and just pointed my dev env at the production cluster, which works fine. 

Anyway, that's the current status - will update as I figure out more.

JeffLowe:
@yusefnapora is this part/all of #472 ?

redaphid (commented):
Looking good!

The containers don't seem to be shutting down for me once vscode exits, though. Do they for you?

I'm missing a few env vars in my local dev. Not sure if that's related.

packages/tools/utils/localenv.js
```diff
@@ -0,0 +1,256 @@
...
+      console.error(
+        'Your config seems to have errors. Please address the issues printed above and try again'
+      )
+    }
+  }
+
+  static async checkWranglerIsAuthenticated() {
+    console.log('verifying that wrangler is authenticated...')
+    try {
+      childProcess.execSync('wrangler whoami')
```
redaphid, line 130: This is returning the user `node` for me. Then complains about not being able to find `/home/node/.wrangler/config/default.toml` :(

redaphid:
Other things I've run into: 

* The local env script flags missing optional env values as errors:
```
Your local .env file has some issues:
missing required environment variable DAG_CARGO_HOST: hostname / ip for dag-cargo database (read-only replica). Get from team 1Password if working on PL-managed instance.
missing required environment variable DAG_CARGO_DATABASE: hostname / ip for dag-cargo database (read-only replica). Get from team 1Password if working on PL-managed instance.
missing required environment variable DAG_CARGO_USER: username for dag-cargo database (read-only replica). Get from team 1Password if working on PL-managed instance.
missing required environment variable DAG_CARGO_PASSWORD: password for dag-cargo database (read-only replica). Get from team 1Password if working on PL-managed instance.
```

Whether or not I comment out the checks for these env vars, I can't curl the service @ 8787. Even after confirming the docker image is running, and looking at that port:
```❯ docker ps
CONTAINER ID   IMAGE                                                  COMMAND                  CREATED         STATUS         PORTS                                         NAMES
508c310fb979   nftstorage_devcontainer_postgrest-tunnel               "lt"                     3 seconds ago   Up 2 seconds                                                 nftstorage_devcontainer_postgrest-tunnel_1
daf2d664e642   nftstorage_devcontainer_ipfs-api-proxy-tunnel          "lt"                     3 seconds ago   Up 2 seconds                                                 nftstorage_devcontainer_ipfs-api-proxy-tunnel_1
409f5705640a   nftstorage_devcontainer_cluster-api-tunnel             "lt"                     3 seconds ago   Up 2 seconds                                                 nftstorage_devcontainer_cluster-api-tunnel_1
440cafddfd3f   postgrest/postgrest:v8.0.0                             "/bin/postgrest"         3 seconds ago   Up 2 seconds   0.0.0.0:3000->3000/tcp, :::3000->3000/tcp     nftstorage_devcontainer_rest_1
dba7622f1834   ipfs/ipfs-cluster:latest                               "/sbin/tini -- /usr/…"   3 seconds ago   Up 2 seconds   0.0.0.0:9094-9095->9094-9095/tcp, 9096/tcp    nftstorage_devcontainer_cluster0_1
a506a9139872   ipfs/go-ipfs:latest                                    "/sbin/tini -- /usr/…"   4 seconds ago   Up 3 seconds   4001/tcp, 5001/tcp, 8080-8081/tcp, 4001/udp   nftstorage_devcontainer_ipfs0_1
8e8b1d68fd2b   nftstorage_devcontainer_db                             "docker-entrypoint.s…"   4 seconds ago   Up 3 seconds   0.0.0.0:5432->5432/tcp, :::5432->5432/tcp     nftstorage_devcontainer_db_1
99038f99c456   vsc-nft.storage-b67da050d84e066c07118061da437233-uid   "docker-entrypoint.s…"   4 seconds ago   Up 3 seconds   0.0.0.0:8787->8787/tcp                        nftstorage_devcontainer_dev_1


curl localhost:8787/version
```

Let me know if there's anything I can do to help. I'd really like it if I could get my local environment set up

yusefnapora:
gonna close this since Hugo's simple-dev branch makes it happily obsolete 😄

Issue #742, opened 2021-11-08
Update to @web-std/* 3.0


Gozala:
I've updated @web-std packages to take advantage of node16 [web streams](https://nodejs.org/api/webstreams.html#webstreams_web_streams_api) and [Blob](https://nodejs.org/api/buffer.html#class-blob) APIs where they are available.

I think we should update our uses of those libs to take advantage of those APIs.

PR #833, opened 2021-11-22
chore!: update dependencies

alanshaw:
This moves to the new v3 @web-std modules so is a breaking change.

resolves https://github.com/nftstorage/nft.storage/issues/742


vasco-santos (approved):
LGTM

PR #838, opened 2021-11-22
feat: blog

### LINKED ISSUE #722: Blog for NFT.Storage
We should have a blog website for NFT.Storage! Would be a great place to announce new features, etc.

Related to blog request for web3 https://github.com/web3-storage/web3.storage/issues/591

## Acceptance criteria:
- [x] Email subscriber list implemented
- [x] insert first blog post content https://docs.google.com/document/d/1E_f9iMmaYuPgw1cj17-3l0GHVHvqwkEKLhp5Bsx3fig/edit

- [x] Blog design implemented at blog.nft.storage
- [x] Repo for writing blog content
- [x] Blob placed on /blog route, with blog.nft.storage redirecting to /blog
- [x] ensure blog top nav matches nft.storage top nav

Codigo-Fuentes:
**in progress**

- adds blog to website
- closes #722


the-simian:
ok I've made several changes in the following PRs, since It might be better so show than tell, here's some before/afters to help see what is different from before

Page layout (before)
image

Pay Layout (after)
image

Tags (before left, after right)
1-tags

Card Interactivity (before left after right
2-card
)

the-simian:
Pager (Before/ After)
pager

the-simian:
Accessibility (before = left, After = right)
image

I'm still trying to find that empty button that is throwing the one error, but I think it might not be in the blog, maybe the header or an extension.
Still huge improvement on accessibility.

the-simian:
More small things
- button animations (the scaling looked fuzzy and clunky so I've replace that)
- header uses the nft.storage font to match the swagger docs and main site
- header-blog layout uses grid. It can stretch but doesn't stretch too wide

the-simian:
Main article (accessibility)
(before left, after right). I had to switch the markdown to a more accessible theme as well.
image

Before/ After (feel free to swap back out for a light theme if you find one that doesn't make Waves angry, but I'm cool with this one)
image

the-simian:
More small things. The scaling effect looks fuzzy on my screen, so I made a cleaner animation for these (also added aria tags):
small things

Mobile layout is centered: (also button animation changed from fuzzy-jumpy-scale to something cleaner)
small-screeens
this particular post doesn't have tags, but they appear in about the same y-position

That's everything. Everything else looked good to me!

Codigo-Fuentes:
@the-simian the layout changes and animation changes look good to me. Using grid for the blog cards is a big improvement and the code blocks in the blogs look much better. Re animation, I wonder what you think about trying it in the other direction? I.e., hovering makes it expand rather than extract.

Re other changes to styles:

- The change of heading fonts makes sense but i think we should check with the designer; the previous font matches spec
- changing to centering content at breakpoint to me debatably makes sense but I strongly prefer left-aligned which matches the design and the other cards as well. If anything, I think that there should be reduction in font sizes at some breakpoint for the highlight card.
- The spacing inside the "highlight card" now looks incorrect to me. I strongly prefer all previous styles internal to the highlight card, which precisely matched the design. Can we change these back, please? See figma and photos below for comparison:

https://www.figma.com/file/mxRXqlzGd4MovE85euZtu8/NFT-STORAGE?node-id=1307%3A354

prev
curr
curr-sm
prev-sm

the-simian:
> The change of heading fonts makes sense but i think we should check with the designer; the previous font matches spec

I hadn't cross-referenced the Figma but it does stick out that the most noticeable font that is in heavy use is unused in the blog. I assumed that was an oversight, because this will be a subroute in the rest of the site. It would be very weird if you went to a subsection of any site and all the fonts and colors are different, and in a way that seems unintentional. I don't care which font is chosen (chivo or chicago FLF), my only preference if my voice matters is that H1's/titles/fonts/buttons/components are thematically consistent. The blog design does not match the docs or the main site as it is; maybe it was designed without knowledge of these other pages? I really can't say but I agree we should surface this concern to the designer and choose something sitewide that makes sense.

> changing to centering content at breakpoint to me debatably makes sense but I strongly prefer left-aligned which matches the design and the other cards as well. If anything, I think that there should be reduction in font sizes at some breakpoint for the highlight card.

I am ambivalent on this, if it looks correct on small devices then no problem. Its more common to see center alignment on mobile since it backhandedly resolves a lot of balance and positioning issues very easily on small screens. That said, let's just change it back and verify it looks ok on mobile if you prefer the way it was beforehand.

> The spacing inside the "highlight card" now looks incorrect to me. I strongly prefer all previous styles internal to the highlight card, which precisely matched the design. Can we change these back, please? See figma and photos below for comparison:


I agree that the y-positioning (and some x-positioning) looks off now and we should change this back. Really appreciate a second set of eyes here. The thing I notice first is the space between the text and the button. When working, I was focused on the fact that on very large screens, when you extended the width the material was docked to the left, which looked unusual after making the change to center the body of the document. (That and the way the button was constructed made it difficult to modify.) As long as those two things 'stay' I am happy. Anyways, you're right and I'd say of every item on your feedback this is the most important one and should be addressed before merging.

Codigo-Fuentes:
Also, I just want to add that, at least in my mind, every blog should probably have at least one tag. It will work either way, but I think everything will look best when all metadata is consistently provided.

Codigo-Fuentes:
Just want to add that I think that every blog should probably have at least one tag. It will work either way, but I think everything will look best when all metadata is consistently provided.

the-simian:
I reviewed this, and I made some CSS changes
1. I fixed something I somewhat broke but was unnoticed - I added back in some vars that weren't there and were affecting the 'hologram' on the card. They were missing, now they're back.
2. Added in some CSS that makes the border-bottom show on the highlight-card. Super small regression I noticed, likely also after merging in the refactored css
3. I made a change on the highlight card that only affects the highlight card at very wide sizes. This doesn't touch any of the positioning inside of the highlight card, just big-wide browser sizes:
Before:
ultrawide-example

After: 
image

the-simian:
Ok @Codigo-Fuentes, I've changed the posts to be only the one "hello-world" post, I also added decimals and discs to the line items inside the markdown itself, I could see you had pulled them out of the main blog to make the tags and stuff work so this is just the prose.

The only thing I want to verify is the behavior for selecting the highlighted post. Right now, it will just take whichever post it sees first, and I can see you sort them by birthtime (when the file was created on the actual filesystem). Could there ever be a situation where the blog dates inside the front matter of the post should be preferred to the actual files creation time?

In other words, this blog post was added to the filesystem on x date, but y blog post was added later but dated earlier. My instinct is we should use the frontmatter's date and not the filesystem date.

The other thing is do the authors ever want to explicitly state that a certain post (or posts) is "the highlighted post". This, and also the preference for frontmatter date,  could certainly be added in a separate pr. If we don't want to block release, we might consider adding that to the backlog just to flag that we know we'll want that in the future.

Also to verify the tag/filter behavior, this willfully removes the first post (highlighted post) except in situations where you are filtering, then you'd see it in both places as it is now.

the-simian (approved):
OK I reviewed this, and I made some CSS changes, so I'm going to have you review my work in return

the-simian (changes requested):
In order for this to work in prod, we'll need to make sure to add the Mailchimp API key to the CF env (this is why the subscribe list works in dev but not prod currently)

However, I had an issue where the success and error states didn't seem to update the UI when the call returns; so this still needs some code changes first

Codigo-Fuentes:
> The only thing I want to verify is the behavior for selecting the highlighted post. Right now, it will just take whichever post it sees first, and I can see you sort them by birthtime (when the file was created on the actual filesystem). Could there ever be a situation where the blog dates inside the front matter of the post should be preferred to the actual files creation time?
> 

> In other words, this blog post was added to the filesystem on x date, but y blog post was added later but dated earlier. My instinct is we should use the frontmatter's date and not the filesystem date.
> 
> The other thing is do the authors ever want to explicitly state that a certain post (or posts) is "the highlighted post". This, and also the preference for frontmatter date, could certainly be added in a separate pr. If we don't want to block release, we might consider adding that to the backlog just to flag that we know we'll want that in the future.
> 

@the-simian Yes, I sorted by birthtime just as a starting point; i didn't know how or by whom the blogs were going to be created so I didn't want to assume a certain format (e.g., a timestamp) as the date entry and maybe multiple blogs could drop on the same day. I agree (and i left a note in the code) that we could just have a flag that tells it which blog to highlight but then that means each time a new blog is added that flag will have to be deleted from the previously highlighted blog (or some other system could work like flag + timestamp so we know which is latest flagged.

> Also to verify the tag/filter behavior, this willfully removes the first post (highlighted post) except in situations where you are filtering, then you'd see it in both places as it is now.

That is the expected and intended behavior yes.

Codigo-Fuentes:
> In order for this to work in prod, we'll need to make sure to add the Mailchimp API key to the CF env (this is why the subscribe list works in dev but not prod currently)
> 
> However, I had an issue where the success and error states didn't seem to update the UI when the call returns; so this still needs some code changes first

Cool. How can i / we get access to that? Or who can we pass the key to to handle that?

the-simian:
> Cool. How can i / we get access to that? Or who can we pass the key to to handle that?
@JeffLowe is working on this so I'm sure it will be resolved soon. 

> Strange I was not experiencing this. Can you provide steps to reproduce?
Just used the form normally. I checked to see if this was my DNS sinkhole being aggressive and that wasn't the case. However, I tried it again today and it *did* work so maybe it was some sort of intermittent deal. Now that I see it working it looks good, the  only thing I would add is a CTA to go back after success, and since I'm in it right now I'll add that

the-simian:
handled new form accessibility
image

the-simian (approved):

the-simian:
here's the first post as well, just for a visual:
image

packages/api/src/index.js
```diff
@@ -104,6 +105,9 @@ r.add('get', '/internal/tokens', withMode(tokensList, RO), [postCors])
 r.add('post', '/internal/tokens', withMode(tokensCreate, RW), [postCors])
 r.add('delete', '/internal/tokens', withMode(tokensDelete, RW), [postCors])
 
+// Blog
+r.add('post', '/blogSubscribe', withMode(blogSubscribe, RW), [postCors])
```
alanshaw, line 109: Please can we put this under the /internal namespace so people know it's not a public API?

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,86 @@
+import { JSONResponse } from '../utils/json-response.js'
+import btoa from 'btoa'
+
+const SERVER_PREFIX = 'us5'
+const LIST_ID = '64f6e3fd11'
+const API_KEY = process.env.MAILCHIMP_API_KEY
```
alanshaw, line 6: CF will make this a global variable. Please deal with that and import from `constants.js`

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,86 @@
+import { JSONResponse } from '../utils/json-response.js'
+import btoa from 'btoa'
```
alanshaw, line 2: Isn't btoa native - why do we need to import?

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,86 @@
...
+const headers = {
+  'Content-Type': 'application/json; charset=UTF-8',
+  Authorization: `Basic ${TOKEN}`,
+}
+
+/**
+ *  @param {string} email
+ */
+export const getMailChimpUser = async (email) => {
+  const url = `https://${SERVER_PREFIX}.api.mailchimp.com/3.0/lists/${LIST_ID}/members/${email}`
```
alanshaw, line 18:
```suggestion
  const url = `https://${SERVER_PREFIX}.api.mailchimp.com/3.0/lists/${LIST_ID}/members/${encodeURIComponent(email)}`
```

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,86 @@
...
+    headers,
+    body,
+  })
+}
+
+/**
+ *  @param {string} email
+ */
+const updateSubscriber = async (email) => {
+  const url = `https://${SERVER_PREFIX}.api.mailchimp.com/3.0/lists/${LIST_ID}/members/${email}/tags`
```
alanshaw, line 47:
```suggestion
  const url = `https://${SERVER_PREFIX}.api.mailchimp.com/3.0/lists/${LIST_ID}/members/${encodeURIComponent(email)}/tags`
```

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * @param {any} mailChimpUser
+ * @returns {boolean}
+ */
+const shouldUpdateMailChimpUser = (mailChimpUser) =>
+  mailChimpUser?.statusText === 'OK'
+
+/** @type {import('../bindings').Handler} */
+export const blogSubscribe = async (event) => {
+  console.log('HERE IS PROCESS.ENV, ', process.env)
```
alanshaw, line 68:
```suggestion
```

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,86 @@
...
+      ? await updateSubscriber(body.email)
+      : await addSubscriber(body.email)
+
+    return new JSONResponse({
+      ok: true,
+      value: response,
+    })
+  } catch (/** @type {any} */ error) {
+    console.error('ping: ', error)
+    throw new Error(error)
```
alanshaw, line 82:
```suggestion
    throw error
```

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,86 @@
...
+
+const headers = {
+  'Content-Type': 'application/json; charset=UTF-8',
+  Authorization: `Basic ${TOKEN}`,
+}
+
+/**
+ *  @param {string} email
+ */
+export const getMailChimpUser = async (email) => {
```
alanshaw, line 17: This should return a user object or undefined.

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,86 @@
...
+const shouldUpdateMailChimpUser = (mailChimpUser) =>
+  mailChimpUser?.statusText === 'OK'
+
+/** @type {import('../bindings').Handler} */
+export const blogSubscribe = async (event) => {
+  console.log('HERE IS PROCESS.ENV, ', process.env)
+  const body = await event.request.json()
+  try {
+    const mailChimpUser = await getMailChimpUser(body.email)
+    const response = shouldUpdateMailChimpUser(mailChimpUser)
```
alanshaw, line 72: The call to this function should not be necessary as you should have a user or not at this point. Just check for existence. Make the getter return a user or undefined.

packages/api/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "allowJs": true,
     "checkJs": true,
     "target": "esnext",
-    "lib": ["ESNext"],
+    "lib": ["ESNext", "DOM.Iterable"],
```
alanshaw, line 8: Why please?

packages/client/README.md
```diff
@@ -31,7 +31,7 @@ First, obtain an API token from https://nft.storage and use it in place of `API_
 import { NFTStorage, File } from 'nft.storage'
 const client = new NFTStorage({ token: 'API_TOKEN' })
 
-async function main () {
+async function main() {
```
alanshaw, line 34: Please revert

packages/website/tsconfig.json
```diff
@@ -21,7 +21,7 @@
     },
     "incremental": true
   },
-  "include": ["lib", "pages", "components"],
+  "include": ["lib", "pages", "components", "pages/blog/index/index.js"],
```
alanshaw, line 24: Does the "pages" declaration not include this?

alanshaw (changes requested):

packages/api/package.json
```diff
@@ -22,11 +22,15 @@
...
     "@magic-sdk/admin": "^1.3.0",
+    "@mailchimp/mailchimp_marketing": "^3.0.72",
     "@nftstorage/ipfs-cluster": "^3.4.3",
     "@supabase/postgrest-js": "^0.34.1",
+    "@types/btoa": "^1.2.3",
+    "btoa": "^1.2.1",
     "debug": "^4.3.2",
     "merge-options": "^3.0.4",
     "multiformats": "^9.4.10",
+    "node-fetch": "^3.1.0",
```
alanshaw, line 33: Why need this?

packages/website/pages/_document.js
```diff
@@ -24,6 +24,10 @@ class MyDocument extends Document {
             type="font/woff2"
             crossOrigin="true"
           ></link>
+          <link
+            href="https://fonts.googleapis.com/css2?family=Chivo:wght@400;700&display=swap"
+            rel="stylesheet"
+          />
```
hugomrdias, line 30: why do we need a new font ? and if we really do it should be processed and served from cf pages

packages/website/pages/api/addTags.js
```diff
@@ -0,0 +1,37 @@
+import mailchimp from '@mailchimp/mailchimp_marketing'
+
+const LIST_ID = '64f6e3fd11'
+
+/**
+ * Subscribe user to blog mailing list
+ * @param {import('next').NextApiRequest} req
+ * @param {import('next').NextApiResponse} res
+ */
+export default async (req, res) => {
```
hugomrdias, line 10: can you please fix the linter warnings please ?

packages/website/pages/api/user.js
```diff
@@ -0,0 +1,22 @@
...
+
+const LIST_ID = '64f6e3fd11'
+
+/**
+ * Get user subscription info
+ * @param {import('next').NextApiRequest} req
+ * @param {import('next').NextApiResponse} res
+ */
+export default async (req, res) => {
+  mailchimp.setConfig({
```
hugomrdias, line 11: where is this running? cf pages does not support this

hugomrdias (changes requested):
why are we using `react-markdown` instead of `@next/mdx` or `next-mdx-remote`, these do the work either in the a loader or getStaticProps and have a very small footprint.

packages/api/src/routes/blog-subscribe.js
Codigo-Fuentes: because there were errors saying it was deprecated. Settled on: Buffer.from(`:${API_KEY}`).toString('base64') with no type errors

packages/api/tsconfig.json
Codigo-Fuentes: failed one or more checks with type errors complaining about this I believe:

https://github.com/nftstorage/nft.storage/blob/cb5171ffe480a03f55ebbcde05a5eb762f5721e9/packages/api/src/routes/nfts-store.js#L31

packages/client/README.md
Codigo-Fuentes: sorry i'm not sure why that happened

packages/website/tsconfig.json
Codigo-Fuentes: not sure how this got in there

packages/api/package.json
Codigo-Fuentes: we don't. removing.

packages/website/pages/api/addTags.js
Codigo-Fuentes: next api routes have been removed

packages/website/pages/api/user.js
Codigo-Fuentes: next api routes have been removed

packages/website/pages/_document.js
Codigo-Fuentes: New font was specified by designer for blog. Further discussion led to replacing bold weight with title fonts consistent with rest of site. Will change to handle in same way as other font.

packages/api/src/routes/blog-subscribe.js
Codigo-Fuentes: refactored to just check if the user exists

packages/api/src/routes/blog-subscribe.js
Codigo-Fuentes: refactored to just check if the user exists

Codigo-Fuentes:
swapped in next-mdx-remote for react-markdown

packages/api/src/routes/blog-subscribe.js
alanshaw: I can't see any indication that [`btoa`](https://developer.mozilla.org/en-US/docs/Web/API/btoa) is deprecated on MDN.

Personally I'd reach for this over pulling in a Node.js specific library to the bundle and suppress the warning instead.

packages/api/src/index.js
```diff
@@ -104,6 +105,11 @@ r.add('get', '/internal/tokens', withMode(tokensList, RO), [postCors])
 r.add('post', '/internal/tokens', withMode(tokensCreate, RW), [postCors])
 r.add('delete', '/internal/tokens', withMode(tokensDelete, RW), [postCors])
 
+// Blog
+r.add('post', '/internal/blogSubscribe', withMode(blogSubscribe, RW), [
```
alanshaw, line 109: Not strong feels and not blocking, but what about `/internal/blog/subscribe` - just incase we ever want to add more blob specific endpoints?

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,81 @@
+import { JSONResponse } from '../utils/json-response.js'
+
+const SERVER_PREFIX = 'us5'
+const LIST_ID = '64f6e3fd11'
+const API_KEY = MAILCHIMP_API_KEY
+const TOKEN = Buffer.from(`:${API_KEY}`).toString('base64')
```
alanshaw, line 6: As mentioned in another comment, we should avoid Node.js specific APIs as we're not running in a Node.js environment.

> I can't see any indication that [`btoa`](https://developer.mozilla.org/en-US/docs/Web/API/btoa) is deprecated on MDN.
>
> Personally I'd reach for this over pulling in a Node.js specific library to the bundle and suppress the warning instead.

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,81 @@
+import { JSONResponse } from '../utils/json-response.js'
+
+const SERVER_PREFIX = 'us5'
+const LIST_ID = '64f6e3fd11'
+const API_KEY = MAILCHIMP_API_KEY
```
alanshaw, line 5:
```suggestion
import { JSONResponse } from '../utils/json-response.js'
import { secrets } from '../constants.js'

const SERVER_PREFIX = 'us5'
const LIST_ID = '64f6e3fd11'
const API_KEY = secrets.mailchimp
```

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,81 @@
...
+ */
+export const isChimpUser = async (email) => {
+  const url = `https://${SERVER_PREFIX}.api.mailchimp.com/3.0/lists/${LIST_ID}/members/${encodeURIComponent(
+    email
+  )}`
+  const res = await fetch(url, {
+    method: 'GET',
+    headers,
+  })
+  return res.statusText === 'OK'
```
alanshaw, line 24:
```suggestion
  return res.ok
```

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,81 @@
...
+      : await addSubscriber(body.email)
+
+    return new JSONResponse({
+      ok: true,
+      value: response,
+    })
+  } catch (/** @type {any} */ error) {
+    console.error(error)
+    throw Error(error)
+  }
```
alanshaw, line 78: You don't need to catch this to log the error. We have a global error handler that will deal with that.

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,81 @@
...
+export const blogSubscribe = async (event) => {
+  const body = await event.request.json()
+  try {
+    const response = (await isChimpUser(body.email))
+      ? await updateSubscriber(body.email)
+      : await addSubscriber(body.email)
+
+    return new JSONResponse({
+      ok: true,
+      value: response,
```
alanshaw, line 73: `response` is a response object and not really JSON serializable. You either want to do this:

```suggestion
      value: await response.json(),
```

...or what I'd recommend is just to return no value here - `ok: true` is enough.

packages/website/components/blog/cards.js
```diff
@@ -0,0 +1,110 @@
...
+ * @param {Object} props
+ * @param {import('../types').PostMeta} props.post
+ * @param {() => void} props.onClick
+ * @returns {JSX.Element}
+ */
+
+export const Card = ({ post, onClick }) => (
+  <Link href={`/blog/post/${post.slug}`}>
+    {/* 👇 because next/link won't accept onClick */}
+    <a
```
alanshaw, line 21: Maybe a generic container element instead so we don't end up with nested anchor tags?

packages/website/lib/subscribe.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param {string} email
+ * @param {string} [version]
+ */
+
+export const subscribe = async (email, version = '') => {
+  const loginURL = version
+    ? `/v${version}/internal/blogSubscribe`
+    : '/internal/blogSubscribe'
+  const res = await fetch(API + loginURL, {
```
alanshaw, line 15: Version stuff is legacy...no need to include here (and can be removed from existing routes).

```suggestion
export const subscribe = async (email) => {
  const subscribeURL = '/internal/blogSubscribe'
  const res = await fetch(API + subscribeURL, {
```

packages/website/lib/subscribe.js
```diff
@@ -0,0 +1,29 @@
...
+    }),
+  })
+  const body = await res.json()
+  if (body.ok) {
+    return body.value
+  } else {
+    if (JSON.parse(body.response.text).title === 'Member Exists')
+      throw new Error('exists')
+    throw new Error(body.error)
+  }
```
alanshaw, line 28: I think the member exists thing does not happen - they get updated, and this always succeeds unless network error or something.

```suggestion
  if (!body.ok) {
    throw new Error(body.error)
  }
```

packages/website/package.json
```diff
@@ -12,21 +12,28 @@
   },
   "dependencies": {
     "@magic-ext/oauth": "^0.11.2",
+    "@mailchimp/mailchimp_marketing": "^3.0.72",
+    "@types/mailchimp__mailchimp_marketing": "^3.0.2",
+    "@types/react-syntax-highlighter": "^13.5.2",
```
alanshaw, line 17: Move `@types` to `devDependencies`?

packages/website/pages/blog/index.js
```diff
@@ -0,0 +1,319 @@
...
+import clsx from 'clsx'
+import fs from 'fs'
+import matter from 'gray-matter'
+import { usePagination } from '../../lib/usePagination'
+import { useRouter } from 'next/router'
+
+const BLOG_ITEMS_PER_PAGE = 9
+
+export async function getStaticProps() {
+  const files = fs.readdirSync('all-blogs')
```
alanshaw, line 17: Is there a plugin that causes this to get run at compile time?

packages/website/pages/blog/subscribe.js
```diff
@@ -0,0 +1,125 @@
...
+
+    setStatus('pending')
+    if (errorMsg) {
+      setErrorMsg('You could not subscribe to the list')
+    }
+    setDisabled(true)
+    // try to subscribe user
+    try {
+      await subscribe(userMail)
+      console.log('SUCCESS! ')
```
alanshaw, line 54:
```suggestion
```

packages/website/pages/blog/subscribe.js
```diff
@@ -0,0 +1,125 @@
...
+      setStatus('success')
+    } catch (/** @type {any} */ error) {
+      console.error('ERROR SUBSCRIBING USER: ', error.message)
```
alanshaw, line 57: So we get the stack also:

```suggestion
      console.error('ERROR SUBSCRIBING USER: ', error)
```

packages/website/pages/blog/subscribe.js
```diff
@@ -0,0 +1,125 @@
...
+    if (status === '') setDisabled(false)
+    setErrorMsg('')
+  }, [status])
+
+  /**
+   * @param {import('react').ChangeEvent<HTMLFormElement>} e
+   */
+  const onSubmit = async (e) => {
+    e.preventDefault()
+    const userMail = user?.email || email
```
alanshaw, line 42: Should we prefer the email address input by the user? They may not want to use their signup email for email marketing.

packages/website/pages/blog/index.js
```diff
@@ -0,0 +1,319 @@
...
+
+  let featuredImage = null
+
+  // we can create a "featured" flag if we want and feature blogs that way
+
+  /**
+   * @param {string} fn
+   */
+  const getBirthtime = (fn) =>
+    fs.statSync(`all-blogs/${fn}`).birthtime.getTime()
```
alanshaw, line 27: Does this work if I freshly clone the repo? Don't all the files get the same birthtime at that point? Personally, I'd use a convention of `YYYY-MM-DD-post-name.mdx` and sort alphabetically.

You also don't need a `date` in the front matter.

packages/website/pages/blog/index.js
```diff
@@ -0,0 +1,319 @@
...
+
+  files.sort((a, b) => getBirthtime(b) - getBirthtime(a))
+
+  const posts = files
+    ? files
+        .filter((filename) => filename.toLowerCase() !== '.ds_store')
```
alanshaw, line 33: Hopefully this is always compiled on a mac 😱 (not the case - especially in CI). It might be better to filter files that begin with `.` (hidden).

```suggestion
        .filter((filename) => !filename.startsWith('.'))
```

packages/website/components/tags.js
```diff
@@ -0,0 +1,56 @@
...
+  const isString = typeof tag === 'string'
+  const inner = (
+    <span className={clsx('ph2 pv1 f6 ba ttc mr2', isString && 'select-none')}>
+      {isString ? tag : tag.label}
+    </span>
+  )
+  return isString ? (
+    inner
+  ) : (
+    <div className="">
```
alanshaw, line 21:
```suggestion
    <div>
```

packages/api/src/constants.js
```diff
@@ -1,10 +1,11 @@
-// let MAGIC_SECRET_KEY, SALT, PINATA_JWT, SENTRY_DSN, DATABASE_TOKEN, CLUSTER_SERVICE
+// let MAGIC_SECRET_KEY, SALT, PINATA_JWT, SENTRY_DSN, DATABASE_TOKEN, CLUSTER_SERVICE, MAILCHIMP_API_KEY
```
alanshaw, line 1: Please update the README with instructions to set this secret along with the others.

alanshaw (changes requested):
Please could you also address the warnings from github actions?

packages/api/src/routes/blog-subscribe.js
Codigo-Fuentes: i went down a rabbithole of getting warnings/errors telling me multiple things were deprecated or to use other things until settling on this. I'll just use btoa and ignore the warnings

packages/website/pages/blog/index.js
Codigo-Fuentes: I believe it works; didn't know at the time if I should assume naming conventions. I'll change to the way you suggested. 

I also thought that we might want to add a flag because we might want to highlight a blog that isn't the newest. Can always change in the future if desired. 

@the-simian also suggested we could do a carousel if we wanted to highlight multiple blogs.

packages/api/src/constants.js
the-simian: I ticketed this: https://github.com/nftstorage/nft.storage/issues/904

packages/website/lib/subscribe.js
Codigo-Fuentes: sorry this was prior to api refactor

packages/website/lib/subscribe.js
the-simian: I ticketed the removal of the version info from the other routes https://github.com/nftstorage/nft.storage/issues/906

packages/api/src/constants.js
Codigo-Fuentes: readme updated for production.

dchoi27 (approved):
all good from content perspective

packages/website/pages/blog/post/[slug].js
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * @param {Object} props
+ * @param {import('../../../components/types').Post} props.post
+ * @returns {JSX.Element}
+ */
+const Post = ({ post }) => {
+  const SHARE_TEXT = 'Store NFTs for free!'
+  // localhost will not work as currentUrl with fb or linkedin
+  const [currentUrl, setCurrentUrl] = useState('')
+  useEffect(() => setCurrentUrl(window.location.href))
```
Comment (hugomrdias) [inline packages/website/pages/blog/post/[slug].js:51]:
```suggestion
  useEffect(() => setCurrentUrl(window.location.href), [])
```

packages/website/components/blog/cards.js
hugomrdias: next/link does not duplicate child `a`

hugomrdias (changes requested):
This looks really good ! nice job

I left some comments inline and also the following :

- Are we using both versions of the dark log ? also why use the svg version if the png version is much smaller ?

- I think it would be better to use cloudflare pages functions for the mailchimp stuff instead of the API, this way its deployed with the pages and not the API.

Note: don't use `/api` for the functions because that is already being routed to the API worker, something like `/web-api` or similar works.
  
https://developers.cloudflare.com/pages/platform/functions#setup

- last nitpick (sorry :) ) can we renamed the `all-blogs` folder to just `posts` please ?

packages/website/package.json
```diff
@@ -16,17 +16,21 @@
...
     "nft.storage": "^3.3.0",
     "postcss-flexbugs-fixes": "^5.0.2",
     "postcss-import": "^14.0.2",
     "postcss-preset-env": "^7.0.1",
     "querystring-browser": "^1.0.4",
     "react": "17.0.2",
     "react-dom": "17.0.2",
+    "react-icons": "^4.3.1",
     "react-if": "^4.0.1",
+    "react-paginate": "^8.0.0",
```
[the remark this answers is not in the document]
Codigo-Fuentes: nope. removed.

JeffLowe:
Switch to use cloudflare pages functions for the blog's mailchimp subscription (as @hugomrdias suggested) ticketed here https://developers.cloudflare.com/pages/platform/functions#setup as a post - deploy improvement.  We'll get current functioning blog out and usable by @dchoi27 and team, then improve with this later.

packages/api/src/index.js
```diff
@@ -104,6 +105,11 @@ r.add('get', '/internal/tokens', withMode(tokensList, RO), [postCors])
 r.add('post', '/internal/tokens', withMode(tokensCreate, RW), [postCors])
 r.add('delete', '/internal/tokens', withMode(tokensDelete, RW), [postCors])
 
+// Blog
+r.add('post', '/internal/blog/subscribe', withMode(blogSubscribe, RW), [
+  postCors,
+])
```
alanshaw, line 111: Doesn't need maintenance mode restrictions.

```suggestion
r.add('post', '/internal/blog/subscribe', blogSubscribe, [postCors])
```

packages/api/src/routes/blog-subscribe.js
```diff
@@ -0,0 +1,79 @@
...
+  try {
+    const response = (await isChimpUser(body.email))
+      ? await updateSubscriber(body.email)
+      : await addSubscriber(body.email)
+    return new JSONResponse({
+      ok: true,
+    })
+  } catch (/** @type {any} */ error) {
+    throw Error(error)
+  }
```
alanshaw, line 76: This catch block is redundant

alanshaw (approved):
Minor suggestions but this LGTM 🎉

hugomrdias:
Why use the svg logo and not the png (its smaller) ?

Theres still linter warnings to be fixed.

After discussing with @dchoi27 we think its best to use the same body text font everywhere in the website, so please remove the new font introduced for the blog and fallback to the system font used in the rest of the website.

JeffLowe:
Use of svgs for logos is preferred due to better scaling.  They look crisp despite resolutions.

All checks are green across the board now. 

Regarding font change...sure.  We've now switched away from the design-specified font to the same font used on nft.storage.

packages/website/styles/markdown.css
```diff
@@ -0,0 +1,998 @@
...
+  display: inline-block;
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' version='1.1' width='16' height='16' aria-hidden='true'%3E%3Cpath fill-rule='evenodd' d='M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z'%3E%3C/path%3E%3C/svg%3E");
+}
+.post {
+  -ms-text-size-adjust: 100%;
+  -webkit-text-size-adjust: 100%;
+  line-height: 1.5;
+  color: #24292e;
+  font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Helvetica, Arial,
+    sans-serif, Apple Color Emoji, Segoe UI Emoji;
```
hugomrdias, line 65:
```suggestion
  font-family: var(--sansserif);
```

packages/website/styles/markdown.css
```diff
@@ -766,7 +766,7 @@ pre > div code span {
 }
 
 .post table th {
-  font-family: 'Chivo', sans-serif;
+  font-family: sans-serif;
```
hugomrdias, line 769:
```suggestion
  font-family: var(--sansserif);
```

packages/website/styles/markdown.css
```diff
@@ -0,0 +1,998 @@
...
+
+.post h1,
+.post h2,
+.post h3,
+.post h4,
+.post h5,
+.post h6 {
+  margin-top: 0;
+  margin-bottom: 0;
+  font-family: 'ChicagoFLF', sans-serif;
```
hugomrdias, line 206:
```suggestion
  font-family: var(--nstitle);
```

packages/website/styles/markdown.css
```diff
@@ -0,0 +1,998 @@
...
+
+.post dd {
+  margin-left: 0;
+}
+
+.post code,
+.post pre {
+  font-family: SFMono-Regular, Consolas, Liberation Mono, Menlo, monospace;
+  font-size: 12px;
+}
```
hugomrdias, line 284: `code` and `pre` are already defined a few lines up can we merge these ?

Also can you make a var for `SFMono-Regular, Consolas, Liberation Mono, Menlo, monospace;` and use it everywhere we want monospace ?

hugomrdias (changes requested):

hugomrdias (approved):

PR #840, opened 2021-11-23
feat: add marketplace logos

### LINKED ISSUE #736: Add logos of big name marketplaces using NFT.Storage to homepage
Will reach out to marketplaces this week asking if it'd be OK - will follow up with who we want to include

When we put this on our homepage, we should do it in a way that auto-scales as more users' logos are added, so future users can PR to the folder including these logos and make it easy to add more logos.

## Designs
unknown

## Logo Assets for engineers
https://drive.google.com/drive/folders/1RpvlUcMTFIPVwhyPPS1nnK3GiXQYQMyw?usp=sharing

## AC / Tasks
- [x] developers can PR to add new logos
- [x] Homepage has section showing all marketplace logos
- [ ]  Before deploying out to production, share review link with David (so he can share with OpenSea who has requested review before publishing)
- [ ] add a "Trusted by" header
- [ ] move all the logos  into one row if possible (so it doesn't take up too much space)

Codigo-Fuentes:
closes #736 

@dchoi27 please review images below. Current implementation tries to do something that looks decent and catches any logos we get. We should, however, definitely implement a spec for any logo submissions detailing filetype, size, etc. We would then have better control of how the logos get displayed. E.g., Hola Plex and Magic Eden having white and black backgrounds, respectively, breaks the aesthetic.

full
med
iphonex


dchoi27:
Thanks! Can we add a "Trusted by" header? And try to move all the logos right now into one row (so it doesn't take up too much space)

Also, think having the section up top right now makes sense, but if the list gets much longer we should move to below the About section.

Will try to hunt down transparent logos from the two offenders

dchoi27 (changes requested):
Thanks! Can we add a "Trusted by" header? And try to move all the logos right now into one row (so it doesn't take up too much space)

Also, think having the section up top right now makes sense, but if the list gets much longer we should move to below the About section.

Will try to hunt down transparent logos from the two offenders

Codigo-Fuentes:
also, maybe it's fine: https://swiperjs.com/

looking at examples and at this page has like five sets of examples of similar things. At the very bottom it looks like how we have it.

Codigo-Fuentes:
Made updates ... the row still needs to wrap eventually...  also changed bg color rather than having a divider I think it looks better.

full
iphonex

Codigo-Fuentes:
Trusted by

Screen Shot 2021-11-23 at 6 14 18 PM

dchoi27:
Amazing looks great. Nit - can we decapitalize "by"

Codigo-Fuentes:
✅

dchoi27 (approved):

dchoi27:
Holding off on merging until we have sign-off from OpenSea on how it looks!

dchoi27:
Have the OK to push! Let's merge!

Issue #839, opened 2021-11-23
Button component in /website triggers accessibility issues


the-simian:
Discovered while working on issue #838, the remaining accessibility issues are in the header. The root cause is how the button is constructed. My proposal is we take the construction pattern used in the blog (in progress) and the swagger docs, since it does not have this issue and make this the core pattern for button construction.

dashcraft:
I ran axe dev tools, the issue is the relatively positioned button for the header login hologram effect.
https://axe.deque.com/issues/d71db089-0945-468a-b506-ae8053594ed8

This is the new CSS that corrects the issue:
https://github.com/nftstorage/nft.storage/pull/838/files#diff-becf08281a6c408e392af7727401de82afa7f73656dc6eb76ce8493f7b633ebcR130

PR #841, opened 2021-11-23
fix (website): CSS refactor for Accessibility Fixes

the-simian:
This is co-authored with @dashcraft as an onboarding issue/PR

References: https://github.com/nftstorage/nft.storage/issues/839

Summary of changes in this PR :

Important Changes:
1. The `global.css` file has been refactored into separate files and those files are being imported using the existing post-css pipeline in next.js. I've been working on a few tickets now and I made the judgment call to break it apart on this otherwise small issue because it was hitting critical mass. This accounts for the majority of the files changed

    This means:
    - There are several new css files, that contain the previous contents of global.css
        - Two of them, button and decorators contain some changes needed during the refactoring
    - the necessary post-css plugins are added in the package.json
    - postcss.config.json has been added; it is doing exactly the same thing it was doing before, but w/added import plugin

2. The button.js file has been improved so it does not construct the button as before, which was triggering the accessibility errors. 
    - The new approach is much better and is a CSS solution (vs dom hacks/wrapper classes).  
    - The button is also now situationally aware and assigns the proper role when it is a button or a link.
    - there is now a cautious variant for things like delete, with good contrast
  
3. Minor Accessibility fixes and refactoring on the other pages that had buttons, (even if it wasn't a button). (eg. Login.js, Upload.js)
   - Inputs have labels, labelfor, arial labeled by etc.
   - there's a screenreader class so the style should be visually identical to before.
   - Look at the bottom of the PR for accessibility before/after fixes

Unimportant Changes:
1. nfts-list and nfts-upload.js , lib.js:  these are non-functional changes, but the imports have been sorted; basically just ignore this. They're fine to merge.

2. countly.js, faqContent.js : throwing linter warnings because it had unnamed exports. Named exports were added, and the warnings are gone. Again, the change is nonfunctional but they're more in line with the existing project standards, also safe to merge. We noticed the linter warnings while testing the postcss refactor in prod mode


packages/website/components/button.js
```diff
@@ -62,7 +67,7 @@ export default function Button({
     [tracking, onClick, href]
   )
 
-  wrapperClassName = clsx(
```
dashcraft, line 65: this is because mutating props is a no-no

packages/website/components/button.js
```diff
@@ -73,16 +78,8 @@ export default function Button({
...
-    case 'dark':
-      variantClasses = 'bg-black white'
-      break
-
-    case 'light':
-      variantClasses = 'bg-white black'
-      break
-  }
+  // variant can never be null, it has an explicit default of light
+  const variantClasses = buttonVariants[variant]
```
dashcraft, line 82: technically, this const is unnecessary, but i do think it reads well, just my opinion

the-simian:
Note: @Codigo-Fuentes your recent logo work was pulled in and the classes were delegated to new css files (about.css and logos.css). Should be still functionally identical from main. Wanted to point that out for when you reviewed.

the-simian:
Here's some A/B comparisons of the audits for screens that had buttons:
Login Page 
image

the-simian:
Homepage
(before = left After = Right)
image
Note: the contrast error was fixed by slightly increasing the font size of the h2 text, which was too small against the orange.

packages/website/components/button.js
the-simian: Some thought and updates:

I got rid of this wrapper class prop, which was not needed for a button to be a button. Now just 'className' is all that's needed to override buttons styles. The button also supports an 'unstyled' prop that was prototyped on the blog PR, and was very useful. This iteration of the button is an improvement on that (@Codigo-Fuentes , if this goes into main before the blog, I will help integrate this change into that branch). I'd like both of you to look over the new button code on this PR if possible, I think you'll like the changes a lot.

As for wrapperClass, I've checked everywhere it was used, and was able to just use 'className'. Way simpler, and more canonical usage of class overrides IMO.

As for the variants, I am applying the variants and we can still type check the expected strings, but the actual implementation is in CSS now, sort of like how semantic UI and several other frameworks handle this. Overrides live in button.css, its open for extension this way. (We can add more colors, sizes and so on if needed).

The only place I saw using the small variant was the hamburger btw, and it wasn't even really 'small' just 'icon sized' so I did not go so far as to rename this but IMO it should not be called 'small' which in most frameworks implies, small...everything. Not in this PR but maybe sometime a button that can take an icon as an argument would be a better abstraction, but this PR kind of does plenty as is so I held back on that.

the-simian:
Files.js
(before/ After)
image

the-simian:
@dashcraft and @Codigo-Fuentes you're all clear to review, I've looked over all the affected pages and I can't find any regressions

packages/website/components/button.js
```diff
@@ -62,63 +63,44 @@ export default function Button({
...
+    )
+  } else {
+    btn = (
+      <div {...btnProps} {...props}>
+        {children}
+      </div>
+    )
+  }
+
+  console.log('children', children)
```
alanshaw, line 103:
```suggestion
```

packages/api/src/routes/nfts-list.js
```diff
@@ -1,8 +1,8 @@
-import { Validator } from '@cfworker/json-schema'
+import { HTTPError } from '../errors.js'
 import { JSONResponse } from '../utils/json-response.js'
-import { validate } from '../utils/auth.js'
+import { Validator } from '@cfworker/json-schema'
 import { toNFTResponse } from '../utils/db-transforms.js'
-import { HTTPError } from '../errors.js'
+import { validate } from '../utils/auth.js'
```
alanshaw, line 5: Please revert all these import order changes in the API. It is more time consuming for reviewers to review when there are no material changes in your diff because reviewers have to look at it to figure out there was nothing added or removed. It also triggers "release please" to create a release PR for the API and include this PR in the changelog.

If you want to run a code formatter against the code then the changes should be sent in a separate PR.

alanshaw (changes requested):
I'm stoked for all the accessibility fixes and CSS organisation 🎉

packages/website/pages/files.js
```diff
@@ -152,7 +148,9 @@ export default function Files({ user }) {
                           <th className="pa2 tl bg-nsgray br b--black w-33">
                             Size
                           </th>
-                          <th className="pa2 tc bg-nsgray" />
+                          <th className="pa2 tc bg-nsgray">
+                            <span className="sr-only">File Actions</span>
```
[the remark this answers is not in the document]
funkytek: This is screen-reader only

packages/website/pages/files.js
```diff
@@ -113,8 +109,8 @@ export default function Files({ user }) {
           </When>
           <When condition={status !== 'loading'}>
             <>
-              <div className="flex mb3 items-center">
-                <h1 className="chicagoflf mv4 flex-auto">Files</h1>
+              <div className="flex items-center mb3">
+                <h1 className="flex-auto chicagoflf mv4">Files</h1>
```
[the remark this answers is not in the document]
funkytek: Totally get the annoyance but the re-arrangement isn't arbitrary I believe this is a plugin that alphabetizes class names for best-practice readability

dashcraft:
@alanshaw I went ahead and updated the file changes per your request. If there were no material changes and it was only formatting I went ahead and reverted.

For the class name organization in files that were substantially changed, I went ahead and left those, I _can_ revert those if you believe it's too much, I personally like the idea of cleaning up class names and other general housekeeping items while working inside of the file, but I also see your point.

Give it a once over and if you still feel that it's too much noise, I'll go through and revert those changes as well!

Also, thank you for the time and energy you gave this pr for code review! We're incrementally fixing a lot of issues and as someone who just started contributing, it _feels_ good!

alanshaw (approved):

PR #842, opened 2021-11-23
fix (niftsave): fixes hanging batch process in sql

the-simian:
The batch processor was hanging for a long time even though the timer stated it was running in a matter of a few milliseconds and this was due to an inefficient select statement in the readout I was using to measure it.

Was also using this as a succinct issue to onboard @ice-breaker-tg demonstrate how Hasura migrations work


the-simian:
@alanshaw, I was going to have @ice-breaker-tg  review this just to go through the process, but I'm marking you as a reviewer too just for awareness. (for some reason I cannot add him as a reviewer though, still)

This is just for niftysave:  a small change to this batch procedure that runs on the ingestion queue table - it removes a SELECT COUNT(*) from statement that was being used to display a returned table showing the number of items remaining, and was much slower with such a large record set.

ice-breaker-tg (approved):
LGTM

PR #853, opened 2021-11-29
feat: marketplace-logo-update; add new logo, style tweaks

Codigo-Fuentes:
- Adds new holaplex logo
- tweaks spacing for true centering of "trusted by" section
- makes logos un-selectable

trusted-by


the-simian (approved):
This looks good, the only thing I'll say is if at any point these become clickable we'll want to have some sort of CSS besides user-select: none. Since they aren't clickable (at least now) I think this can ship. 

Also glad you cut out the background on that holoplex png, looks much nicer!! :)

PR #860, opened 2021-11-30
add logging to log tail

### LINKED ISSUE #725: Log monitoring solution
Evaluate and set up a monitoring solution for Cloudflare Worker logs

Related to #746

- [x] Logtail account setup
- [x] Add support for Logtail to our API code (PR pending)
- [x] Hook up heroku to Logtail
- [x] Add everyone to the account

Also need to do some of this in web3.storage

### LINKED ISSUE #857: Remove api redirects
https://github.com/nftstorage/nft.storage/blob/49d799b457b324935c63cff11e047b00064b530a/packages/api/src/errors.js#L36
https://github.com/nftstorage/nft.storage/blob/49d799b457b324935c63cff11e047b00064b530a/packages/api/src/errors.js#L79-L82

hugomrdias:
- fix: remove debug file 15b1d83ffc4e986a7b3f4690ab80511f16f80dbd
- fix: add email to db user output for logging 6a3c29c453489e4349e257d895633fe29dfc5a23
- fix: improve RouteContext handling 4588f7e47e7ba4fac33886acbaa4cb468d5cb25f
- feat: add logging and merge sentry in it 74375e14ea176faef63c86aa2579347dd5d077fb

DEBUG constant controls dev console logs and Server Timing header in the response

closes #857
closes #725


packages/api/test/scripts/worker-globals.js
```diff
@@ -11,3 +11,4 @@ export const VERSION = 'test'
 export const COMMITHASH = 'test'
 export const MAINTENANCE_MODE = 'rw'
 export const METAPLEX_AUTH_TOKEN = 'metaplex-test-token'
+export const LOGTAIL_TOKEN = ''
```
vasco-santos, line 14: Please add the docs on how this is configured for production/dev

packages/api/src/utils/logs.js
```diff
@@ -0,0 +1,286 @@
+import { nanoid } from 'nanoid/non-secure'
+
+const logtailApiURL = 'https://in.logtail.com/'
```
vasco-santos, line 3: Should we move this to the constants file?

packages/api/src/utils/logs.js
```diff
@@ -0,0 +1,286 @@
...
+
+  /**
+   * Add log entry to batch
+   *
+   * @param {any} body
+   */
+  _add(body) {
+    this.logEventsBatch.push(body)
+
+    return true
```
vasco-santos, line 108: do we need this?

vasco-santos (commented):
Overall looking good 🚀 

Minor details inline

packages/api/src/utils/logs.js
hugomrdias: this is very specific to logtail and does not change, so i think it can stay here

packages/api/test/scripts/worker-globals.js
hugomrdias: fixed

packages/api/package.json
```diff
@@ -27,6 +27,7 @@
     "debug": "^4.3.2",
```
alanshaw, line 27: Can we remove debug as a dependency?

packages/api/src/utils/logs.js
```diff
@@ -0,0 +1,284 @@
...
+   * @param {any} body
+   */
+  _add(body) {
+    this.logEventsBatch.push(body)
+  }
+
+  async postBatch() {
+    const batchInFlight = [...this.logEventsBatch]
+    this.logEventsBatch = []
+    const rHost = batchInFlight[0].metadata.request.headers.host
```
alanshaw, line 112: Is there always 1? We should guard incase there's no logs to send.

packages/api/src/utils/router.js
```diff
@@ -167,7 +169,10 @@ class Router {
       rsp = this.options.onNotFound(req)
     }
 
-    return postHandlers.reduce((r, handler) => handler(req, r), rsp)
+    const out = postHandlers.reduce((r, handler) => handler(req, r), rsp)
+
+    ctx.log.timeEnd('request')
+    return ctx.log.end(out)
```
alanshaw, line 175: What happens if this fails? Can the logging not be on the critical path - i.e. in a `waitUntil`?

packages/api/src/utils/auth.js
```diff
@@ -29,6 +26,10 @@ export async function validate(event, { sentry }) {
       if (user.data) {
         const key = user.data.keys.find((k) => k?.secret === token)
         if (key) {
+          log.setUser({
+            email: user.data.email,
```
alanshaw, line 30: I'm not sure we should be adding user emails to our logs...ID is enough to lookup their email. Do we need both?

alanshaw (changes requested):
Looks great, some minor suggestions/questions left.

packages/api/src/utils/router.js
hugomrdias: ah its not so bad, `log.end` just adds the `Server-Timing` header and returns the response the logging is done inside `waitUntil` so i think its ok to keep this as is

packages/api/src/utils/logs.js
```diff
@@ -0,0 +1,282 @@
...
+      this.opts.sentry.captureException(message)
+    } else {
+      log = {
+        ...log,
+        message,
+      }
+    }
+
+    if (this.opts?.debug) {
+      console[level](`[${dt}] `, log.message)
```
alanshaw, line 195: Can we log the context here too please?

alanshaw (changes requested):

packages/api/src/utils/logs.js
```diff
@@ -0,0 +1,282 @@
+import { nanoid } from 'nanoid/non-secure'
+
+const logtailApiURL = 'https://in.logtail.com/'
+
+const buildMetadataFromHeaders = (/** @type {Headers} */ headers) => {
+  /** @type {Record<string, string>} */
+  const responseMetadata = {}
+  Array.from(headers).forEach(([key, value]) => {
+    responseMetadata[key.replace(/-/g, '_')] = value
```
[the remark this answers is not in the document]
hugomrdias: this is taken from the logtail cloudflare workers snippet, it how it expects the entry to be

PR #874, opened 2021-12-01
chore(website): add info about size limits to http api docs

### LINKED ISSUE #832: /upload in HTTP API docs should be flagged to only accept requests that are <100MB
Should add a note on this in the [docs](https://nft.storage/api-docs/) and suggest they use a chunked CAR file upload. https://github.com/nftstorage/nft.storage/issues/413#issuecomment-918410081

Let's align on the best path to do this - think it's probably using only CLI tools like https://github.com/web3-storage/ipfs-car for CAR generation and https://github.com/nftstorage/carbites-cli/ for chunking.

Should probably add a note that /store also faces the same limits, but there's no good workaround (I think).

yusefnapora:
This adds a note about the 100MiB size limit to the `store` and `upload` endpoints in the API schema.

Closes #832

### DIFF
--- a/packages/website/public/schema.yml
+++ b/packages/website/public/schema.yml
@@ -52,6 +52,11 @@ paths:
         We highly recommend that you only use content with an `image/*` content type for your `image` field, and include other content types in the
         `properties` field as additional references.
 
+        ### Size limitations
+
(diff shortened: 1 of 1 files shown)


alanshaw (approved):

dchoi27:
Let's merge! @yusefnapora

Issue #876, opened 2021-12-02
Module not found: Can't resolve 'ipfs-car/blockstore/memory'


mjamaleddine:
When trying to upgrade to the newest version of nft storage I receive this error:

`Module not found: Can't resolve 'ipfs-car/blockstore/memory' in '\node_modules\nft.storage\src'`

Before I was using version 3.2.0 and it was fine. How can I fix this? Adding ipfs-car to the dependencies did not help.

hanlife:
![1638449952(1)](https://user-images.githubusercontent.com/21124833/144426778-37bc14fc-d4ee-406d-9381-21754e6cd93a.jpg)

yusefnapora:
Any chance you're using webpack 4? I'm investigating a similar issue that seems to be caused by webpack 4 not supporting the `exports` field in package.json.

alanshaw:
Yes, webpack 4 does not support `package.json` `exports` field so is unable to resolve the dependency.

If you can't upgrade to webpack 5, you can now (as of nft.storage@5.1.3) use the prebuilt library:

```js
import { NFTStorage } from 'nft.storage/dist/bundle.esm.min.js'
```

juliendorra:
Hi, 

if like me you have this bug because you are using Parcel, here's the issue about it at Parcel's repo:

https://github.com/parcel-bundler/parcel/issues/4155

You will find a few different workarounds in the thread.

PR #878, opened 2021-12-02
fix: add prebuilt bundle

alanshaw:
I didn't realise this didn't exist yet for nft.storage.

Now that we're pulling in more dependencies, and dependencies that are incompatible wth webpack4 and skypack we need a way for people to import the code in bundlers and on the browser.

After merge, webpack4 users should be able to:

```js
import { NFTStorage } from 'nft.storage/dist/bundle.esm.min.js'`
```

...and directly in the browser like;

```js
import { NFTStorage } from 'https://cdn.jsdelivr.net/npm/nft.storage/dist/bundle.esm.min.js'
```

refs https://github.com/nftstorage/nft.storage/issues/876

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -39,6 +39,7 @@
     "build": "run-s build:*",
     "build:ts": "tsc --build",
     "build:cjs": "rollup --config rollup.config.js",
+    "build:esm": "rollup --config rollup.esm.config.js",
     "coverage": "nyc report --reporter=text-lcov > coverage.lcov && npx codecov",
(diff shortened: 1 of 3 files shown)


yusefnapora (approved):
Sweet!

JoseGabriel-web:
So how can I fix it, I still don't get it?

develo-pera:
@alanshaw It's still not working for me, it says it "Cannot resolve file 'bundle.esm.min.js' "

Gozala:
@JoseGabriel-web @develo-pera It probably was not deployed when you tried it, it does seem to work now as you can see in this observable notebook https://observablehq.com/@protocol/import-nft-storage-in-browsers

PR #881, opened 2021-12-03
Chore (niftysave): update niftysave deps

the-simian:
Also tightens some typings in some files as well with the latest dependencies.

I'm checking each script still. It all looks correct but verifying each part for regressions.


the-simian:
@alanshaw this seems to be working ok, the only thing not all-the-way up to date is `"multiformats": "^9.4.10"`, and I locked it to the same version the rest of the project is at, can you look over these and see if anything stick out as a version you know won't work well or anything like that. Alll my scripts seem to function just fine at these versions

the-simian:
actually @Gozala  or @alanshaw It think either of you could approve this

Gozala (approved):
Looks good to me

the-simian:
@alanshaw I needed to lock a dependency that was not in niftysave, I opened an issue with more detail here:
https://github.com/ipld/js-dag-json/issues/55

Just wanted you to be aware, otherwise this is as it was when gozala reviewed it, but caught up to m ain.

rvagg:
`@ipld/dag-json@8.0.7` fixes the problem, sorry for the hiccup, I was changing the auto-releaser but accidentally left the old one in place so it was doing a borked release and then skipping the proper release 🤦 thanks for alerting me @the-simian

PR #894, opened 2021-12-03
feat: update browser examples

alanshaw:
Updates the examples to use the bundle from the jsdelivr CDN instead of skypack (we are still working through the issues with that bundler).

Added a `storeCar.html` browser example which uses the `NFTStorage.encodeBlob` function.

Also, since the bundle now works, the `car` example has been renamed to be an example of how to bundle using vite.

I also removed the `endpoint` from the examples since it's an option that most users do not need to change.


jimaek:
jsDelivr also has bundling https://www.jsdelivr.com/esm
Let us know if its missing anything you need

yusefnapora (approved):
lgtm 👍

vasco-santos (approved):
Awesome! ❤️

alanshaw:
@jimaek thank you, but it looks like it doesn't support [`package.json` `exports`](https://nodejs.org/api/packages.html#packages_exports) field yet :(

Issue #637, opened 2021-10-19
nft.storage should make sure that blocks don't exceed 1MiB


Gozala:
Larger blocks are problematic because IPFS would accept them but then fail to provide. It would gladly accept a giant json in `store` API and produce a block that could exceed 1MiB limit

https://github.com/ipfs-shipyard/nft.storage/blob/f8b56d1577f5abc1fb640bcd19f8b3e4ade03e85/packages/api/src/routes/nfts-store.js#L53-L62

### Solution

Our encoder should use chunking strategy and factor out large fields e.g. data urls into separate blocks and validate that in the end all blocks fit the limits. If this naive strategy fails it could error and ask user to refactor metadata instead.

JeffLowe:
related to #390 per Alan

dchoi27:
Yeah more precisely #645 will close this

Issue #390, opened 2021-09-03
Write CAR uploads to S3


dchoi27:
Sister issue to: https://github.com/web3-storage/web3.storage/issues/395 Some good discussion on that thread for exact implementation. Always good to have redundancy!

PR #645, opened 2021-10-21
feat: s3 backups

alanshaw:
This uploads data sent to the `/upload` endpoint to S3 for disaster recovery. All data is converted to CAR (if not CAR already) before being uploaded to S3.

Note: does not backup pinned data or IPNFT's uploaded to the `/store` endpoint

resolves https://github.com/nftstorage/nft.storage/issues/390
resolves https://github.com/nftstorage/nft.storage/issues/637
resolves https://github.com/nftstorage/nft.storage/issues/355

PR #899, opened 2021-12-06
feat: s3 backups

alanshaw:
This uploads data sent to the `/upload` endpoint to S3 for disaster recovery. All data is converted to CAR (if not CAR already) before being uploaded to S3.

Note: does not backup pinned data or IPNFT's uploaded to the `/store` endpoint

resolves https://github.com/nftstorage/nft.storage/issues/390
resolves https://github.com/nftstorage/nft.storage/issues/637
resolves https://github.com/nftstorage/nft.storage/issues/355

supersedes https://github.com/nftstorage/nft.storage/pull/645


packages/api/src/routes/nfts-upload.js
```diff
@@ -21,78 +25,212 @@ export async function nftUpload(event, ctx) {
...
     const blob = await event.request.blob()
     if (blob.size === 0) {
-      throw new HTTPError('Empty payload', 400)
+      throw new HTTPError('empty payload', 400)
     }
     const isCar = contentType.includes('application/car')
+    const uploadType = isCar ? 'Car' : 'Blob'
+    // If this upload is a CAR file it is not necessarily complete. i.e. it may
+    // be a chunked CAR.
+    const structure = isCar ? 'Unknown' : 'Complete'
```
vasco-santos, line 63: nit: Given we already to this validation below, we could use the same as car, create a let and use the conditions below.

vasco-santos (approved):
LGTM

alanshaw:
TODO: `UNIQUE` constraint needs to be changed to apply to `upload_id` AND `url`. Backups of historical data are _complete_ CAR files and are saved as `complete/<CID>.car`, so two different uploads may have the same URL.

Issue #877, opened 2021-12-02
Documents rate limits 


hugomrdias:
We need to document the rate limits and try to make them per use instead of global.

JeffLowe:
@yusefnapora is this one handled by #902 ?  If so I'll reassign to you and connecte the issue to the PR

dchoi27:
@hugomrdias planning on syncing with Cloudflare on rate limits, and IP-based vs. token-based

alanshaw:
Cloudflare need to give us the ability to set the status, content-type and payload for rate limiting per `Authorization` HTTP header.

Until then I have a PR open to workaround that and improve the error message when we receive a 429 status: https://github.com/nftstorage/nft.storage/pull/1154

I have switched to Authorization header based limiting as detailed here: https://github.com/nftstorage/nft.storage/issues/935#issuecomment-1020278468

PR #902, opened 2021-12-06
chore(website): add info about rate limits to http docs

yusefnapora:
Adds a note about rate limits to the HTTP spec (copied from this similar [w3s PR](https://github.com/web3-storage/docs/pull/216)), but I'm not 100% sure it's accurate, since #877 suggests that we have global limits here instead of per-token limits.


alanshaw:
Yes they are _currently_ set per IP not per auth token but it'll be correct soon.

alanshaw (approved):

alanshaw:
I believe this is now correct!

Issue #823, opened 2021-11-19
NFT.Storage gateway cache MVP


dchoi27:
We want to provide users with an HTTP endpoint that can fetch any data off the public IPFS network, but be super performant with NFT.Storage data in particular.

The current work plan consists on:

- [x] gateway.nft.storage naive implementation https://github.com/nftstorage/nft.storage/issues/867
- [x] setup Pinata gateway https://github.com/nftstorage/nft.storage/issues/869
- [x] gateway race with ipfs.io + CF + Pinata https://github.com/nftstorage/nft.storage/issues/868
- [x] setup logtail https://github.com/nftstorage/nft.storage/issues/1036
- [x] setup first Cache layer - CF caching on http request - validation - nftstorage/nft.storage#1180 +  nftstorage/nft.storage#1181
- [x] track and prevent rate limiting on IPFS public gateways https://github.com/nftstorage/nft.storage/issues/1165 
- [x] load test with gateway logs https://github.com/nftstorage/nft.storage/issues/870
- [x] `/ipfs/` route to gateway `nft.storage/ipfs/<cid>` (just a redirect) https://github.com/nftstorage/nft.storage/issues/1049
- [x] Setup rate limit on CF Worker https://github.com/nftstorage/nft.storage/issues/1196
- [x] Get 2-5 users to sign up for testing
- [x] **Public Launch 🎉**

Further plans on the pipeline:
- [x] Website and client with Gateway domain URL https://github.com/nftstorage/nft.storage/issues/1345
- [ ] Track wether requested content is stored in DB nftstorage/nftstorage.link#14
  - [ ] Have cache in API check https://github.com/nftstorage/nft.storage/pull/1391
- [ ] Augment gateway metrics nftstorage/nftstorage.link#18
- [ ] Add all nice metrics to Grafana
- [ ] Improve rate limit on the integrated public gateways https://github.com/nftstorage/nft.storage/issues/1195
- [ ] (should have rate limit improvements in place) warm cache + async gateway read test https://github.com/nftstorage/nft.storage/issues/755
- [ ] (should have rate limit improvements in place) async gateway read test alerts on performance degradation https://github.com/nftstorage/nft.storage/issues/871

Other ideas:
- [ ] Setup second Cache layer
  - Consider S3 with "complete" metadata flag
- [ ] add S3 as gateway alternative
- [ ] Analyze logs and optimize caching based on user patterns
[...]

PR #908, opened 2021-12-07
feat: nft.storage naive gateway implementation

### LINKED ISSUE #787: Add Indexes to foreign keys
We should add indexed for foreign keys given the aim of them is to link data together and use them in a query. We currently not use all of them, but that might change and indexes are cheap

### LINKED ISSUE #867: gateway.nft.storage - CF Worker
> As part of https://github.com/nftstorage/nft.storage/issues/823 epic

Setup a CF Worker package to act as the entry point for the gateway.nft.storage DNS.

Requirements:
- Use ESM
- Support IPFS subdomain `<cid>.ipfs.nft.storage`
- Setup deployment actions

vasco-santos:
This PR adds a naive implementation of the gateway as a CF Worker using subdomains, where it simply forwards requests to the IPFS public gateway for now as part of https://github.com/nftstorage/nft.storage/issues/823

It includes:
- worker implementation with ESM
- test setup with miniflare
- actions for tests / release

ACM was configure in the TLS setup for the nft.storage domain, so that we can do:

- `bafkreidchi5c4c3kwr5rpkvvwnjz3lh44xi2y2lnbldehwmpplgynigidm.ipfs.nft.storage`
- `bafkreidchi5c4c3kwr5rpkvvwnjz3lh44xi2y2lnbldehwmpplgynigidm.ipfs-staging.nft.storage`
- `bafkreidchi5c4c3kwr5rpkvvwnjz3lh44xi2y2lnbldehwmpplgynigidm.ipfs-staging.nft.storage/path`

Besides follow up work tracked on #823 needs:

- follow up PR with sentry configured

Closes #867


packages/gateway/wrangler.toml
```diff
@@ -0,0 +1,39 @@
...
+[build.upload]
+format = "service-worker"
+
+# PROD!
+[env.production]
+# name = "gateway-nft-storage-production"
+account_id = "fffa4b4363a7e5250af8357087263b3a" # Protocol Labs CF account
+zone_id = "fc6cb51dbc2d0b9a729eae6a302a49c9"    # nft.storage zone
+route = "*gateway.nft.storage/*"
+vars = { IPFS_GATEWAY = "https://ipfs.io" }
```
vasco-santos, line 23: Note: We will need to setup multiple gateways, either we need to hardcode multiple here or we will need to use a string array and parse it. Leaving for when needed

packages/gateway/src/errors.js
```diff
@@ -0,0 +1,25 @@
+export class InvalidIpfsPathError extends Error {
+  /**
+   * @param {string} cid
+   */
+  constructor(cid) {
+    super(`invalid ipfs path: invalid path "/ipfs/${cid}/"`)
+    this.name = 'InvalidIpfsPath'
```
alanshaw, line 7:
```suggestion
    this.name = 'InvalidIpfsPathError'
```

For consistency with `AbortError`:

> Note: When `abort()` is called, the `fetch()` promise rejects with an `Error` of type `DOMException`, with name `AbortError`.

https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort

packages/gateway/src/errors.js
```diff
@@ -0,0 +1,25 @@
...
+}
+InvalidIpfsPathError.CODE = 'ERROR_INVALID_IPFS_PATH'
+
+export class InvalidUrlError extends Error {
+  /**
+   * @param {string} url
+   */
+  constructor(url) {
+    super(`invalid url: ${url}`)
+    this.name = 'InvalidUrl'
```
alanshaw, line 20:
```suggestion
    this.name = 'InvalidUrlError'
```

packages/gateway/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+ * @property {string} IPFS_GATEWAY
+ */
+
+/**
+ * Handle gateway request
+ * @param {Request} request
+ * @param {Env} env
+ */
+async function handleRequest(request, env) {
+  const publicGatewayUrl = new URL('ipfs', env.IPFS_GATEWAY)
```
alanshaw, line 16: Can be re-used for every request.

packages/gateway/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+  const url = new URL(request.url)
+  const cid = getCidFromSubdomainUrl(url.hostname)
+  const response = await fetch(
+    `${publicGatewayUrl.toString()}/${cid}${url.pathname || ''}`
+  )
```
alanshaw, line 21: I _think_ just putting it in a string will automatically call `toString`...

```suggestion
  const response = await fetch(`${publicGatewayUrl}/${cid}${url.pathname || ''}`)
```

packages/gateway/src/utils/cid.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Parse subdomain URL and return cid
+ *
+ * @param {string} url
+ */
+export function getCidFromSubdomainUrl(url) {
+  // Replace "ipfs-staging" by "ipfs" if needed
+  const nUrl = url.replace('ipfs-staging', 'ipfs')
+  const splitUrl = nUrl.split('.ipfs.')
```
alanshaw, line 13: At this point is `splitUrl` not `['https://bafy...', 'nft.storage']`? `splitUrl[0]` will fail to pass as a CID.

packages/gateway/src/utils/json-response.js
```diff
@@ -0,0 +1,16 @@
...
+  constructor(body, init = {}) {
+    init.headers = init.headers || {}
+    init.headers['Content-Type'] = 'application/json;charset=UTF-8'
+    super(JSON.stringify(body), init)
+  }
+}
+
+export function notFound(message = 'Not Found') {
+  return new JSONResponse({ message }, { status: 404 })
+}
```
alanshaw, line 16: I think we don't need JSON responses.

packages/gateway/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "type": "module",
+  "module": "./dist/index.mjs",
+  "scripts": {
+    "build": "node scripts/cli.js build",
+    "dev": "miniflare --watch --debug",
+    "deploy": "wrangler publish --env production",
+    "pretest": "npm run build ",
+    "test": "npm-run-all -p -r mock:ipfs.io test:worker",
+    "test:worker": "ava --verbose test/*.spec.js",
+    "mock:ipfs.io": "smoke -p 9081 test/mocks/ipfs.io"
```
alanshaw, line 15: Can we spin up an IPFS node instead? We can use the IPFS HTTP API to add content that we want to fetch.

alanshaw (changes requested):
❤️ Awesome work so far!

packages/gateway/src/index.js
vasco-santos: It would be nice, but can it? `IPFS_GATEWAYS` are dynamically set and the env context is only accessible on request handler. Any way you aware?

packages/gateway/src/utils/cid.js
vasco-santos: Ok, the parameter name was unfortunate, it is the hostname:

```js
getCidFromSubdomainUrl(url.hostname)
```

So, I guess just rename the parameter should be enough

packages/gateway/package.json
vasco-santos: Let's do this in the follow up PR (Race), I just commented to remember: https://github.com/nftstorage/nft.storage/pull/961#pullrequestreview-845775128

That's because I need to have tests where I can manipulate who is faster. In that PR it will be easier to do that, perhaps using a real ipfs node + smoke tests for other gateways

packages/gateway/src/utils/json-response.js
vasco-santos: Maybe for CID status, but removing until needed

packages/gateway/package.json
alanshaw: Yes fine to do in follow up PR 😄 

> That's because I need to have tests where I can manipulate who is faster. In that PR it will be easier to do that, perhaps using a real ipfs node + smoke tests for other gateways

Maybe just a simple proxy server that adds a delay? I find smoke to have quite a big overhead in terms of knowing what to call the file, what the response object should look like, how to get requests parsed correctly etc.

packages/gateway/src/error-handler.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @param {Error & {status?: number;code?: string;}} err
+ */
+export function errorHandler(err) {
+  // TODO: setup sentry
+  console.error(err.stack)
+
+  let status = err.status || 500
```
alanshaw, line 8:
```suggestion
  const status = err.status || 500
```

packages/gateway/src/error-handler.js
```diff
@@ -0,0 +1,11 @@
...
+
+  return new Response(err.message || 'Server Error', { status })
```
alanshaw, line 10: It would be helpful to show the full stack here if not production.

packages/gateway/src/utils/cid.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param {string} cid
+ */
+export function normalizeCid(cid) {
+  try {
+    const c = CID.parse(cid)
+    return c.toV1().toString()
+  } catch (err) {
+    throw new InvalidIpfsPathError(cid)
+  }
```
alanshaw, line 33: We shouldn't throw this error here, it's an invalid CID here, and should be caught in `getCidFromSubdomainUrl` and thrown with an appropriate error.

```suggestion
  const c = CID.parse(cid)
  return c.toV1().toString()
```

packages/gateway/src/utils/cid.js
```diff
@@ -0,0 +1,34 @@
...
+export function getCidFromSubdomainUrl(url) {
+  // Replace "ipfs-staging" by "ipfs" if needed
+  const nUrl = url.hostname.replace('ipfs-staging', 'ipfs')
+  const splitUrl = nUrl.split('.ipfs.')
+
+  if (!splitUrl.length) {
+    throw new InvalidUrlError(url.hostname)
+  }
+
+  return normalizeCid(splitUrl[0])
```
alanshaw, line 19: It's not really an invalid "IPFS path" it's an invalid CID in the hostname of the URL, I'd probably do:

```suggestion
  try {
    return normalizeCid(splitUrl[0])
  } catch (err) {
    throw new InvalidUrlError(`invalid CID: ${splitUrl[0]}: ${err.message}`)
  }
```

...you'd need to also change `InvalidUrlError` to accept a message instead of a URL.

packages/gateway/src/errors.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+}
+InvalidIpfsPathError.CODE = 'ERROR_INVALID_IPFS_PATH'
+
+export class InvalidUrlError extends Error {
+  /**
+   * @param {string} url
+   */
+  constructor(url) {
+    super(`invalid url: ${url}`)
```
alanshaw, line 19: Probably don't need to echo the URL back...

```suggestion
  /**
   * @param {string} message
   */
  constructor(message = 'invalid URL') {
    super(message)
```

packages/gateway/src/utils/cid.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Parse subdomain URL and return cid
+ *
+ * @param {URL} url
+ */
+export function getCidFromSubdomainUrl(url) {
+  // Replace "ipfs-staging" by "ipfs" if needed
+  const nUrl = url.hostname.replace('ipfs-staging', 'ipfs')
+  const splitUrl = nUrl.split('.ipfs.')
```
alanshaw, line 13:
```suggestion
  const host = url.hostname.replace('ipfs-staging', 'ipfs')
  if (!host.includes('.ipfs.')) {
    throw new InvalidUrlError('missing IPFS subdomain in host')
  }
  const splitHost = host.split('.ipfs.')
```

packages/gateway/src/utils/cid.js
```diff
@@ -0,0 +1,34 @@
...
+
+  if (!splitUrl.length) {
+    throw new InvalidUrlError(url.hostname)
+  }
```
alanshaw, line 17: It will always have a length

packages/gateway/src/errors.js
```diff
@@ -0,0 +1,25 @@
+export class InvalidIpfsPathError extends Error {
```
alanshaw, line 1: I think we don't need this error (see other feedbacks).

alanshaw (approved):
A few minor tidyup things but this is good to go otherwise!

PR #909, opened 2021-12-07
chore: update licences to dual mit and apache

vasco-santos:


hugomrdias (approved):

ribasushi:
@vasco-santos in golang land we use a unified `LICENSE.md`, instead of the two license files as in this repo. The benefit of doing so is that github's own `License` link will go the the right both-license file.

Encourage you to just copy ours: https://github.com/ipfs/ipfs-repository-template/blob/main/LICENSE.md

Background ( do not copy that one - it has wrong `ipfs.org` links )
https://github.com/ipld/go-car/pull/59/files

LICENSE.md
```diff
@@ -0,0 +1,232 @@
+The contents of this repository are Copyright (c) corresponding authors and
+contributors, licensed under the `Permissive License Stack` meaning either of:
+
+- Apache-2.0 Software License: https://www.apache.org/licenses/LICENSE-2.0
+  ([...4tr2kfsq](https://gateway.ipfs.io/ipfs/bafkreiankqxazcae4onkp436wag2lj3ccso4nawxqkkfckd6cg4tr2kfsq))
+
+- MIT Software License: https://opensource.org/licenses/MIT
+  ([...vljevcba](https://gateway.ipfs.io/ipfs/bafkreiepofszg4gfe2gzuhojmksgemsub2h4uy2gewdnr35kswvljevcba))
```
ribasushi, line 8: Whooops you copied the wrong one ;) 
We want the `dweb.link` ones, like here: https://raw.githubusercontent.com/ipfs/ipfs-repository-template/main/LICENSE.md

vasco-santos: ah, I got confused by the suggestion and thought it was the opposite 😆

ribasushi (approved):
Perfect!

packages/api/package.json
```diff
@@ -16,7 +16,7 @@
     "db-types": "./scripts/cli.js db-types"
   },
   "author": "Hugo Dias <hugomrdias@gmail.com> (hugodias.me)",
-  "license": "MIT",
+  "license": "(Apache-2.0 AND MIT)",
```
ribasushi, line 19: @vasco-santos actually one last question ( sorry for being such a PITA!!! )

Is this field supposed to be a SPDX identifier? If yes it should read ( like what you just copied ) `Apache-2.0 OR MIT`

vasco-santos: Yes, you are right! https://docs.npmjs.com/cli/v8/configuring-npm/package-json#license

We have AND in all the repos in nft and web3, but it should be as you suggest

ribasushi (approved):
@vasco-santos I tried really hard to find more problems with this PR
I am sorry.
I failed you.

:shipit:

PR #914, opened 2021-12-08
fix (website): safari buttons for mobile menu

the-simian:
fixes the way the mobile menu looks in safari


the-simian:
Got this looking correct now: 
image

the-simian:
@dashcraft come check this out and review it

dashcraft (approved):
LGTM!

dashcraft:
It feels like the interface or rendering pattern might be a few more iterations from settling down, but this feels much better and solves the safari issue.

Issue #912, opened 2021-12-07
Blog cleanup


dchoi27:
We launched the blog, woo! Some cleanup items that come to mind on https://nft.storage/blog/
- [x] Could we make sure the title on the spotlighted article is also a link to the post?
- [ ] Could we include the spotlighted article under the list of all blog posts at the bottom?

PR #916, opened 2021-12-08
feat: blog cleanups

the-simian:
Addresses some issues in #912 
- Title link leads to blog article when you click it
- include the spotlighted article below in the list (do not exclude it by default)


dashcraft (approved):
Code seems sane, approved as long as cloudflare builds adequately.

PR #922, opened 2021-12-08
fix: remove dagcargo materialized views

alanshaw:
Sister PR to https://github.com/web3-storage/web3.storage/pull/735

The dagcargo materialized views are taking too long to refresh and are timing out. The entries table is just too big now. We could maybe increase the fetch size to temporarily fix but instead we decided to just remove the materialized views and make the API tolerant of failure.


vasco-santos (approved):
LGTM

Issue #905, opened 2021-12-06
Add Makersplace logo to homepage + make Holaplex logo transparent


dchoi27:
https://makersplace.com/static/img/logo-main-square.png

And let's make the Holaplex logo transparent as well!
## Use these transparent logos
- [ ]  Holaplex  https://drive.google.com/file/d/1C6YlYXcVxaGwO6qpoWpaVguMG8KNJujr/view?usp=sharing
- [ ] Makersplace https://drive.google.com/file/d/1cAFtjURvuQ-jvd_yEc9SsKcoSh2fR16O/view?usp=sharing

PR #934, opened 2021-12-09
fix: Add Makersplace logo to homepage + make Holaplex logo transparent

redaphid:
Can Aaron add logos to a website? And other burning questions related to #905


alanshaw (approved):
LGTM

PR #943, opened 2021-12-10
fix: increase website token ttl and update deps

hugomrdias:
- increased token ttl to allow longer uploads
- remove one extra call to magic.link
- updated next and others


packages/website/pages/index.js
```diff
@@ -9,15 +9,15 @@ import Link from 'next/link'
...
+  const logos = fs.readdirSync('public/images/marketplace-logos')
   // make opensea be the first logo
   logos.sort((a, b) =>
     a.includes('opensea') ? -1 : b.includes('opensea') ? 1 : 0
   )
 
   return {
     props: {
-      needsUser: false,
+      needsUser: true,
```
alanshaw, line 20: It does?

packages/website/lib/magic.js
```diff
@@ -50,14 +50,13 @@ export async function login(token, type = 'magic', data = {}, version = '') {
...
-  if (isLoggedIn) {
+  try {
     const meta = await getMagic().user.getMetadata()
     return {
       ...meta, // we dont actually need the user info
     }
-  } else {
-    return undefined
+  } catch (err) {
+    // do nothing
```
alanshaw, line 59:
```suggestion
  } catch {
    // do nothing
```

packages/website/lib/api.js
```diff
@@ -3,7 +3,7 @@ import constants from './constants'
 
 export const API = constants.API
 
-const LIFESPAN = 900
+const LIFESPAN = 60 * 60 * 2 // 2 hours
```
alanshaw, line 6: Is there a way to auto renew the token automatically if the user is still on the page?

This will do for now but won't fix for really big uploads.

alanshaw (approved):
LGTM just a couple of questions.

packages/website/lib/api.js
hugomrdias: not really, we gonna need to discuss this further theres no good option right now

packages/website/pages/index.js
hugomrdias: yes and no, it allows us to skip one magic round-strip in the callback route and avoid the layout shift of landing in the homepage logged out for a bit and re-use session after the request.

alanshaw (approved):

PR #948, opened 2021-12-10
feat(api): add mintingAgent tag to metaplex token

yusefnapora:
This updates the metaplex endpoint to require a `mintingAgent` tag (added in https://github.com/nftstorage/metaplex-auth/pull/22), and pulls the `agentVersion` tag out of the payload if it's present.

The metaplex-auth PR above also changes the name of the `solana-cluster` tag to `solanaCluster`, since I originally used kebab-case for some reason and now regret my life choices. So this PR updates the backend to read the new tag. I was thinking we could briefly support both, but since we don't have any real traffic here yet it seems alright to just make the breaking change, especially since requests made with the old version would fail anyway due to the missing `mintingAgent` tag.

### DIFF
--- a/packages/api/src/routes/metaplex-upload.js
+++ b/packages/api/src/routes/metaplex-upload.js
@@ -110,6 +110,8 @@ async function validate(db) {
  * @property {string} iss the public key of the uploader, as a key:did string
  * @property {string} rootCID the root CID of the upload, as a CIDv1 string
  * @property {string} solanaCluster the name of the solana cluster being targetted
+ * @property {string} mintingAgent string identifying the "user agent" that prepared the upload
+ * @property {string} [agentVersion] optional string identifying which version of the `mintingAgent` was used
  *
  * @param {string} token - an encoded JWT token from an x-web3auth header.
  * @returns {Promise<MetaplexMetadata>} - metadata extracted from token payload.
@@ -147,7 +149,6 @@ async function parseMetaplexJWT(token) {
   }
 
   // validate header
-
   if (header.alg !== 'EdDSA') {
     throw new ErrorInvalidMetaplexToken('invalid signing algorithm')
   }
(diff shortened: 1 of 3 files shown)


yusefnapora:
I realized that it's worth "doing the right thing" here after all, since we probably want to have this merged to staging for a bit before we push to production. So the last commit makes `mintingAgent` optional (with default to "unknown" if missing), and accepts the old `solana-cluster` tag name if `solanaCluster` is not set.

This should let the backend work with either kind of token, so we can merge this PR independently of publishing the client package.

packages/api/src/routes/metaplex-upload.js
```diff
@@ -172,21 +173,39 @@ async function parseMetaplexJWT(token) {
...
+  // temporarily support old kebab-case tag name
+  const solanaCluster =
+    payload.req.put.tags.solanaCluster || payload.req.put.tags['solana-cluster']
+  if (typeof solanaCluster !== 'string') {
+    throw new ErrorInvalidMetaplexToken(
+      '"solanaCluster" tag not present in payload'
+    )
+  }
+
+  const { mintingAgent, agentVersion } = payload.req.put.tags
```
Gozala, line 189: Suggestion: This all would read a lot simpler if you refactor things a bit e.g.

```js
/**
  * @param {{mintingAgent?:unknown}} input
  * @returns {string}
  */
const parseMintingAgent = ({mintingAgent}) => {
  if (mintingAgent != null && typeof mintingAgent !== 'string') {
    throw new ErrorInvalidMetaplexToken(
      '"mintingAgent" tag must have a string value if present'
    )
  }
  // Also normalizes '' to 'unknown'
  return mintingAgent || 'unknown'
}

const parseAgentVersion = ({agentVersion}) => { ... }
const parseAgen = (input) => {{
   mintingAgent: parseMintingAgent(input),
   agentVersion: parseAgentVersion(input)
}
```

Gozala (approved):
I am not really familiar with this code, but from little I gathered looking at PR it looks good to me! Made minor suggestions, feel free to address or disregard

PR #959, opened 2021-12-12
feat: deals-ui stand up

Codigo-Fuentes:
- adds deals ui


packages/website/pages/files.js
```diff
@@ -96,6 +101,118 @@ export default function Files({ user }) {
...
+    if (queuedDeals.length) {
+      const message = `The content from this upload has been aggregated for storage on Filecoin and is queued for deals with ${
+        queuedDeals.length
+      } storage provider${
+        queuedDeals.length > 1 ? 's' : ''
+      }. Filecoin deals will be active within 48 hours of upload.`
+      deals.push(
+        <span key={nft.cid + '-pending'}>
+          {`${deals.length ? ', ' : ''}${queuedDeals.length} pending`}
+          <Tooltip
```
dashcraft, line 146: Tooltips are tricky, you'll need to make sure the parent element that has the tooltip applied to it, has the correct aria attributes:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role

packages/website/pages/files.js
```diff
@@ -96,6 +101,118 @@ export default function Files({ user }) {
 
   const hasZeroNfts = nfts.length === 0 && befores.length === 1
 
+  /**
+   * @param {any} nft
+   */
+  const TableItem = ({ nft }) => {
```
dashcraft, line 107: TableItem seems to be large and decently complex, we should probably move this to it's own file/seperate component to reduce the noise in this one.

packages/website/pages/files.js
```diff
@@ -96,6 +101,118 @@ export default function Files({ user }) {
...
+            overlay={<span>{message}</span>}
+            overlayClassName="table-tooltip"
+          >
+            <VscQuestion size={16} className="ml2" />
+          </Tooltip>
+        </span>
+      )
+    }
+
+    if (!nft.deals.length) {
```
dashcraft, line 157: Having some if statements mutate an array seems iffy, this is subjective, but it feels like we could probably
move some of this visual complexity into the returned dom and avoid the need to mutate the deals array.

packages/website/pages/_app.js
```diff
@@ -20,7 +20,7 @@ const queryClient = new QueryClient({
  */
 export default function App({ Component, pageProps }) {
   useEffect(() => {
-    Router.events.on('routeChangeComplete', route => {
+    Router.events.on('routeChangeComplete', (route) => {
```
dashcraft, line 23: linter?

packages/website/next.config.js
```diff
@@ -17,7 +17,7 @@ const withBundleAnalyzer = require('@next/bundle-analyzer')({
 const nextConfig = withBundleAnalyzer({
   trailingSlash: true,
   reactStrictMode: true,
-  exportPathMap: async function() {
+  exportPathMap: async function () {
```
dashcraft, line 20: linter?

packages/website/components/tooltip.js
```diff
@@ -0,0 +1,32 @@
...
+  children,
+  placement = 'top',
+  overlay,
+  destroyTooltipOnHide,
+  overlayClassName = '',
+  ...props
+}) => (
+  <RCTooltip
+    placement={placement}
+    overlay={<div className="bg-white p2 f6 flex">{overlay}</div>}
```
dashcraft, line 22: I would set a default overlay, but allow an overlay to be passed into this via props as a component.
Right now you just have overlay content, instead, make it the full overlay component.

pseudocode
```
/*
props.overlay= (<div className="bg-white p2 f6 flex">Some Text Here</div>)
*/

overlay={props.overlay}
```

packages/website/pages/files.js
```diff
@@ -44,6 +46,9 @@ export default function Files({ user }) {
   /** @type {[string, { before: string, limit: number }]} */
   const queryKey = ['get-nfts', queryParams]
 
+  let isDev
+  if (!!globalThis.window && location.host === 'localhost:4000') isDev = true
```
dashcraft, line 50: Comment:
Is this how we find the environment of this app? I thought there should be an env variable set somewhere that would make this much easier.

packages/website/pages/index.js
```diff
@@ -314,7 +314,7 @@ function About() {
 }
 
 function GettingStarted() {
-  const onClickHandler = useCallback(event => {
+  const onClickHandler = useCallback((event) => {
```
dashcraft, line 317: linter?

dashcraft (changes requested):
Should be minor changes, nothing massive, great work!

packages/website/components/tooltip.js
dashcraft: This is a refactor request, we can do this another time.

packages/website/pages/files.js
dashcraft: This is a request for more info, can be solved another time.

packages/website/pages/files.js
dashcraft: This is a refactor request, can be answered another time.

packages/website/pages/files.js
dashcraft: Refactor request, can be solved another time.

alanshaw (changes requested):
Looks awesome! Can the storage provider deal links be black instead of blue?

Codigo-Fuentes:
> Looks awesome! Can the storage provider deal links be black instead of blue?

Changed

dashcraft (approved):

JeffLowe:
@alanshaw just pinging you on this.  If change request looks good, we can get this merged and out.

alanshaw (approved):
LGTM - small suggestion though can date, pin status and size columns be `nowrap`?

PR #984, opened 2021-12-15
chore: update pgrest to v9

### LINKED ISSUE #883: Update heroku buildpack to postgrest v9

hugomrdias:
- updates pgrest to v9
- changes db url to heroku's instead of cloudflare its just faster

closes #883


packages/infra/heroku/init.sh
```diff
@@ -59,21 +59,32 @@ heroku pg:credentials:create nft-storage-prod-0 --name=nft_storage --app=nft-sto
...
-heroku config:set DB_URI=$(heroku config:get DATABASE_URL --app=nft-storage-prod) --app=nft-storage-pgrest-prod
-# Obtain secret from 1password vault!
-heroku config:set JWT_SECRET="supersecret" --app=nft-storage-pgrest-staging
-heroku config:set JWT_SECRET="supersecret" --app=nft-storage-pgrest-prod
+# Attach databases to apps
+heroku addons:attach nft-storage-prod-0 --app=nft-storage-prod
+heroku addons:attach nft-storage-replica-0 --app=nft-storage-prod
+heroku addons:attach nft-storage-staging-0 --app=nft-storage-staging
+
+# PGREST config vars
```
vasco-santos, line 67: Does this not work with the config files anymore?

hugomrdias: it still works with config files, but i feel the `heroku git:remote` step is super error prone and we cant have different configs for prod and staging in the same repo. So i think its best to just use the config vars until we find a better solutions.

vasco-santos (approved):

vasco-santos (approved):

PR #991, opened 2021-12-16
docs: add cidv0 example

alanshaw:
`ipfs-car@0.6.1` now supports the `rawLeaves` option.

It means that if absolutely necessary, a v0 CID can be obtained for a DAG by downgrading the CID returned from the library (note this is only possible by using `rawLeaves: false`).

How far do we want to go with this? We could stop here, or we could add it as an option to the client that `encodeDirectory` and `encodeBlob` might take - the example becomes simplier but we'll need to release a new version of the client. My gut instinct is to leave it like this.


dchoi27:
@alanshaw should we merge this?

vasco-santos (approved):
LGTM 🎉

PR #997, opened 2021-12-18
feat: add copy icon to files for copying cid

### LINKED ISSUE #969: Add copy function next to CID in Files page
Right now, the CID in the Files page is a link (making it difficult to highlight), and there's no copy function, meaning copying the link is tough. 

Let's add a copy function next to each CID listed in the Files table. Let's offer two options: copy CID, and copy IPFS URL (which is just `ipfs://<cid>`). Would love input if the most recognizable pattern / easiest way to do this this is two separate buttons next to each CID, or a single copy button with a menu. The former is probably better for mobile, but the latter is probably cleaner.

EDIT: Let's go with the menu option.

Note:  Model UI approach as seen on ipfs

dashcraft:
Uses existing copy-to-clipboard package to copy cid to clipboard for easier copying and pasting of cid.

Fixes #969

### DIFF
--- a/packages/website/components/copyButton.js
+++ b/packages/website/components/copyButton.js
@@ -49,6 +49,11 @@ function CopyButton({ text, title, popupContent, ...props }) {
         data-content={popupContent}
         className={clsx('tool', { active: tool })}
       />
+      {tool && (
+        <span className="sr-only" aria-live="assertive">
+          {popupContent}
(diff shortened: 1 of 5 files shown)


dashcraft:
image

cmunns (approved):
Pulled locally and it works great.

PR #999, opened 2021-12-19
Feat/simple-dev

hugomrdias:
I have been working on our dev setup and im glad to announce that we can now run website and API without Cloudflare account
The only account we need now is a personal free magic.link nothing else!
Also the local docker compose setup includes ipfs cluster :tada:
No more tunnels, wrangler, personal cloudflare account and settings, cluster mocks or out of band cluster docker setup.

Merry Christmas :santa: :christmas_tree:

- adds real cluster to local dev
- removes wrangler dev with the help of miniflare
- some other tweaks and improvements


README.md
```diff
@@ -59,14 +59,45 @@ We use `yarn` in this project and commit the `yarn.lock` file.
...
+SENTRY_UPLOAD=false
+
+## API Postgrest
+DATABASE_URL=http://localhost:3000
+DATABASE_TOKEN=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJzdXBhYmFzZSIsImlhdCI6MTYwMzk2ODgzNCwiZXhwIjoyNTUwNjUzNjM0LCJyb2xlIjoic2VydmljZV9yb2xlIn0.necIJaiP7X2T2QjGeV-FhpkizcNTX8HjDDBAxpgQTEI
+
+# Postgres Database
+DATABASE_CONNECTION=postgresql://postgres:postgres@localhost:5432/postgres
+
+# Vars below this line are optional or can have fake values
```
vasco-santos, line 98: Can't sentry be optional as well?

packages/api/README.md
```diff
@@ -2,116 +2,33 @@
 
 The nft.storage public API.
 
-## Getting Started
-
-### Local env vars
+## Usage
 
-Follow instructions in the root `README.md`.
+Make sure you already did step 1 and 2 from these [instructions](/#getting-started).
```
vasco-santos, line 7:
```suggestion
Make sure you already did step 1 and 2 from the [root instructions](/#getting-started).
```
Let's be more clear to avoid people to need to navigate if they know they just did it

packages/api/db/docker/docker-compose.yml
```diff
@@ -26,3 +26,28 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
       POSTGRES_PORT: 5432
+  ipfs0:
```
vasco-santos, line 29: Can we have this in a different docker file? There are a quite of reasons to separate it:

- this is quite strange to have it in the db folder (I would suggest adding it to https://github.com/nftstorage/nft.storage/tree/main/packages/tools
- for local development testing, might be import to use a persistent local DB (previously populated with some data), instead of keeping starting and stopping one, but we still would like to run cluster container.
- we might want to debug a cluster instance

packages/api/db/docker/docker-compose.yml
```diff
@@ -26,3 +26,28 @@ services:
...
+    depends_on:
+      - ipfs0
+    environment:
+      CLUSTER_PEERNAME: cluster0
+      CLUSTER_SECRET: ${CLUSTER_SECRET} # From shell variable if set
+      CLUSTER_IPFSHTTP_NODEMULTIADDRESS: /dns4/ipfs0/tcp/5001
+      CLUSTER_CRDT_TRUSTEDPEERS: '*' # Trust all peers in Cluster
+      CLUSTER_RESTAPI_HTTPLISTENMULTIADDRESS: /ip4/0.0.0.0/tcp/9094 # Expose API
+      CLUSTER_RESTAPI_BASICAUTHCREDENTIALS: test:test
+      CLUSTER_RESTAPI_CORSALLOWEDMETHODS: GET,POST,OPTIONS
```
vasco-santos, line 47: Surprised that this worked, I got CORS issues trying this in web3.storage, not sure why

packages/api/package.json
```diff
@@ -6,12 +6,11 @@
...
   "scripts": {
-    "lt": "npm run lt:cluster & npm run lt:ipfs-proxy & npm run lt:postgrest",
-    "lt:cluster": "npx localtunnel --port 9094 --subdomain \"$(whoami)-cluster-api-nft-storage\"",
-    "lt:ipfs-proxy": "npx localtunnel --port 9095 --subdomain \"$(whoami)-ipfs-proxy-api-nft-storage\"",
-    "lt:postgrest": "npx localtunnel --port 3000 --subdomain \"$(whoami)-postgrest-nft-storage\"",
     "deploy": "wrangler publish --env production",
-    "dev": "wrangler dev",
+    "predev": "./scripts/cli.js db --start && ./scripts/cli.js db-sql --cargo --testing --reset",
+    "dev": "miniflare --watch --debug --env ../../.env",
+    "clean": "./scripts/cli.js db --clean && rm -rf db/docker/compose",
```
vasco-santos, line 12: let's add a "postdev" script for cleanup?

vasco-santos (changes requested):
Great work 🎉

README.md
hugomrdias: No, it's for miniflare the global needs to be set to something

packages/api/package.json
hugomrdias: doesn't work because dev stay running we kill it manually, so i thought `clean` is more explicit

packages/api/db/docker/docker-compose.yml
hugomrdias: i can move the full docker setup to a different place with a proper name definitely 

we can persist db data by just stopping the project instead of cleaning with `yarn clean`

why can't we debug cluster with this ?

README.md
hugomrdias: no, this is need for miniflare the global var needs to be set even if its a fake value.

packages/api/db/docker/docker-compose.yml
vasco-santos: > why can't we debug cluster with this ?

Because we need to run everything in the compose. If we want to debug cluster locally, we need to run the DB docker independently, and just run our cluster locally.

packages/api/package.json
vasco-santos: but, once we force exit miniflare (CTRL+C), we need  to run clean, right?

hugomrdias: or let it run in the background depends on what you are doing 

maybe we can also add a stop, the cli already has it `./scripts/cli.js --stop`

packages/api/db/docker/docker-compose.yml
hugomrdias: not following sorry.. if you are debug something you can just comment out the things you dont need

JeffLowe:
@hugomrdias is this one completion of #472

hugomrdias:
> @hugomrdias is this one completion of #472

Some of that yes but not all

JeffLowe:
@hugomrdias I was checking a different PR and noticed login is broken on the CF preview links.  I checked this preview link, and it has the same breakage (See console logs in screenshot):  
Screen Shot 2021-12-22 at 11 21 30 AM
Could the cloudflare work have impact4ed the auth?

hugomrdias:
> @hugomrdias I was checking a different PR and noticed login is broken on the CF preview links. I checked this preview link, and it has the same breakage (See console logs in screenshot). Screen Shot 2021-12-22 at 11 21 30 AM Could the cloudflare work have impact4ed the auth? Steps to replicate: Open preview link in this pr. Click Login. Attempt to login with either magic link email or GitHub integration. Result: Screen returns user to homepage in unauthenticated state. Console log shows above errors

yes its probably because of the dev deploy override, those previews only worked because of a random db i had setup in supabase. 
that will be properly fixed in a different way with unique sandboxed previews

packages/api/db/docker/docker-compose.yml
vasco-santos: We can do that, but easier to just run/change package json commands according to needs. Anyway, I think this is cleaner outside the db folder/context

vasco-santos:
For some reason, I can't use this branch on my local environment. I get the following error in the worker:

```
Access to fetch at 'http://localhost:9094/add?meta-size=12&local=true&stream-channels=false&raw-leaves=true&cid-version=1' from origin 'http://127.0.0.1:50863' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled
```

It also fails in https://github.com/web3-storage/web3.storage/pull/847 but I don't know exactly why, it works in this repo CI

packages/api/package.json
```diff
@@ -6,11 +6,11 @@
   "type": "module",
   "main": "dist/worker.js",
   "scripts": {
-    "lt": "npm run lt:cluster & npm run lt:postgrest",
-    "lt:cluster": "npx localtunnel --port 9094 --subdomain \"$(whoami)-cluster-api-nft-storage\"",
-    "lt:postgrest": "npx localtunnel --port 3000 --subdomain \"$(whoami)-postgrest-nft-storage\"",
     "deploy": "wrangler publish --env production",
-    "dev": "wrangler dev",
+    "predev": "./scripts/cli.js db --start && ./scripts/cli.js db-sql --cargo --testing --reset",
+    "dev": "miniflare --watch --debug --env ../../.env",
```
alanshaw, line 11: does the "pre" prefix have special meaning in npm? Why not:

```suggestion
    "dev": "run-s dev:**",
    "dev:setup:db": "./scripts/cli.js db --start",
    "dev:setup:sql": "./scripts/cli.js db-sql --cargo --testing --reset",
    "dev:miniflare": "miniflare --watch --debug --env ../../.env",
```

packages/api/src/constants.js
```diff
@@ -32,10 +32,7 @@ switch (CLUSTER_SERVICE) {
 
 export const cluster = {
   apiUrl: clusterUrl,
-  basicAuthToken:
-    typeof CLUSTER_BASIC_AUTH_TOKEN !== 'undefined'
-      ? CLUSTER_BASIC_AUTH_TOKEN
-      : '',
+  basicAuthToken: CLUSTER_BASIC_AUTH_TOKEN,
```
alanshaw, line 35: Makes this global mandatory...but there is no basic auth by default...

alanshaw (approved):

packages/api/package.json
hugomrdias: yes, it runs before `yarn dev`

packages/api/src/constants.js
hugomrdias: there is now with this setup, we already had issues in the past for not testing with auth. I think gozala even added auth to the cluster client tests

.github/workflows/api.yml
```diff
@@ -63,6 +59,7 @@ jobs:
         uses: cloudflare/wrangler-action@1.3.0
         env:
           SENTRY_TOKEN: ${{secrets.SENTRY_TOKEN}}
+          DATABASE_URL: 'https://nft-storage-pgrest-staging.herokuapp.com'
```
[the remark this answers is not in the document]
hugomrdias: this should go away we dont need this, the wrangler file handles this

vasco-santos (approved):

Issue #319, opened 2021-08-17
Guardrails for using Pinning Service API


dchoi27:
There needs to be guardrails around using the Pinning Service API. This epic will track what we want to implement / discussion for the holistic solution.

- [x] https://github.com/nftstorage/nft.storage/issues/148

Issue #148, opened 2021-05-24
Allowlist for access to pinning service API


alanshaw:
We're getting a significant amount of abuse of people using the pinning service API. Large amounts of non-NFT data. We should maintain an allowlist and have users apply to use this API.

refs https://github.com/ipfs-shipyard/nft.storage/issues/70
refs https://github.com/ipfs-shipyard/nft.storage/issues/55

PR #1013, opened 2021-12-28
feat: psa allowlist

alanshaw:
Here's a MVP for an allowlist for the pinning service API. Set a global var called `PSA_ALLOW` with user IDs (CSV) of users allowed to use the API. If the variable is not set we allow everyone.

resolves https://github.com/nftstorage/nft.storage/issues/319
resolves https://github.com/nftstorage/nft.storage/issues/148

### DIFF
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -124,6 +124,7 @@ wrangler secret put CLUSTER_BASIC_AUTH_TOKEN --env production # Get from nft.sto
 wrangler secret put CLUSTER_SERVICE --env production # Which cluster should be used. Options 'IpfsCluster' or 'IpfsCluster2'
 wrangler secret put MAILCHIMP_API_KEY --env production # Get from mailchimp
 wrangler secret put LOGTAIL_TOKEN --env production # Get from Logtail
+wrangler secret put PSA_ALLOW --env production # Get from 1password vault
 
 wrangler publish --env production
 ```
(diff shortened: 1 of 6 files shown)


vasco-santos (changes requested):
This looks good overall as an initial solution.

Even though the allowlist will not change all the time, we should have it in the DB and "editable" via the Admin dashboard. While anyone with the permissions could change the secret in wrangler, it seems error prone to me having this to be replaced there.

Can you please add the secret to https://github.com/nftstorage/nft.storage/tree/main/packages/api#production-setup-envproduction ?

vasco-santos:
I am also not sure about whether we should have the allowlist per user or per auth token. A user could have multiple applications, each one under a different auth token, and we might want more granular permissions. Any particular reason you opted for the user instead?

alanshaw:
> I am also not sure about whether we should have the allowlist per user or per auth token. A user could have multiple applications, each one under a different auth token, and we might want more granular permissions. Any particular reason you opted for the user instead?

We discussed this in the colo and recognised that allowing by auth key would create an admin burden on ourselves when users change/rotate keys as we'd also have to remove old keys and add new ones. We'd like to minimise that admin by allowing access on a user level instead, so that new keys created by the same user automatically have access. We'd also like to not discourage good security practices of rotating keys - folks will be less inclined to rotate keys if they have to talk to us first.

> Even though the allowlist will not change all the time, we should have it in the DB and "editable" via the Admin dashboard. While anyone with the permissions could change the secret in wrangler, it seems error prone to me having this to be replaced there.

Agree. This is the stop gap solution. I'll open an issue to get this added to admin.storage.

vasco-santos (approved):
LGTM :ship:

PR #1017, opened 2021-12-30
chore:move docs to standalone repository for cloudflare pages

dashcraft:
In order for docs to be deployed to the subdomain, cloudflare pages requires it to have its
own repository per deployed project. It doesn't support monorepo's at the moment.


JeffLowe:
Currently blocked until we get the new repository setup to be cloudflare accessible.  @hugomrdias are you the one to help with this?

JeffLowe:
per verified spec this morning, we'll be deploying the docs to /docs subroute thus will not need a separate repo.

Issue #795, opened 2021-11-16
Add mvp admin website for nft.storage 


JeffLowe:
## Overview
As an admin user, I want an admin area on the website, so I can in the future manage users, turn and off features and administer the site. 
We need to develop a unified admin for both web3 and nft.storage.  (But initial implementation needed only for web3) 

## Rough Approach:
* spin up next.js site, 
* confirm UI 
~~* connect auth~~
* house codebase in the nft.storage github organization (per Mikel on 11.19) https://github.com/nftstorage/admin.storage

AC:
- [ ] admins can see admin area
- [ ] admins can block/unblock individual api tokens for nft.storage

Longer term thoughts on authentication and authorization:  users who have access to cloulflare worker we create for this can access the admin website (POSSIBLY using the cloudflare teams feature longterm)

Impacted by https://github.com/nftstorage/nft.storage/issues/826
Mirror task of https://github.com/web3-storage/web3.storage/issues/563

PR #1022, opened 2022-01-03
Feature #795: Schema updates for admin site

jsdevel:
Per @hugomrdias, this diff will need to be manually run against the nft schema after it's merged as there is no migration system in place.

### DIFF
--- a/packages/api/db/tables.sql
+++ b/packages/api/db/tables.sql
@@ -1,3 +1,12 @@
+-- Auth key blocked status type is the type of blocking that has occurred on the api
+-- key.  These are primarily used by the admin app.
+CREATE TYPE auth_key_blocked_status_type AS ENUM (
+    -- The api key is blocked.
+    'Blocked',
+    -- The api key is unblocked.
+    'Unblocked'
+);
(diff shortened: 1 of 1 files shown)


packages/api/db/tables.sql
```diff
@@ -131,6 +150,22 @@ CREATE TABLE IF NOT EXISTS upload
...
+  ak.secret as token,
+  ak.id as token_id,
+  ak.deleted_at as deleted_at,
+  akh.inserted_at as reason_inserted_at,
+  akh.reason as reason,
+  akh.status as status
+from public.user u
+right join auth_key ak on ak.user_id = u.id
+full outer join auth_key_history akh on akh.auth_key_id = ak.id
+where akh.deleted_at is null;
```
jsdevel, line 167: i couldn't find a way to `.or()` across multiple tables in a single postrest-js query, and the postrest docs recomment adding a view for more complicated filtering: https://postgrest.org/en/stable/api.html#operators

hugomrdias (approved):
can you look at the CI please

PR #1032, opened 2022-01-05
fix: remove confusing language about 100MB uploads

### LINKED ISSUE #998: website-pages-faq-filesLimit: update text, issue #220 has been resolved
website issue:
https://nft.storage/faq/#is-there-a-limit-on-numbers-of-files-in-a-directory

updates requested here:
https://github.com/nftstorage/nft.storage/blob/8fc8c4554763b0f921d3dbcacbe4961c9b656aae/packages/website/pages/faq.js#L64-L67
https://github.com/nftstorage/nft.storage/blob/8fc8c4554763b0f921d3dbcacbe4961c9b656aae/packages/website/lib/faqContent.js#L141-L148

resolved issue #220

dchoi27:
Closes #998

### DIFF
--- a/packages/website/lib/faqContent.js
+++ b/packages/website/lib/faqContent.js
@@ -138,48 +138,13 @@ const faqContent = {
   ),
   filesLimit: (
     <p className="lh-copy white mb4">
-      It depends. If you&#39;re using <InlineCode>storeDirectory</InlineCode>{' '}
-      from the JS library (i.e. sending a multipart HTTP request) then the
-      request will be limited by 100MB so you&#39;ll only be able to send as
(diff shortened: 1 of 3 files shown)


alanshaw:
I'm not sure about this, Cloudflare still imposes the 100MB per request limit, and not everyone uses the JS client...

dchoi27:
There's been a number of folks saying there's a 100MB limit to use NFT.Storage, and assuming that's causing them issues when it's almost always something else. I think we can just make the language super clear in the HTTP API docs, which users would have to go to anyway in order to encounter the 100MB limit. Thoughts?

alanshaw (approved):

Issue #1040, opened 2022-01-06
Adjust blog post page hero image


JeffLowe:
Reported on https://nft.storage/blog/post/2022-01-05-mikeal-modern-finance/

PR #1042, opened 2022-01-06
fix: image positioning and use high-res for new blog entry

cmunns:
addresses #1040
will center position the blog images and prevent them from automatically being 100% if the desired width is not added to avoid upscale and crop issues
before:
Screen Shot 2022-01-06 at 12 40 37 PM
after:
fixed-screenshot

NOTE: the new image will accommodate a wide full width banner assuming that the artwork is setup that wide. I'd recommend designing with an art board using 3072 x 800 (double size for retina screens) with 1600px wide centered main area which will serve as the thumbnail on cards.


dashcraft (approved):
Looks good!

PR #1051, opened 2022-01-10
fix: metrics

alanshaw:
* They should not be generated on-the-fly
    * It restricts us to 30s per query (pgrest heroku dyno)
    * Can cause us problems with prom agent reading them (it times out)
    * ...needs to cache or expose to perf issues
* They should be calculated from the replica
    * So they don't impact performance of the production DB

This PR:

* Stores the current value for each metric in the DB
* Adds a cron job for re-calculating metrics every 10 minutes
    * Connects to the replica DB to calculate stats
    * Writes to a new table `metric`
* Updates the `/metrics` endpoint to query the `metric` table for the latest value

### DIFF
--- a/packages/api/scripts/cmds/db-types.js
+++ b/packages/api/scripts/cmds/db-types.js
@@ -14,6 +14,7 @@ export async function dbTypesCmd() {
 
   try {
     await dbSqlCmd({ cargo: true, testing: true })
+    await delay(2000)
     const url = `${process.env.DATABASE_URL}/?apikey=${process.env.DATABASE_TOKEN}`
     await execa(
       'openapi-typescript',
--- a/packages/api/src/routes/metrics.js
+++ b/packages/api/src/routes/metrics.js
@@ -1,6 +1,4 @@
-const UPLOAD_TYPES = ['Car', 'Blob', 'Multipart', 'Remote', 'Nft']
-const PIN_SERVICES = ['IpfsCluster2', 'IpfsCluster', 'Pinata']
-const PIN_STATUSES = ['PinQueued', 'Pinning', 'Pinned', 'PinError']
+import { UPLOAD_TYPES, PIN_SERVICES, PIN_STATUSES } from '../utils/db-client.js'
 
 /**
  * TODO: basic auth
--- a/packages/api/src/routes/metrics.js
+++ b/packages/api/src/routes/metrics.js
@@ -1,99 +1,64 @@
-import { secrets, database } from '../constants.js'
-import { DBClient } from '../utils/db-client.js'
-
-const db = new DBClient(database.url, secrets.database)
-
-export async function getUserMetrics() {
-  const query = db.client.from('user')
-  const res = await query
-    .select('*', { head: true, count: 'exact' })
-    .range(0, 1)
-  if (res.error) {
-    throw res.error
-  }
-  return { total: res.count }
-}
-
-export async function getNftMetrics() {
-  const types = ['Car', 'Blob', 'Multipart', 'Remote', 'Nft']
-  const totals = await Promise.all(
-    types.map(async (t) => {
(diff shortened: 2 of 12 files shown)


packages/api/scripts/cmds/db-types.js
```diff
@@ -14,6 +14,7 @@ export async function dbTypesCmd() {
 
   try {
     await dbSqlCmd({ cargo: true, testing: true })
+    await delay(2000)
```
vasco-santos, line 17: Is there a reason for this?

packages/api/src/routes/metrics.js
```diff
@@ -1,99 +1,64 @@
...
-          }
-          return { [status]: res.count }
-        })
-      )
-      return { [service]: Object.assign({}, ...totals) }
-    })
-  )
-  return { totals: Object.assign({}, ...totals) }
-}
+const UPLOAD_TYPES = ['Car', 'Blob', 'Multipart', 'Remote', 'Nft']
```
vasco-santos, line 1: nit: it would be great to have these in a single place. Currently, we have them at list here and in the cron code. This might be error prone in the case of a new PinService, or type of upload as we can forget to update the other side.

packages/cron/src/jobs/metrics.js
```diff
@@ -0,0 +1,89 @@
...
+ON CONFLICT (name) DO UPDATE
+        SET value = $2, updated_at = TIMEZONE('utc', NOW())
+`
+
+const UPLOAD_TYPES = ['Car', 'Blob', 'Multipart', 'Remote', 'Nft']
+const PIN_SERVICES = ['IpfsCluster2', 'IpfsCluster', 'Pinata']
+const PIN_STATUSES = ['PinQueued', 'Pinning', 'Pinned', 'PinError']
+
+/**
+ * Calculate metrics and update their current values in the DB.
```
vasco-santos, line 27:
```suggestion
 * Calculate metrics from RO DB and update their current values in the RW DB.
```

packages/cron/src/jobs/metrics.js
```diff
@@ -0,0 +1,89 @@
...
+    if (promise.isRejected) {
+      error = error || promise.reason
+      console.error(promise.reason)
+      continue
+    }
+    const metric = promise.value
+    await pg.query(UPDATE_METRIC, [metric.name, metric.value])
+  }
+
+  if (error) throw error
```
vasco-santos, line 51: Good call that we update what we can and only error in the end. We still need to create an issue to integrate these errors with an alert system.

vasco-santos (approved):
LGTM overall

Minor details commented inline

packages/api/scripts/cmds/db-types.js
alanshaw: Yes I was getting a socket hangup error. I think because the new postgrest version restarts itself when changes are made. It just needs a second or two to come back.

packages/cron/src/jobs/metrics.js
alanshaw: I saw this in the pins job https://github.com/nftstorage/nft.storage/blob/20ee0b6241917c92821e4d7b79e19e4ae36ccea1/.github/workflows/cron-pins.yml#L35-L37

Is that what we want @hugomrdias?

packages/api/src/routes/metrics.js
alanshaw: We already do this in the cron package:

```js
import { DBClient } from '../../../api/src/utils/db-client.js'
```

So I think should be ok to import them...

packages/cron/src/jobs/metrics.js
vasco-santos: Oh yes, I think that's it. Not sure if we need any config in the expected opsgenie heartbeat timings

PR #961, opened 2021-12-13
feat: gateway race with multiple public gateways + durable object metrics

vasco-santos:
This PR adds a race between multiple public IPFS gateway responsing with the first respond arriving. A metrics endpoint with prometheus format was also wired up to keep state about the gateway behaviour.

Two Durable objects were added. One keeps all the stats for the gateways in terms of number of req, fastest count, error count, response time average and a histogram (per initial metrics, I think we need to rethink the histogram values). The other keeps a record of the gateways that previously were able to get a given CID.

Needs:
- Reset migrations (likely needs a new worker name)

<details>
  <summary> Metrics response can also be seen in the staging endpoint</summary>
# HELP nftstorage_gateway_total_response_time Average response time.
# TYPE nftstorage_gateway_total_response_time gauge
nftstorage_gateway_total_response_time{gateway="https://ipfs.io",env="staging"} 21958
nftstorage_gateway_total_response_time{gateway="https://cf-ipfs.com",env="staging"} 6967
nftstorage_gateway_total_response_time{gateway="https://nft-storage.mypinata.cloud/",env="staging"} 12885
# HELP nftstorage_gateway_total_fastest_response_time Total requests performed.
# TYPE nftstorage_gateway_total_fastest_response_time counter
nftstorage_gateway_total_fastest_response_time{env="staging"} 2707
# HELP nftstorage_gateway_total_requests Total requests performed.
# TYPE nftstorage_gateway_total_requests counter
nftstorage_gateway_total_requests{gateway="https://ipfs.io",env="staging"} 25
nftstorage_gateway_total_requests{gateway="https://cf-ipfs.com",env="staging"} 25
nftstorage_gateway_total_requests{gateway="https://nft-storage.mypinata.cloud/",env="staging"} 25
# HELP nftstorage_gateway_total_successful_requests Total successful requests.
# TYPE nftstorage_gateway_total_successful_requests counter
nftstorage_gateway_total_successful_requests{gateway="https://ipfs.io",env="staging"} 18
nftstorage_gateway_total_successful_requests{gateway="https://cf-ipfs.com",env="staging"} 18
nftstorage_gateway_total_successful_requests{gateway="https://nft-storage.mypinata.cloud/",env="staging"} 18
# HELP nftstorage_gateway_total_failed_requests Total failed requests.
# TYPE nftstorage_gateway_total_failed_requests counter
nftstorage_gateway_total_failed_requests{gateway="https://ipfs.io",env="staging"} 7
[...]

PR #1064, opened 2022-01-12
feat: api ipfs path

### LINKED ISSUE #1049: Add /ipfs route to API and proxy it to gateway
As part of https://github.com/nftstorage/nft.storage/issues/823

vasco-santos:
Add ipfs route to proxy CIDv0 to the gateway

Closes #1049 

Needs:

- [x] #961 
- [x] Production release

### DIFF
--- a/packages/api/src/bindings.d.ts
+++ b/packages/api/src/bindings.d.ts
@@ -14,6 +14,7 @@ declare global {
   const MAGIC_SECRET_KEY: string
   const DATABASE_URL: string
   const DATABASE_TOKEN: string
+  const GATEWAY_NFT_STORAGE: string
   const MAILCHIMP_API_KEY: string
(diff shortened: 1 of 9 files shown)


alanshaw (changes requested):
Can we redirect instead of proxy? HTTP 302 response?

vasco-santos:
> Can we redirect instead of proxy? HTTP 302 response?

Yes :) I want to use redirect, but did not test yet

alanshaw (changes requested):

vasco-santos:
@alanshaw forgot to mention, I am not a fan of using real production services for testing. The issue here is using subdomains, which make it difficult to spin smoke server or even run the gateway in Miniflare to use it as a testing backend...  basically DNS fails to resolve

PR #1085, opened 2022-01-15
feat: making changes to the copy options per new designs

dashcraft:
Updating to use 2 copy buttons, one for cid and the other for IPFS URL with option for redirect.
This will be consistent with new large file uploader designs.

### DIFF
--- a/packages/website/components/copyButton.js
+++ b/packages/website/components/copyButton.js
@@ -45,7 +45,7 @@ function CopyButton({ text, title, popupContent, children, ...props }) {
       <button
         onClick={handleCopy}
         title={title}
-        className="icon-button transparent v-mid ml1"
+        className="icon-button transparent v-mid mr2"
         {...props}
       >
         <span
--- a/packages/website/components/copyButton.js
+++ b/packages/website/components/copyButton.js
@@ -40,24 +40,26 @@ function CopyButton({ text, title, popupContent, children, ...props }) {
   }
 
   return (
-    <button
-      onClick={handleCopy}
-      title={title}
-      className="icon-button transparent mr1"
-      {...props}
-    >
-      <span
-        data-content={popupContent}
-        className={clsx('popup', { active: popup })}
-      />
-      {popup && (
-        <span className="sr-only" aria-live="assertive">
-          {popupContent}
-        </span>
-      )}
-      <Copy tab-index={-1} />
+    <>
+      <button
+        onClick={handleCopy}
+        title={title}
+        className="icon-button transparent mr1"
+        {...props}
+      >
+        <span
+          data-content={popupContent}
+          className={clsx('popup', { active: popup })}
+        />
+        {popup && (
+          <span className="sr-only" aria-live="assertive">
+            {popupContent}
+          </span>
+        )}
+        <Copy tab-index={-1} />
+      </button>
       {children}
-    </button>
+    </>
   )
 }
 
--- a/packages/website/components/copyButton.js
+++ b/packages/website/components/copyButton.js
@@ -45,7 +45,7 @@ function CopyButton({ text, title, popupContent, children, ...props }) {
       <button
         onClick={handleCopy}
         title={title}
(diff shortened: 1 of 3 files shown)


dashcraft:
image

dashcraft:
@dchoi27 this is using the same design as the new large file uploader

dashcraft:
@dchoi27 the current design for the large file uploader doesn't show the CID number, would you like me to keep it in the UI?
Maybe right after the copy buttons or right underneath?

dchoi27:
Yes - we should keep the CID in the table. Think maybe just having the CID in the first row, and the copy buttons in the second row would do the trick.

dashcraft:
@dchoi27 
image

alanshaw:
I'm not sold on the placement of these controls and the wasted whitespace it'll create. Perhaps we could take some inspiration from the IPFS webui?:

Screenshot 2022-01-19 at 10 50 10

Clicking on the share link icon pops up this dialog:

Screenshot 2022-01-19 at 10 54 07

Proposal:

1. Two icons next to each CID - the current copy icon and a share link icon
    * Personally I feel these should be displayed _after_ the CID
2. The copy icon copies the CID as expected
3. The share icon opens a modal which displays text input boxes with the following content:
    * Gateway URL (http://ipfs.io/ipfs/bafy...)
    * IPFS URL (ipfs://bafy...)
    * Both of these text inputs has a copy icon next to them which copies the URL

In the future we could have a way of switching between gateways, or maybe we just add another (gateway.nft.storage for example). We could also put the plain CID in the modal but that's maybe a little redundant.

JeffLowe:
I recommend we refine what is needed surrounding CID & gateway/ipfs url display/copy/placement for this /files view in our LFU electron app designs (link to this in your slack).  It will be much more efficient to visually lay it out and discuss opposed to threaded convo here (and unblocks so we can get features out to users quicker).  

For this PR to get the core copy and display functionality out for users immediately, I propose we:
* Keep the CID Displayed as David suggests
* Move copy icons to after data as Alan suggests
* Only include the gateway URL that is included with ability to copy (more robust sharing features can be defined in the LFU mock). 

This reflects what's proposed for this PR:  
Screen Shot 2022-01-19 at 8 36 02 AM

@alanshaw I'll also be updating #1110 with your comments here as well as inserting comments into the LFU design which will inform/guide this /files view in future iterations.

dchoi27:
Yeah that works for me, except we do not want to include a gateway URL copy link at the moment - friction needs to be kind of high to put gateway links in NFTs. It should be the `ipfs://` link.

JeffLowe:
Understood.  We will use whatever url is currently returned (eg https://bafkreigqkpn5uiuulz3hz4r6obk4wi5y5h3yoamjtafgb5putjmowlcu4i.ipfs.dweb.link/ ) for the URL copy value. I adjusted above comment to reflect this.  
We'll proceed with this for the immediate term, unless we hear otherwise.

JeffLowe:
Correction on the URL.  Instead of any gateway URL, we will display 
URI (copy icon) 
...which will contain the ipfs uri for the cid
eg:  ipfs://bafkreigqkpn5uiuulz3hz4r6obk4wi5y5h3yoamjtafgb5putjmowlcu4i

dashcraft:
Current url from copy/URL link:
ipfs://bafybeic3a6txgjsavv57lvfud67y7zvz2eq2acp3wzbk5tey4xy72zvuyu

current screenshot:
image

dchoi27:
Cool! Only nit from me - think the copy button should be closer to the CID, and then more space between the first copy button and where `URL:` is. I.e. right now it goes `<CID><big space><copy button><small space><URL:><big space><copy button>` but think it'd look better as  `<CID><small space><copy button><big space><URL:><small space><copy button>`

cmunns (approved):
@dchoi27 updated spacing per request and I agree on the need for it. All else looks good, thanks @dashcraft.

PR #1095, opened 2022-01-17
fix: catch more errors

### LINKED ISSUE #986: Invalid tokens should not create p1 alerts in opsgenie.
Magic Admin SDK Error: [ERROR_MALFORMED_TOKEN] might not be the right error to be tracking to generate alerts. If it is, we should verify that the volume of these requests have reached some threshold before error (the alert should specify what this threshold is - probably shouldn't be P1 in either case).

hugomrdias:
This shoud catch more errors and avoid sending them to sentry.

Includes sentry, maintenance and db errors.

closes #986

### DIFF
--- a/packages/api/src/utils/auth.js
+++ b/packages/api/src/utils/auth.js
@@ -13,55 +13,47 @@ export const magic = new Magic(secrets.magic)
 export async function validate(event, { log, db }) {
   const auth = event.request.headers.get('Authorization') || ''
   const token = magic.utils.parseAuthorizationHeader(auth)
-  try {
-    // validate access tokens
-    if (await verifyJWT(token, secrets.salt)) {
-      const decoded = parseJWT(token)
 
-      const user = await db.getUser(decoded.sub)
-      if (user.error) {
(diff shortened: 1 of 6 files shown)


packages/api/src/utils/auth.js
```diff
@@ -13,55 +13,47 @@ export const magic = new Magic(secrets.magic)
...
+    const user = await db.getUser(claim.iss)
+    if (user) {
       log.setUser({
-        id: user.data.id,
+        id: user.id,
       })
 
       return {
-        user: user.data,
+        user: user,
```
alanshaw, line 51:
```suggestion
        user,
```

alanshaw (approved):
Code changes look good, but this **doesn't** close https://github.com/nftstorage/nft.storage/issues/986 correct?

hugomrdias:
> Code changes look good, but this **doesn't** close #986 correct?

this closes 986 because now we dont catch magic errors in auth.js so they bubble up to errors.js and are catched there

PR #1099, opened 2022-01-17
fix: query cluster 3 for status

alanshaw:
Content from Cluster 1 has been moved to Cluster 3.

When cluster 2 has finished transferring we'll migrate the database, but this will get the pin cron job moving until then.


alanshaw:
I'm going to merge this with admin powers - the pins statuses are getting more and more out of date and people will start raising issues.

PR #1104, opened 2022-01-18
perf: concurrent pin status sync

alanshaw:
This PR updates the pin status sync cron job to add concurrency and speed up processing.


vasco-santos (approved):
nice LGTM

PR #1106, opened 2022-01-18
chore: improve gateway tests mocking

### LINKED ISSUE #1091: Gateway follow up improvements
Per review in https://github.com/nftstorage/nft.storage/pull/961 and sync discussion, let's do a few changes/improvements:

- [ ] Use a real IPFS node for tests instead of smoke tests
- [x] Having a separate metrics instance per gateway so that we can update each one as soon as it finishes rather than wait for all of them to finish. It also simplifies the logic in that class significantly.

vasco-santos:
This PR runs a docker container with a go-ipfs node to act as gateway. In the startup, it will add some files to the node for being able to fetch the CIDs in the tests

Closes #1091

### DIFF
--- a/packages/gateway/scripts/cli.js
+++ b/packages/gateway/scripts/cli.js
@@ -1,12 +1,19 @@
 #!/usr/bin/env node
+import net from 'net'
 import fs from 'fs'
 import path from 'path'
+import { execa } from 'execa'
 import { fileURLToPath } from 'url'
 import sade from 'sade'
 import { build } from 'esbuild'
 import git from 'git-rev-sync'
 import Sentry from '@sentry/cli'
+import delay from 'delay'
+import pWaitFor from 'p-wait-for'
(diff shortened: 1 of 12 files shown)


packages/gateway/scripts/cli.js
```diff
@@ -18,52 +25,178 @@ prog
...
+    console.log('docker started')
+
+    // Add fixture files
+    const client = create({ url: IPFS_API_URL })
+
+    await Promise.all([
+      // bafkreidyeivj7adnnac6ljvzj2e3rd5xdw3revw4da7mx2ckrstapoupoq
+      client.add('Hello nft.storage! 😎', {
+        chunker: 'size-1048576',
+        rawLeaves: true,
```
alanshaw, line 116: We don't override these settings in NFT.Storage - I think this might be confusing.

alanshaw (approved):
LGTM, just one small suggestion.

PR #1112, opened 2022-01-18
fix: refactor metrics durable object

vasco-santos:
As part of https://github.com/nftstorage/nft.storage/issues/1091 , this PR handles:

> Having a separate metrics instance per gateway

Previous metrics durable object was responsible for tracking the metrics of all the gateways, as well as generic metrics (total sum of winner response times). As a result, this PR creates two new Durable Objects. A GatewayMetrics DO, which will have an instance per gateway and a GenericMetrics DO, which will keep generic metrics agnostic to specific gateways.

An implementation detail worth mentioning is when metrics per gateway are updated. The winner gateway metrics is updated right away (after the winner request finishes), while all the other gateways will be updated in parallel once the remaining requests end. We could store every single request as soon as it finishes, but it would add complexity to the codebase for having the metrics a bit later updated. The extra complexity is likely not worth for having the metrics sooner updated. I am still considering wether adding the winner first is worth it, but given we need to update the generic metrics I added it earlier on.


packages/gateway/src/durable-objects/gateway-metrics.js
```diff
@@ -0,0 +1,97 @@
...
+ * @property {number} totalWinnerRequests number of performed requests where winner
+ * @property {Record<string, number>} responseTimeHistogram
+ *
+ * @typedef {Object} ResponseStats
+ * @property {boolean} ok request was successful
+ * @property {number} [responseTime] number of milliseconds to get response
+ * @property {boolean} [winner]
+ */
+
+const GATEWAT_METRICS_ID = 'gateway_metrics'
```
alanshaw, line 15:
```suggestion
const GATEWAY_METRICS_ID = 'gateway_metrics'
```

alanshaw (approved):

PR #1113, opened 2022-01-19
perf: add indexes for pins and metrics cron improvements

vasco-santos:
Crons for Pins and Metrics perform queries filtering out by columns of Pin and Upload table. We need to add indexes to these columns to boost performance.

TODO:
- [x] Add indexes to production once PR approved


alanshaw (approved):

vasco-santos:
Added to prod and staging DBs

PR #1120, opened 2022-01-20
Fix: Make the service fail early if environment variables are not defined.

redaphid:
Would anyone mind if we made the service explode if it doesn't have the environment variables it needs to run? 
This would make it easier to debug local dev, and give us some safeguard when deploying.

### DIFF
--- a/packages/api/src/constants.js
+++ b/packages/api/src/constants.js
@@ -1,4 +1,36 @@
 // let MAGIC_SECRET_KEY, SALT, PINATA_JWT, SENTRY_DSN, DATABASE_TOKEN, CLUSTER_SERVICE, LOGTAIL_TOKEN, MAILCHIMP_API_KEY, METAPLEX_AUTH_TOKEN
+import { config } from 'dotenv'
+
+config({ debug: true })
+
+const {
+  MAGIC_SECRET_KEY,
+  SALT,
+  SENTRY_DSN,
+  DATABASE_TOKEN,
+  CLUSTER_SERVICE,
+  LOGTAIL_TOKEN,
+  MAILCHIMP_API_KEY,
+  CLUSTER_API_URL,
+  DATABASE_URL,
+  DEBUG,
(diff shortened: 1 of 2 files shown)


packages/api/src/constants.js
```diff
@@ -1,4 +1,36 @@
...
+  SALT,
+  SENTRY_DSN,
+  DATABASE_TOKEN,
+  CLUSTER_SERVICE,
+  LOGTAIL_TOKEN,
+  MAILCHIMP_API_KEY,
+  CLUSTER_API_URL,
+  DATABASE_URL,
+  DEBUG,
+} = process.env
```
alanshaw, line 17: This is not a node application, it's a service worker. There's no process object and no access to the filesystem so this won't work.

packages/api/src/constants.js
```diff
@@ -1,4 +1,36 @@
...
+
+if (!SALT) {
```
alanshaw, line 19: Accessing any of these globals when they are undefined will throw an error so we already fail fast.

This is also why you see `typeof CLUSTER_BASIC_AUTH_TOKEN !== 'undefined'` in this file for example, because this var is optional.

alanshaw (changes requested):

packages/api/src/constants.js
redaphid: Alright, no worries. Thanks for taking the time to review it.

packages/api/src/constants.js
redaphid: Is this new(ish)? I ended up doing this because I was missing env vars, and had to dig around trying to find what was missing.

That said, we'd have to go a different route if we're not using process.env. 
I'll see if I hit it again, and open a new PR if that happens.

redaphid:
tl;dr it's not a node app. Thanks for taking the time to review. I'll document where the undefined vars were messing me up if I run into them again.

Issue #782, opened 2021-11-14
Postgres improvements


dchoi27:
We have Eric from Trigram working on our current postgres setup including schema, config, migrations. 

Tasks: 

- [x] Repo setup
- [ ] Baseline metrics (heroku dashboard)
- [ ] Baseline benchmarks with local setup mirroring heroku
  - [ ] `POST /upload`, `POST /pins`, `GET /check` 
- [ ] Test out table partition to enable better scaling
- [x] Analise postgres config
- [ ] Analise current postgres indexes
- [ ] Cloudflare sync for the new db connectors
- [ ] Setup https://pganalyze.com/  for our dbs
- [x] Add migration system 
- [ ] Split out read to read replica

hugomrdias:
@yocontra can you please review the list above and give some feedback if it makes sense ?

feel free to add remove etc

yocontra:
Here's my current TODO list for my first PR of DB changes:
- [x] Basic analysis of production index usage, heroku stats to figure out where out issues are
- [x] Create partition schemas for pin, upload, and content tables
  - Ended up going with hash partitioning into 100 tables by CID - this should speed up reads/writes by isolating IO into smaller tables with ~20K rows. You will be able to grow these tables up to 2M rows before you need to revisit and consider more partitions, so this should be acceptable for a long while.
- [] Optimize the create_upload function
- [] Optimize basic DB configuration flags (specifically when autovacuum should run, table statistics, etc. to keep indexes cleaner than they are now since this is write-heavy workload)
- [] Create benchmark system in-repo and create a report comparing before vs. after the changes
- [] Write migrations for production/staging to move data from the existing single tables into the new partitioned model


TODO for future PRs:
- [] Postgres index analysis + make changes based on unused indexes
- [] Move all read queries to the read replica
- [] Use a connection pooler like pgbouncer or https://github.com/yandex/odyssey
- [] Set up pganalyze (or pgbadger) to get better statistics on the production DB
- [] Set up a migration system for production and staging DB changes (sqitch suggested by @hugomrdias)
- [] New cloudflare worker DB connection work if needed (#829)

yocontra:
[...]

PR #1126, opened 2022-01-20
Postgres Partitioning, Partial of #782

yocontra:
Included in this PR:
- Partition content, pin, and upload tables by cid into 100 partitions each
  - This should radically speed up reads, writes, vacuums, and analyze.
  - Luckily the way hashing of the CID text works, this provides an extremely even distribution across partitions.
  - This should future-proof the DB up to about 1B records where we will need to add additional partitions - at this point we will probably want to look into sharding the DB.
  - This code is very ugly - better syntax for this is supposed to land in PG 15, but for now we have to be very verbose. See notes at the top of the partition.sql file for more info.
- Improves the docker-compose for the database by specifying CPU and RAM constraints
- Updates the postgres dockerfile to use the latest PG (13) to match what we run in production/staging
  - Also improves it by adding additional logging, which is useful when working with the DB on local dev
- Improve the DB configuration
  - Many other settings aren't able to be configured like this (Heroku doesn't expose most flags), noted in #782 so I did what I could within the constraints of what we can do inside the DB

TODO:
- Add benchmark system to current main branch + new main branch, collect measurements and compare
- Write migration for staging/production - this is going to be a pretty large migration since we will need to move all of the current records into temp tables, create the new tables/partitions, then move everything into the new tables.


hugomrdias (approved):
Looks perfect!! thank you 
What are the next steps to run the benchmark ?

yocontra:
@hugomrdias I wanted a thumbs up on the general approach before I proceeded doing the benchmarks and migration which are going to be the most time consuming part - now that I got your eyes on it I'll rebase + finish up by implementing those and re-apply for your approval when done. Will post results of benchmark before + after these changes in this thread as well.

packages/api/benchmark/cli.js
```diff
@@ -0,0 +1,22 @@
...
+  path: path.join(__dirname, '..', '..', '..', '.env'),
+})
+
+// set up env
+console.log('Spinning up environment...')
+await dbCmd({ init: true, start: true })
+try {
+  await dbSqlCmd({ cargo: true, testing: true })
+} catch (err) {
+  console.error(err)
```
alanshaw, line 19: You can use a `finally` block to "spin down" instead of explicitly logging the error and it more accurately describes the intention of the code.

packages/api/db/functions.sql
```diff
@@ -63,6 +63,7 @@ BEGIN
             (data ->> 'files')::jsonb,
             (data ->> 'origins')::jsonb,
             (data ->> 'meta')::jsonb,
+            json_arr_to_text_arr(data -> 'backup_urls'),
             (data ->> 'updated_at')::timestamptz,
             (data ->> 'inserted_at')::timestamptz)
     ON CONFLICT ( user_id, source_cid )
```
alanshaw, line 69: Need to append the new backup URLs `ON CONFLICT`.

packages/api/db/reset.sql
```diff
@@ -15,4 +15,6 @@ DROP TABLE IF EXISTS public.user CASCADE;
 DROP TABLE IF EXISTS cargo.aggregate_entries;
 DROP TABLE IF EXISTS cargo.aggregates;
 DROP TABLE IF EXISTS cargo.deals;
-DROP SERVER IF EXISTS dag_cargo_server CASCADE;
\ No newline at end of file
+DROP SERVER IF EXISTS dag_cargo_server CASCADE;
+
+ALTER DATABASE postgres RESET ALL;
```
alanshaw, line 20: Was this meant to be included? If so, can we please get a comment to explain why?

alanshaw (changes requested):
We need a migration script also. Can you please fork the DB and time it so we know how long to schedule the maintenance period for?

(I'm guessing migration from backup table to `TEXT[]` can probably be done separately without maintenace period)

packages/api/db/functions.sql
hugomrdias: sorry didnt get this one ? it needs backup_urls in the on conflict clause ?

hugomrdias (approved):

yocontra:
@alanshaw Yep that's the next step (after finishing some basic benchmarks) - the size of the DB is not massive so it shouldn't be a terribly long migration, I've done a 1B+ row partition migration before and it took a few hours so I would anticipate this might take a few minutes given that we're operating against a few 12M row tables. We may want to temporarily upsize the instance if we want to speed things up, and I'll need to check that we're below 50% storage utilization before we run the migration since this will need to temporarily duplicate those tables.

After I make the migration script I'd like to make a clone of production and test against it locally, then I'd like to clone the production DB onto staging and deploy this code + run the migration on staging and run it for a few days and make sure all is well.

There is one outlier here that needs to be resolved before any deployment/migrations - the version of postgrest running on production is detached from the codebase, it was deployed manually via the heroku UI and is out of date so it doesn't support partitioned tables - it needs to be brought up to the latest version. I don't know who the owner of that task should be.

packages/api/db/functions.sql
yocontra: Will do

yocontra: Yeah also had a Q about this - the on conflict is handling a case where a user is uploading an identical file for a second time, wouldn't the backup url also stay the same?

yocontra:
Also just addressing migrations for the backups table: If desired I can submit the backup changes as a separate PR (and separate migration) so we can roll that out, then deal with the partition stuff separately instead of bunching them into one big thing. Let me know if you want me to do that.

hugomrdias:
> Also just addressing migrations for the backups table: If desired I can submit the backup changes as a separate PR (and separate migration) so we can roll that out, then deal with the partition stuff separately instead of bunching them into one big thing. Let me know if you want me to do that.

yes please

hugomrdias:
ping me for heroku upgrade

PR #1128, opened 2022-01-21
feat/more generic token check

redaphid:


dchoi27:
@redaphid can we close? cc @JeffLowe

PR #1133, opened 2022-01-21
fix: properly track errors on winner gateway fetch

vasco-santos:
We were not previously detecting in the winner request response if fetch response was successfully. This typically did not represent a problem because usually the errors would be timeouts. However, after trying a load test scenario in my local machine I noticed that `cf-ipfs.com` has rate limited enabled (don't know at this point what are the conditions of it, can't find documentation on this...).

So, when we hit the CF rate limit, every request was immediately responded with the Error, instead of waiting for further responses.

In this PR, I added a loop cycle where we wait until one request succeeds, properly tracking potential errors. A `GatewayResponsePromise` was created to track of the state of each promise and rely on `p-any` to wait for further results in a simple fashion.

As a follow up PR, I need to track potential rate limit errors and keep a state of that, in order to not try further requests to that gateway until it is back...  We need to discuss alternatives here, currently considering a new durable object but we would need a read before the gateway race.

### DIFF
--- a/packages/gateway/src/gateway.js
+++ b/packages/gateway/src/gateway.js
@@ -1,7 +1,6 @@
 /* eslint-env serviceworker, browser */
 
 import pAny from 'p-any'
-import pMap from 'p-map'
 import pSettle from 'p-settle'
 
 import { getCidFromSubdomainUrl } from './utils/cid.js'
@@ -47,10 +46,9 @@ export async function gatewayGet(request, env, ctx) {
 
       await Promise.all([
         // Filter out winner and update remaining gateway metrics
-        pMap(
-          responses.filter((r) => r.value?.url !== winnerGwResponse.url),
-          (r) => updateGatewayMetrics(request, env, r.value, false)
-        ),
+        ...responses
+          .filter((r) => r.value?.url !== winnerGwResponse.url)
+          .map((r) => updateGatewayMetrics(request, env, r.value, false)),
         updateCidsTracker(request, env, successFullResponses, cid),
       ])
     }
@@ -71,8 +69,10 @@ export async function gatewayGet(request, env, ctx) {
       (async () => {
         // Update metrics as all requests failed
         const responses = await pSettle(gatewayReqs)
-        await pMap(responses, (r) =>
-          updateGatewayMetrics(request, env, r.value, false)
+        await Promise.all(
+          responses.map((r) =>
+            updateGatewayMetrics(request, env, r.value, false)
+          )
         )
       })()
     )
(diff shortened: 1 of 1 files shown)


packages/gateway/src/gateway.js
```diff
@@ -25,55 +25,58 @@ export async function gatewayGet(request, env, ctx) {
...
+  )
 
-    let response
-    try {
-      response = await fetch(
-        `${ipfsUrl.toString()}/${cid}${reqUrl.pathname || ''}`,
-        { signal: controller.signal }
+  try {
+    /** @type {GatewayResponse} */
+    let winnerGwResponse = await pAny(gatewayReqs)
```
alanshaw, line 37: Why not use the `options` object of `p-any` to filter out the ones that are not `ok`?

vasco-santos: I tried it at first given it would be so easier, but then I get to the point where I need to track counters for requests failed and I don't know what gateway urls failed because they will throw instead of returning the GatewayResponse object that I will be able to parse

vasco-santos: I was thinking about this yesterday, and using `p-settle` after should be enough to track the fail counters. Need to try it out tomorrow

packages/gateway/src/gateway.js
```diff
@@ -25,55 +25,41 @@ export async function gatewayGet(request, env, ctx) {
...
+      )
 
-  try {
-    const winnerGwResponse = await pAny(gatewayReqs)
+      await Promise.all([
+        // Filter out winner and update remaining gateway metrics
+        pMap(
+          responses.filter((r) => r.value?.url !== winnerGwResponse.url),
+          (r) => updateGatewayMetrics(request, env, r.value, false)
+        ),
```
alanshaw, line 53: Could just map and spread....

packages/gateway/src/gateway.js
```diff
@@ -95,6 +81,60 @@ export async function gatewayGet(request, env, ctx) {
...
+ * Store metrics for winner gateway response
+ *
+ * @param {Request} request
+ * @param {import('./env').Env} env
+ * @param {GatewayResponse} winnerGwResponse
+ */
+async function storeWinnerGwResponse(request, env, winnerGwResponse) {
+  await Promise.all([
+    updateGatewayMetrics(request, env, winnerGwResponse, true),
+    updateGenericMetrics(request, env, winnerGwResponse),
```
alanshaw, line 94: "Gateway" and "Generic" - same length and begin with G....I thought this was calling the same function both times at first glance. Do you think we should rename? Not sure what would be appropriate - summary/main/overall/universal metrics?

alanshaw (approved):

packages/gateway/src/gateway.js
vasco-santos: Summary is a better naming, I'll change it

vasco-santos: Probably will also rename the Durable Object. But when we go for production release to reset all version counters

PR #1140, opened 2022-01-21
fix: decentralizing nft.s content fixes

dchoi27:


JeffLowe (approved):
Copy only changes. Reviewd and approved.

PR #1154, opened 2022-01-24
fix: rate limit error message

alanshaw:
Unfortunately Cloudflare does not yet allow us to set a custom response when rate limiting based on HTTP header value. You get back a HTML page and an HTTP 429 status.

This PR inspects the status message and skips parsing the response, throwng an appropriate error instead.

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -140,6 +140,9 @@ class NFTStorage {
               headers,
               body: carFile,
             })
+            if (response.status === 429) {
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):
This looks good, but will need test coverage to make CI happy

hugomrdias (approved):

dchoi27:
Should we merge this?

alanshaw:
Needs tests...

dchoi27:
@vasco-santos

PR #1163, opened 2022-01-25
docs: update comment on meta field in upload table

alanshaw:
The `meta` field is actually now used by both PSA and the `/metaplex/upload` endpoint.


ribasushi (approved):

vasco-santos (approved):
ty

PR #1171, opened 2022-01-26
chore: improve docs for docker

### LINKED ISSUE #472: Improvements to initial setup for developers

hugomrdias:
This PR and the new simple dev setup should be enough to close #472

closes #472

### DIFF
--- a/packages/api/README.md
+++ b/packages/api/README.md
@@ -8,6 +8,8 @@ Make sure you already did step 1 and 2 from these [root instructions](/#getting-
 
 ### Running Locally
 
+Install docker and make sure that you have at least 8GB alocated.
+
 ```bash
 cd packages/api
 yarn dev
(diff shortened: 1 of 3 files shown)


packages/api/README.md
```diff
@@ -8,6 +8,8 @@ Make sure you already did step 1 and 2 from these [root instructions](/#getting-
 
 ### Running Locally
 
+Install docker and make sure that you have at least 8GB alocated.
```
alanshaw, line 11:
```suggestion
Install docker and make sure that you have at least 8GB of RAM allocated.
```

alanshaw (approved):

PR #1172, opened 2022-01-26
refactor: use native postgres client in pins cron instead of postgrest

alanshaw:
We're 4 million+ pins behind on status updates from cluster. Our pins job keeps failing because sometimes the query takes over 30s. This PR switches to using the native postgres client which maintains an active connection and allows queries to take longer than 30s. It should be slightly quicker and hopefully fail less. The next step is to run the read queries against the RO database (the replica) to avoid stressing out the production DB.


vasco-santos (approved):
LGTM

PR #1178, opened 2022-01-27
feat: gateway rate limiting durable object

### LINKED ISSUE #1165: Track and Prevent IPFS public gateways rate limite
Clouflare IPFS gateway has rate limiting characteristics. As a consequence, our CF worker for the gateway will be affected by it easily.

When Cloudflare Worker is rate limited, we will not be able to use the currently most efficient caching gateway for some time, nor warm its cache. This will cause impact on response time, specially when gateway is blocked for a long period.

vasco-santos:
This PR adds a durable object that tracks the gateway requests over time to avoid rate limiting. For this, it simply keeps a state with the last n timestamps of the requests.

Timings on when the actual request reaches the gateway might influence negatively and we end up being rate limited anyway. While we work on further improvements with the gateways, we will track metrics that allows us to track requests being prevented (for now rate limit only) and status of failed requests (to track potential rate limiting errors).  If we see "rate limit errored requests" we might need to consider having a safe buffer. We can also consider doing a random selection of the gateways and use a subset to decrease load

In the unfortunate event of all gateways getting rate limited, a redirect to the first gateway (ipfs.io) is performed

As discussed with @alanshaw metrics changed to a more concise logic to track response types and reasons for preventing requests. Added `totalResponsesByStatus` and `totalRequestsPreventedByReason`

Closes #1165


packages/gateway/src/durable-objects/gateway-rate-limits.js
```diff
@@ -0,0 +1,100 @@
...
+ * Get rate limiting characteristics of a given Gateway.
+ *
+ * @param {string} gatewayUrl
+ * @return {RateLimitCharacteristics}
+ */
+function getRateLimitingCharacteristics(gatewayUrl) {
+  switch (gatewayUrl) {
+    case 'https://ipfs.io':
+      return Object.freeze({
+        RATE_LIMIT_REQUESTS: 800,
```
vasco-santos, line 80: This one is quite difficult to obtain, considering the infrastructure setup.

ipfs.io public gateway uses several different techniques on load balancing and rate limiting, which makes it difficult to predict the actual rate limit value.

ipfs.io gateway starts by geo routing requests, followed by load balancing. The rate limit is set per load balancer. For instance, same {IP_ADDR, URI} are limited at1/second or 15/minute to a particular load balancer. Then there is a global 800/s on each Load Balancer.

Moreover, it also uses [bursting techniques](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req) that will start by delaying responses before it actually fails.

vasco-santos: I am thinking on performing the load tests once we have this new set of PRs merged in and see if we end up having any gateways rate limited. If so, try to tweak a bit the numbers

packages/gateway/src/durable-objects/gateway-rate-limits.js
```diff
@@ -0,0 +1,100 @@
...
+
+const MINUTE = 1000 * 60
+
+/**
+ * Get rate limiting characteristics of a given Gateway.
+ *
+ * @param {string} gatewayUrl
+ * @return {RateLimitCharacteristics}
+ */
+function getRateLimitingCharacteristics(gatewayUrl) {
```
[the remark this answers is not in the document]
alanshaw: Hard code for now seems fine.

packages/gateway/src/durable-objects/gateway-metrics.js
```diff
@@ -4,20 +4,24 @@
  * @property {number} totalSuccessfulRequests total number of successful requests
  * @property {number} totalFailedRequests total number of requests failed
  * @property {number} totalWinnerRequests number of performed requests where winner
+ * @property {number} totalRateLimitedErroredRequests number of performed requests prevented from rate limiting
+ * @property {number} totalRateLimitedPreventedRequests number of performed requests prevented from rate limiting
```
alanshaw, line 8: What is the difference?

packages/gateway/src/durable-objects/gateway-metrics.js
```diff
@@ -4,20 +4,24 @@
...
  * @property {Record<string, number>} responseTimeHistogram
  *
  * @typedef {Object} ResponseStats
  * @property {boolean} ok request was successful
  * @property {number} [responseTime] number of milliseconds to get response
- * @property {boolean} [winner]
+ * @property {boolean} [winner] response was from winner gateway
+ * @property {boolean} [rateLimitPrevented] response not obtained as request did not happen to prevent rate limit blockage
+ * @property {boolean} [rateLimitErrored] rate limit was not prevented, but worker was still rate limited
```
alanshaw, line 16: Can this just be error response, but we record the status code for errors so we can determine? It would be good to not special case this.

packages/gateway/src/durable-objects/gateway-metrics.js
```diff
@@ -4,20 +4,24 @@
...
+ * @property {number} totalRateLimitedErroredRequests number of performed requests prevented from rate limiting
+ * @property {number} totalRateLimitedPreventedRequests number of performed requests prevented from rate limiting
  * @property {Record<string, number>} responseTimeHistogram
  *
  * @typedef {Object} ResponseStats
  * @property {boolean} ok request was successful
  * @property {number} [responseTime] number of milliseconds to get response
- * @property {boolean} [winner]
+ * @property {boolean} [winner] response was from winner gateway
+ * @property {boolean} [rateLimitPrevented] response not obtained as request did not happen to prevent rate limit blockage
```
alanshaw, line 15:
```suggestion
 * @property {boolean} [rateLimited] request not sent to upstream gateway as it exceeds the rate limit
```

packages/gateway/src/durable-objects/gateway-rate-limits.js
```diff
@@ -0,0 +1,100 @@
...
+  switch (gatewayUrl) {
+    case 'https://ipfs.io':
+      return Object.freeze({
```
alanshaw, line 79: I'd freeze at the call site if you need to prevent augmentations...but I don't think its necessary - the only code touching this object is owned by us. I'd only consider freezing in exports from libraries.

alanshaw (changes requested):

packages/gateway/src/durable-objects/gateway-metrics.js
vasco-santos: In the PR original message I added some details about this: 

> Timings on when the actual request reaches the gateway might influence negatively and we end up being rate limited anyway. While we work on further improvements with the gateways, we will track metrics for "rate limited prevented requests" and "rate limit errored requests". 

If we end up having actual errors of rate limit (that were not prevented), let's count them so that we can improve the tracking to avoid them.

vasco-santos: I will change these to:

```js
/*
 * @property {number} totalRateLimitedErroredRequests number of performed requests that errored with rate limit code
 * @property {number} totalRateLimitedPreventedRequests number of requests not sent to upstream gateway as it would exceed the rate limit
 */
```

packages/gateway/src/durable-objects/gateway-metrics.js
vasco-santos: Changing to

```js
@property {boolean} [rateLimitPrevented] request not sent to upstream gateway as it would exceed the rate limit
```

I think `rateLimited` is not the wording because we were not rate limited by the gateway, but we prevented a request from triggering a rate limit to us. This is why I called `prevented` as we kept it from happening. What do you think?

packages/gateway/src/durable-objects/gateway-metrics.js
vasco-santos: The motivation to track this is to understand how likely we will be rate limited any way, even with the rate limit preventing strategy. For keeping track of general errors we will have sentry + logtail.

This `ResponseStats` come from a fetch body JSON. So, I would need to serialize the Error instance on the gateway side and in the Durable Object deserialize the error, in order to check if its status is 429. That's the main reason to send the boolean.

Do you see any strong reasons to go through the serialization way here instead of as is?

vasco-santos: Either way, I changed jsdocs here to

```
@property {boolean} [rateLimitErrored] request errored with rate limit code without being prevented
```

packages/gateway/src/durable-objects/gateway-rate-limits.js
vasco-santos: Good call, thanks! removed freeze

packages/gateway/src/metrics.js
```diff
@@ -91,23 +102,81 @@ export async function metricsGet(request, env, ctx) {
...
     `# TYPE nftgateway_successful_requests_total counter`,
     ...env.ipfsGateways.map(
       (gw) =>
-        `nftgateway_successful_requests_total{gateway="${gw}",env="${env.ENV}"} ${metricsCollected.ipfsGateways[gw].totalSuccessfulRequests}`
+        `nftgateway_successful_requests_total{gateway="${gw}",env="${
+          env.ENV
+        }"} ${
+          metricsCollected.ipfsGateways[gw].totalResponsesByStatus[
+            HTTP_SUCCESS_CODE
+          ] || 0
```
vasco-santos, line 116: @alanshaw decided to track this individually to have Prometheus metrics to be persisted with a different key to success to have a prometheus name tag, and only use the other key with "failed requests"

If you disagree, happy to change it to simply `nftgateway_requests_by_status_total` and have everything under the same name. 

The main reason for this decision was that I wanted to have a metric that includes a count of all failed requests here, and for queries to Prometheus seemed more reasonable to have success separate. Lmk what you think

packages/gateway/src/metrics.js
```diff
@@ -91,23 +102,81 @@ export async function metricsGet(request, env, ctx) {
...
+                env.ENV
+              }",status="${status}"} ${
+                metricsCollected.ipfsGateways[gw].totalResponsesByStatus[
+                  status
+                ] || 0
+              }`
+          )
+          .join('\n')
+      })
+      .filter((e) => !!e),
```
vasco-santos, line 154: this filter is annoying, but was the sanest way of not having empty lines when status errors only exist for a subset of known gateways

packages/gateway/src/gateway.js
```diff
@@ -77,18 +85,36 @@ export async function gatewayGet(request, env, ctx) {
     // forward winner gateway response
     return winnerGwResponse.response
   } catch (err) {
+    let responses
+    // We can already settle requests if Aggregate Error, as all promises were already rejected
+    if (err instanceof AggregateError) {
+      responses = await pSettle(gatewayReqs)
+    }
```
alanshaw, line 92: What happens if some promises fulfilled but none were `response.ok` and the rest rejected? Do we still get an aggregate error?

Yes, the filter causes a `FilterError` to be thrown here: https://github.com/sindresorhus/p-some/blob/a7030ea6ad9971867ba5dfdb09034a416f3cf8e3/index.js#L60-L62

I think this will _always_ be an `AggregateError` in our case. Or in other words, all of our promises will have rejected or been filtered (effectively rejected) here. We should _always_ settle here to get the responses:

```suggestion
    // All promises will have been rejected or filtered (effectively rejected) here.
    responses = await pSettle(gatewayReqs)
```

packages/gateway/src/constants.js
```diff
@@ -1,3 +1,5 @@
 export const METRICS_CACHE_MAX_AGE = 10 * 60 // in seconds (10 minutes)
 export const CIDS_TRACKER_ID = 'cids'
 export const SUMMARY_METRICS_ID = 'summary-metrics'
+export const RATE_LIMIT_HTTP_ERROR_CODE = 429
+export const HTTP_SUCCESS_CODE = 200
```
alanshaw, line 5:
```suggestion
export const HTTP_STATUS_RATE_LIMITED = 429
export const HTTP_STATUS_SUCCESS = 200
```

packages/gateway/src/gateway.js
```diff
@@ -77,18 +85,36 @@ export async function gatewayGet(request, env, ctx) {
...
         await Promise.all(
-          responses.map((r) =>
+          gatewayResponses.map((r) =>
             updateGatewayMetrics(request, env, r.value, false)
           )
         )
       })()
     )
 
+    // Redirect if all failed and at least one gateway was rate limited
```
alanshaw, line 106: I think we only want to redirect when all the gateways are rate limiting us (or we prevented the request because it would cause rate limiting)...because we're literally not able to service the request. If we got error responses from everyone is probably because the content could not be retrieved OR a problem with the content (not a unixfs thing). It's unlikely that all gateways will be down at the same time and in these cases we should not redirect and get the user to send another request because it's unlikely it'll work, OR all the gateways are on 🔥  and they don't need extra traffic.

packages/gateway/src/metrics.js
alanshaw: I think it'll be unexpected to not see 200 in `nftgateway_requests_by_status_total`...

If you want to have a success vs failure metric for convenience then that's fine but I'd still include 200 in `nftgateway_requests_by_status_total`. You can match a tag via regexp in prometheus so it should be easy to group failure like `/^[45][0-9][0-9]$/` and success like `/^2[0-9][0-9]$/` or something. I won't make it _loads_ easier. Personally I wouldn't complicate it here but it's up to you.

The only other concern with explicitly pulling out 200 is I don't know that _all_ successfull requests from the gateway are 200...but I guess probably yes.

packages/gateway/src/metrics.js
alanshaw: Ya, I think not including success is unnecessarily complicated.

packages/gateway/test/cache.spec.js
```diff
@@ -21,5 +21,6 @@ test('Caches content', async (t) => {
   t.is(await response.text(), content)
 
   const cachedRes = await caches.default.match(url)
-  t.is(await cachedRes.text(), content)
+  // Miniflare cache sometimes is not yet setup...
+  cachedRes && t.is(await cachedRes.text(), content)
```
alanshaw, line 25: hmm, we need to get to the bottom of that...it will be a false positive. Maybe skip the test and open an issue?

alanshaw (changes requested):

packages/gateway/src/metrics.js
vasco-santos: This is not because of success, but because if there is for example status `429` for gateway x, but not for gateway `y`, we wend up with an empty entry

packages/gateway/src/metrics.js
vasco-santos: Yes, but my current naming is `nftgateway_failed_requests_by_status_total`. Anyway, I will just remove failed and add 200 then.

packages/gateway/src/gateway.js
vasco-santos: That's right, thanks

packages/gateway/src/gateway.js
vasco-santos: Yes, if we have one failure that was not rate limit related we should not redirect!

But, there is an edge case here that I was including. For instance: What if we prevent two requests from running and the other is rate limited, or better, if we prevent all requests. So, what we need here is a conditional with:

All response status are `HTTP_STATUS_RATE_LIMITED` OR we prevented the request from happening

packages/gateway/test/cache.spec.js
vasco-santos: https://github.com/nftstorage/nft.storage/issues/1231

packages/gateway/src/metrics.js
alanshaw: Oh ok sorry!

alanshaw (approved):
LGTM

packages/gateway/src/gateway.js
```diff
@@ -1,28 +1,33 @@
 /* eslint-env serviceworker, browser */
 /* global Response caches */
 
-import pAny from 'p-any'
+import pAny, { AggregateError } from 'p-any'
```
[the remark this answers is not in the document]
vasco-santos: I think your view was outdated 😅  I removed it in last commit

PR #1179, opened 2022-01-27
Feat/issue 1160 rework files td actions

cmunns:
- allow copyButton to work asLink
- add explicit [hologram] prop to btn
- add reacty-tiny-popover

Screen Shot 2022-01-27 at 10 25 31 AM


alanshaw (approved):
I love this - please can we release asap!

Small request - please can you make sure that the "View" link for Nft type uploads goes to cid/metadata.json?

JeffLowe:
The "View" link was intended to be the gateway URL (which actually right now in this PR it's the ipfs uri and needs to change).  @alanshaw sounds like you are saying that we need an option that leads to the json metadata when it's an nft.   

@cmunns let's adjust the current "View" link to be the gateway url (same as the hyperlinked CID for now)

@alanshaw I'll make a new ticket (so we don't block getting this out) to add a "view metadata" as an additional option in that menu that leads to the <cid>/metadata.json.

PR #1180, opened 2022-01-28
feat: gateway add cloudflare cache layer

vasco-santos:
This PR adds a caching layer to gateway requests in the context of https://github.com/nftstorage/nft.storage/issues/823

It relies on [Cloudflare cache](https://developers.cloudflare.com/workers/runtime-apis/cache) to better control reading and writing from Cloudflare CDN. Moreover, we now track a metric with number of requests that we immediately return a cached response.

It uses request URL as key for caching instead of the request object to avoid different caching with different [incomingrequestcfproperties](https://developers.cloudflare.com/workers/runtime-apis/request#incomingrequestcfproperties)

GenericMetrics Durable Object also renamed to SummaryMetrics by previous conv, considering a need to a migration here

### DIFF
--- a/packages/gateway/ava.config.js
+++ b/packages/gateway/ava.config.js
@@ -4,5 +4,6 @@ export default {
   },
   files: ['test/*.spec.js'],
   timeout: '5m',
+  concurrency: 1,
   nodeArguments: ['--experimental-vm-modules'],
 }
--- a/packages/gateway/src/gateway.js
+++ b/packages/gateway/src/gateway.js
@@ -156,7 +156,7 @@ async function updateSummaryCacheMetrics(request, env) {
   const id = env.summaryMetricsDurable.idFromName(SUMMARY_METRICS_ID)
   const stub = env.summaryMetricsDurable.get(id)
 
-  await stub.fetch(_getDurableRequestUrl(request, 'cache-update'))
+  await stub.fetch(_getDurableRequestUrl(request, 'metrics/cache'))
 }
 
 /**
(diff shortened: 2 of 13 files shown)


packages/gateway/ava.config.js
```diff
@@ -4,5 +4,6 @@ export default {
   },
   files: ['test/*.spec.js'],
   timeout: '5m',
+  concurrency: 1,
```
vasco-santos, line 7: Running the test cases in parallel makes cache test flakey, likely because other Miniflare instances run and they reset the cache 🤷🏼  Needed to make concurrency 1 to avoid it

packages/gateway/test/utils.js
```diff
@@ -14,7 +14,7 @@ export function getMiniflare() {
     modules: true,
     durableObjects: {
       GATEWAYMETRICS: 'GatewayMetrics1',
-      GENERICMETRICS: 'GenericMetrics1',
+      SUMMARYMETRICS: 'SummaryMetrics0',
```
alanshaw, line 17: Do we need these - they are read from the wrangler.toml aren't they?

packages/gateway/src/gateway.js
```diff
@@ -58,6 +68,8 @@ export async function gatewayGet(request, env, ctx) {
         await Promise.all([
           storeWinnerGwResponse(request, env, winnerGwResponse),
           settleGatewayRequests(),
+          // Cache request URL in Cloudflare CDN
+          cache.put(request.url, winnerGwResponse.response.clone()),
```
alanshaw, line 72: We can only cache things < 512MB. We should check `Content-Length` before attempting.

This just occurred to me - it would be really great to know the distribution of file sizes the gateway is serving. Can we please add file size to the metrics as a histogram? (separate PR obvs)

alanshaw (approved):

packages/gateway/src/gateway.js
vasco-santos: Good idea! I will do the validation and track metrics in follow up PR

PR #1187, opened 2022-01-28
Fix: issue 1152 cid truncation

cmunns:
Screen Shot 2022-01-28 at 8 49 47 AM
Using flexbox we can have the width get inherited so that it fits as much as it can before truncating the cid.

also updates the popover to prevent it from getting cutoff
Screen Shot 2022-01-28 at 9 14 03 AM


JeffLowe (approved):
Checked on desktop and mobile.  functional and visual looks good.

alanshaw (approved):

Issue #1182, opened 2022-01-28
Gateway metrics renaming per Prometheus best practises 


vasco-santos:
To follow https://prometheus.io/docs/practices/naming/

PR #1194, opened 2022-01-31
fix: gateway prometheus metrics best practices naming

vasco-santos:
This PR fixes prometheus metrics naming per best practices and conventions defined in https://prometheus.io/docs/practices/naming/ . Thanks @alanshaw for the pointer ❤️ 

In summary, the changes are:
- A metric name should have a (single-word) application prefix relevant to the domain the metric belongs to => made it nftgateway instead of previous two words
- A metric name should use base units => milliseconds converted to seconds
- A metric name should have a suffix describing the unit, in plural form (or total for a unit-less accumulating count) => all "total" are now suffixes and unit applied where applicable (response time seconds)

Closes https://github.com/nftstorage/nft.storage/issues/1182

### DIFF
--- a/packages/gateway/src/metrics.js
+++ b/packages/gateway/src/metrics.js
@@ -68,9 +68,9 @@ export async function metricsGet(request, env, ctx) {
   }
 
   const metrics = [
-    `# HELP nftstorage_gateway_total_cache_hit_responses Total cached responses returned.`,
-    `# TYPE nftstorage_gateway_total_cache_hit_responses counter`,
-    `nftstorage_gateway_total_cache_hit_responses{env="${env.ENV}"} ${metricsCollected.summaryMetrics.totalCachedResponses}`,
+    `# HELP nftgateway_cache_hit_responses_total Total cached responses returned.`,
+    `# TYPE nftgateway_cache_hit_responses_total counter`,
+    `nftgateway_cache_hit_responses_total{env="${env.ENV}"} ${metricsCollected.summaryMetrics.totalCachedResponses}`,
     `# HELP nftgateway_winner_requests_total Total winner requests.`,
     `# TYPE nftgateway_winner_requests_total counter`,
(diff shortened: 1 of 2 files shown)


packages/gateway/src/metrics.js
```diff
@@ -68,63 +68,65 @@ export async function metricsGet(request, env, ctx) {
...
-    `# TYPE nftstorage_gateway_total_winner_response_time counter`,
-    `nftstorage_gateway_total_winner_response_time{env="${env.ENV}"} ${metricsCollected.summaryMetrics.totalWinnerResponseTime}`,
-    `# HELP nftstorage_gateway_total_winner_successful_requests Total successful requests.`,
-    `# TYPE nftstorage_gateway_total_winner_successful_requests counter`,
-    `nftstorage_gateway_total_winner_successful_requests{env="${env.ENV}"} ${metricsCollected.summaryMetrics.totalWinnerSuccessfulRequests}`,
-    `# HELP nftstorage_gateway_total_response_time Average response time.`,
-    `# TYPE nftstorage_gateway_total_response_time gauge`,
+    `# HELP nftstorage_gateway_total_cache_hit_responses Total cached responses returned.`,
+    `# TYPE nftstorage_gateway_total_cache_hit_responses counter`,
+    `nftstorage_gateway_total_cache_hit_responses{env="${env.ENV}"} ${metricsCollected.summaryMetrics.totalCachedResponses}`,
```
alanshaw, line 73: Should these be `nftgateway` and suffixed with `total`?

alanshaw (changes requested):

packages/gateway/src/metrics.js
vasco-santos: Yes, sorry! I don't know why I did not change these lol

alanshaw (approved):

Issue #1181, opened 2022-01-28
Limit gateway cache size and keep track of file size distribution


vasco-santos:
Great idea from @alanshaw 

> We can only cache things < 512MB. We should check Content-Length before attempting.
> This just occurred to me - it would be really great to know the distribution of file sizes the gateway is serving. Can we please add file size to the metrics as a histogram? (separate PR obvs)

PR #1197, opened 2022-01-31
fix: only cache when content length smaller than max object size

vasco-santos:
Part of https://github.com/nftstorage/nft.storage/issues/1181 implemented. Only cache when content smaller than max object size limit https://developers.cloudflare.com/workers/platform/limits#cache-api-limits

Will track metrics in follow up PR

### DIFF
--- a/packages/gateway/src/constants.js
+++ b/packages/gateway/src/constants.js
@@ -1,3 +1,4 @@
+export const CLOUDFLARE_MAX_OBJECT_SIZE = 512 * Math.pow(1024, 2) // 512MB to bytes
 export const METRICS_CACHE_MAX_AGE = 10 * 60 // in seconds (10 minutes)
 export const CIDS_TRACKER_ID = 'cids'
 export const SUMMARY_METRICS_ID = 'summary-metrics'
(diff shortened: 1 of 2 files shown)


packages/gateway/src/constants.js
```diff
@@ -1,3 +1,4 @@
+export const CLOUDFLARE_MAX_OBJECT_SIZE = 512 * Math.pow(1024, 2) // 512MB to bytes
```
alanshaw, line 1: Maybe add "cache" to the name so we know it relates the cache that cloudflare provides!

```suggestion
export const CF_CACHE_MAX_OBJECT_SIZE = 512 * Math.pow(1024, 2) // 512MB to bytes
```

packages/gateway/src/gateway.js
```diff
@@ -65,11 +69,16 @@ export async function gatewayGet(request, env, ctx) {
...
+          winnerGwResponse.response.headers.get('content-length')
+        )
+
         await Promise.all([
           storeWinnerGwResponse(request, env, winnerGwResponse),
           settleGatewayRequests(),
-          // Cache request URL in Cloudflare CDN
-          cache.put(request.url, winnerGwResponse.response.clone()),
+          // Cache request URL in Cloudflare CDN if smaller than CLOUDFLARE_MAX_OBJECT_SIZE
+          contentLengthMb < CLOUDFLARE_MAX_OBJECT_SIZE &&
```
alanshaw, line 80: I assume the max is inclusive...

```suggestion
          contentLengthMb <= CLOUDFLARE_MAX_OBJECT_SIZE &&
```

alanshaw (changes requested):

alanshaw (approved):

Issue #1186, opened 2022-01-28
tmpfs is not cleared after adding files with storeDirectory


ljmf00:
By using the `client.storeDirectory`, after adding files to the IPFS network, the tmpfs /tmp is not cleaned up properly, cluttering it over time.

```js
import { NFTStorage, File } from 'nft.storage'
import { pack } from 'ipfs-car/pack'
import process from 'process'
import minimist from 'minimist'
import fs from 'fs'

const args = minimist(process.argv.slice(2))

const NS_API_KEY = fs.readFileSync('ns-api-key.txt','utf8').trim();
const nsclient = new NFTStorage({ token: NS_API_KEY })
const nsfiles = []

function getFiles (dir, files_){
    files_ = files_ || [];
    var files = fs.readdirSync(dir);
    for (var i in files){
        var name = dir + '/' + files[i];
        if (fs.statSync(name).isDirectory()){
            getFiles(name, files_);
        } else {
            files_.push(name);
        }
    }
    return files_;
}

for (const path of args._)
  nsfiles.push(...getFiles(path)
    .map(f => new File([fs.readFileSync(f)], f)))

const nscid = await nsclient.storeDirectory(nsfiles)
console.log(nscid)
```

vasco-santos:
@ljmf00 thanks for the replication case. We had a bug on `storeBlob` and `storeDirectory` which is addressed in https://github.com/nftstorage/nft.storage/pull/1198

ljmf00:
Thanks for fixing it!

HomeSchoolDev:
This still appears to be happening on the latest 7.0.0 store()

Issue #980, opened 2021-12-15
Generate CARs for large files/directories via streaming and let NFT.Storage accept them


dchoi27:
One large technical limitation for uploading large files via the API is the client-side generation of CAR files using our current toolset. Generally speaking, users have out of memory and other performance issues when creating large CAR files using ipfs-car + carbites:
- (Web3.Storage) https://github.com/web3-storage/web3.storage/issues/318
- (Web3.Storage) https://github.com/web3-storage/web3.storage/issues/341
- https://github.com/web3-storage/ipfs-car/issues/69

The way Pinata and others handle this is getting users to upload data to a local IPFS node first and remote pinning it to their account, but for various reasons we want to focus on supporting a path to directly upload.

There are a number of issues to make this a possibility, and then subsequently make it useful for users:
- Allow unixfs-importer (to take?) stream: https://github.com/ipfs/js-ipfs-unixfs/issues/186
- (Potentially) Write streaming CAR file generator
- Back-end work to handle any differences in the data it receives (since CAR chunks created via streaming won't have the root CID of the entire file in each CAR, so the service will need to figure out how to handle and/or display this info to users)

@mikeal feel free to edit this to be more precise!
- Expose these changes in the JS client (e.g. add to `store` methods when a large file is uploaded) and other relevant places (e.g., Electron app in #981)

Gozala:
@dchoi27 reading through the linked issues 

> (Web3.Storage)
Example code fails to put a 9GB file web3-storage/web3.storage#318
(Web3.Storage)
too many open files web3-storage/web3.storage#341
Crash on packing web3-storage/ipfs-car#69

From what I can tell based on comments those are caused by yet unknown bugs in the implementation (not necessarily by an API limitation). That said, current API does not provides a good way to manage resource usage. This makes it

1. Hard to figure out where those issues are coming from
2. Impossible for user to control resource usage

What I am hoping to achieve with #186 is design an API in which user can have direct control of how much memory importer can allocate which I believe would significantly reduce a surface for such bugs.


That alone will not however fully address our problems, that is because car file format contains a header with root CIDs, which can not be known for 9GB file until all of it is chunked into blocks Which implies that program either:

a. chunks it up and stores blocks somewhere on disk
b. blocks are uploaded first while root with dag is uploaded later.


#186 aims to support both options, however I think we'd need to do additional work for either of those options.

dchoi27:
[...]

PR #1198, opened 2022-01-31
fix: client close blockstore on store directory and store blob

vasco-santos:
Given we use a CarReader now, we can only close the blockstore once pack is finished.

Also updated @web-std/fetch to avoid `Error: Cannot find module '/home/runner/work/nft.storage/nft.storage/node_modules/@web-std/fetch/dist/index.cjs'`

Closes https://github.com/nftstorage/nft.storage/issues/1186

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -104,7 +104,7 @@ class NFTStorage {
     let cidString
 
     try {
-      const { cid, car } = await NFTStorage.encodeBlob(blob, blockstore)
+      const { cid, car } = await NFTStorage.encodeBlob(blob, { blockstore })
       await NFTStorage.storeCar(service, car)
       cidString = cid.toString()
     } finally {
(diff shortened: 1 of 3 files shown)


packages/client/src/lib.js
```diff
@@ -99,9 +100,18 @@ class NFTStorage {
...
   static async storeBlob(service, blob) {
-    const { cid, car } = await NFTStorage.encodeBlob(blob)
-    await NFTStorage.storeCar(service, car)
-    return cid.toString()
+    const blockstore = new Blockstore()
+    let cidString
+
+    try {
+      const { cid, car } = await NFTStorage.encodeBlob(blob, blockstore)
+      await NFTStorage.storeCar(service, car)
```
vasco-santos, line 108: We can as an alternative do this within storeCar, but we would need to provide blockstore here, which feels odd as we don't even use `ipfs-car` related things in the function. Also, wrapping everything to guarantee error handling looked the best option

ljmf00: In my opinion, this can be redesigned as if a user want to store a big chunk, let's say 10GB, they need to have it available in tmpfs at once. I would recommend to have only the last pending blocks in the tmpfs and delete them after uploading.

vasco-santos: We are working on a new solution to pack files that will stream directly instead of using a temporary blockstore to keep the blocks. So, this will be an enhancement that we will ship soon to improve this behaviour https://github.com/nftstorage/nft.storage/issues/980

ljmf00 (approved):

hugomrdias (approved):
LGTM

PR #1204, opened 2022-02-01
feat: add local debug configuration for vscode.

redaphid:
This can add local debugging for local dev!

I'm not sure what our policy is re: committing IDE config files, but I thought I'd share what I'm using to debug `api` right now.
This will let you attach a debugger, step through code, set breakpoints etc.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "jest.jestCommandLine": "npx jest"
+}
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "jest.jestCommandLine": "npx jest"
```
hugomrdias, line 2: what's this for ?

hugomrdias (changes requested):

PR #1206, opened 2022-02-01
feat: track content length histogram

vasco-santos:
Adds content length histogram tracking, as well as content length of cached responses, so that we can compare the amount of data we had cached

Closes https://github.com/nftstorage/nft.storage/issues/1181

### DIFF
--- a/packages/gateway/test/metrics.spec.js
+++ b/packages/gateway/test/metrics.spec.js
@@ -21,6 +21,13 @@ test('Gets Metrics content when empty state', async (t) => {
     true
   )
   t.is(metricsResponse.includes('nftgateway_winner_requests_total'), true)
+  t.is(metricsResponse.includes(`_responses_content_length_total{le=`), true)
+  t.is(
+    metricsResponse.includes(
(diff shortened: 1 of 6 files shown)


packages/gateway/test/metrics.spec.js
```diff
@@ -70,6 +77,26 @@ test('Gets Metrics content', async (t) => {
   const response = await mf.dispatchFetch('http://localhost:8787/metrics')
   const metricsResponse = await response.text()
 
+  // content length of responses is 23
```
alanshaw, line 80: ? is zero below

alanshaw (changes requested):

packages/gateway/test/metrics.spec.js
vasco-santos: What do you mean below? Want to get the metrics before the requests and see if content length is zero?

vasco-santos: added that case

alanshaw (approved):

PR #1208, opened 2022-02-01
fix: update gateway miniflare to final version

vasco-santos:
Also simplifies `wrangler.toml` to have unique entry for Durable Objects


alanshaw (approved):
💪

PR #1212, opened 2022-02-02
fix: add durable object types

vasco-santos:
Adds Cloudflare Durable object types inline and the `@cloudflare/workers-types` dev dependency. The Durable objects types are not exported from `@cloudflare/workers-types`

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -33,7 +33,7 @@
     "uint8arrays": "^3.0.0"
   },
   "devDependencies": {
-    "@cloudflare/workers-types": "^3.1.1",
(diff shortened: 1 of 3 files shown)


hugomrdias (approved):
We should probably open an issue upstream about.

packages/gateway/package.json
```diff
@@ -28,6 +28,7 @@
     "toucan-js": "^2.5.0"
   },
   "devDependencies": {
+    "@cloudflare/workers-types": "^3.3.1",
```
alanshaw, line 31: Did you mean to add this as a dep? Not being used...right?

PR #1224, opened 2022-02-02
feat: add tab switcher component to docs

yusefnapora:
This adds a `<Tabs>` component with `<TabItem>`s, shamelessly inspired by Docusaurus's Tabs component. It has the same API, but doesn't support tab groups.

nextra-tab-switcher


packages/website/components/mdx/tabs.js
```diff
@@ -0,0 +1,133 @@
...
+  const [selectedValue, setSelectedValue] = useState(defaultValue)
+  /**
+   * @type Array<HTMLLIElement | null>
+   */
+  const tabRefs = []
+
+  /**
+   * @param {React.FocusEvent<HTMLLIElement> | React.MouseEvent<HTMLLIElement>} e
+   */
+  const handleTabChange = (e) => {
```
dashcraft, line 72: nitpick (non-blocking): We could probably just destructure current target on event and avoid arbitrary assignment

packages/website/components/mdx/tabs.js
```diff
@@ -0,0 +1,133 @@
...
+    const newTab = e.currentTarget
+    const newTabIndex = tabRefs.indexOf(newTab)
+    if (newTabIndex < 0) {
```
dashcraft, line 75: praise:

I think returning here is probably better than throwing an error as well.

packages/website/components/mdx/tabs.js
```diff
@@ -0,0 +1,133 @@
...
+    Children.map(props.children, (child) => {
+      if (isValidElement(child) && isTabItem(child)) {
+        return child
+      }
+      throw new Error(
+        `Unexpected child in Tabs element. All children of <Tabs> should be a <TabItem> with a unique "value" prop`
+      )
+    }) || []
+
+  const values = children.map(({ props: { value, label, attributes } }) => ({
```
dashcraft, line 50: suggestion (non-blocking): Some of this might be good to memoize, I haven't run this in a profiler, but small calculation functions usually work well with memoization for minor performance improvements.

dashcraft (approved):
Approved, left some non blocking comments

PR #1226, opened 2022-02-02
fix: update tooltip styles to match site

cmunns:
... should also addresses ios safari 'actions' button color.
New tooltip looks like this:
Screen Shot 2022-02-02 at 12 58 24 PM


dashcraft (approved):
Style and color changes only, LGTM!

PR #1227, opened 2022-02-03
feat: gateway with x-forwarded-for ip for ipfs.io gateway

vasco-santos:
This PR adds [`x-forwarded-for`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) IP from client to work around rate limiting from ipfs.io gateway. As part of support in https://github.com/protocol/bifrost-infra/issues/1609

The client IP is obtained from https://developers.cloudflare.com/workers/runtime-apis/headers#request-headers

I made this specific to the ipfs.io gateway to make clear that only it has support for this, but we could also just send it in all of them.

Tested by creating a route where response is just the headers of request in staging and confirmed it has my public IP.

image

### DIFF
--- a/packages/gateway/src/gateway.js
+++ b/packages/gateway/src/gateway.js
@@ -173,7 +173,7 @@ async function gatewayFetch(
   try {
     response = await fetch(`${ipfsUrl.toString()}/${cid}${pathname}`, {
       signal: controller.signal,
-      headers: _getHeadersForGateway(gwUrl, request),
+      headers: getHeaders(request),
     })
   } finally {
     clearTimeout(timer)
@@ -189,14 +189,9 @@ async function gatewayFetch(
 }
 
 /**
- * @param {string} gwUrl
  * @param {Request} request
  */
-function _getHeadersForGateway(gwUrl, request) {
-  if (gwUrl !== 'https://ipfs.io') {
-    return {}
-  }
-
+function getHeaders(request) {
   const existingProxies = request.headers['X-Forwarded-For']
     ? `, ${request.headers['X-Forwarded-For']}`
     : ''
--- a/packages/gateway/src/gateway.js
+++ b/packages/gateway/src/gateway.js
@@ -197,8 +197,11 @@ function _getHeadersForGateway(gwUrl, request) {
     return {}
   }
 
+  const existingProxies = request.headers['X-Forwarded-For']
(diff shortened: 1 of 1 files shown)


thattommyhall:
Check if x-forwarded-for is already set, the spec is to append to the list if it already exists

vasco-santos:
> Check if x-forwarded-for is already set, the spec is to append to the list if it already exists

Good catch @thattommyhall  thanks. Just added that logic too

packages/gateway/src/gateway.js
```diff
@@ -147,6 +150,23 @@ async function _gatewayFetch(
...
+function _getHeadersForGateway(gwUrl, request) {
+  if (gwUrl !== 'https://ipfs.io') {
+    return {}
+  }
+
+  const existingProxies = request.headers['X-Forwarded-For']
+    ? `, ${request.headers['X-Forwarded-For']}`
+    : ''
+  return {
+    'X-Forwarded-For': `${request.headers['cf-connecting-ip']}${existingProxies}`,
```
alanshaw, line 166: Shouldn't this have a delimiter?

packages/gateway/src/gateway.js
```diff
@@ -147,6 +150,23 @@ async function _gatewayFetch(
   return gwResponse
 }
 
+/**
+ * @param {string} gwUrl
+ * @param {Request} request
+ */
+function _getHeadersForGateway(gwUrl, request) {
```
alanshaw, line 157: Underscore prefix unnecessary?

packages/gateway/src/gateway.js
```diff
@@ -147,6 +150,23 @@ async function _gatewayFetch(
...
+  if (gwUrl !== 'https://ipfs.io') {
+    return {}
+  }
```
alanshaw, line 160: I'd just send this to all of them...

alanshaw (changes requested):

packages/gateway/src/gateway.js
vasco-santos: The delimeter is set above within `existingProxies` var

alanshaw: Oh damn, yes, I missed that!

alanshaw (approved):

PR #1237, opened 2022-02-04
feat: gateway track redirect counts

vasco-santos:
Adds a counter to track redirects when rate limited to have a metric and better view if all fail in any situation.


alanshaw (commented):
How come a separate DO and not part of gateway metrics?

vasco-santos:
> How come a separate DO and not part of gateway metrics?

I though about adding it to summary. But, the goal is to not have redirects with X-Forwarded-For header, plus other potential improvements with CF and Pinata.

So, decided to have them in separate to be easier to just remove later on, if we don't want it.

alanshaw (approved):

PR #1240, opened 2022-02-04
clarify file size/upload limits

mishmosh:
clarify that the 31gib limit is per upload, and note the extra step required for http api users

### DIFF
--- a/packages/website/lib/faqContent.js
+++ b/packages/website/lib/faqContent.js
@@ -48,8 +48,11 @@ const faqContent = {
   ),
   nftSizeRestrictions: (
     <p className="lh-copy white mb4">
-      NFT.Storage can store NFTs up to <strong>31GiB </strong>
-      in size!
+      NFT.Storage accepts storage requests up to <strong>31GiB</strong> in size! Each upload can include a single file or a directory of files. (If you are using the HTTP API, you'll need to do some manual splitting for files over 100MB. See the{' '}
+      <Link href="/api-docs">
+        <a className="white">HTTP API docs</a>
+      </Link>{' '}
+      for details.)
     </p>
   ),
   nftBestPractices: (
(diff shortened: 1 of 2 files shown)


yusefnapora:
This is a strange build error:

```
 Error:   51:180  error  `'` can be escaped with `&apos;`, `&lsquo;`, `&#39;`, `&rsquo;`  react/no-unescaped-entities
```

The only `'` I see in the diff is inside `{' '}`, which doesn't seem like it should produce an actual `'` character... maybe just the linter getting confused.

@mishmosh maybe try changing those to double quotes?

packages/website/lib/faqContent.js
```diff
@@ -48,8 +48,11 @@ const faqContent = {
   ),
   nftSizeRestrictions: (
     <p className="lh-copy white mb4">
-      NFT.Storage can store NFTs up to <strong>31GiB </strong>
-      in size!
+      NFT.Storage accepts storage requests up to <strong>31GiB</strong> in size! Each upload can include a single file or a directory of files. (If you are using the HTTP API, you'll need to do some manual splitting for files over 100MB. See the{' '}
```
yusefnapora, line 51: @mishmosh apparently you need to use `&apos;` instead of `'` here.

```suggestion
      NFT.Storage accepts storage requests up to <strong>31GiB</strong> in size! Each upload can include a single file or a directory of files. (If you are using the HTTP API, you&apos;ll need to do some manual splitting for files over 100MB. See the{' '}
```

yusefnapora (changes requested):

yusefnapora (approved):
aha, I "requested changes" before but didn't realize I could commit to your branch :) so no worries; will merge in a sec :)

mishmosh:
@yusefnapora sorry about that, the PR looked green initially so I requested review before the test found that glitch! thx for fixing directly.

yusefnapora:
No worries! I do that all the time 😄

PR #1258, opened 2022-02-04
fix: links for nft type to indicate metadata and add toggle for webview

cmunns:
Updates the links title,text,url if nft type but also bridges a gap I feel is needed. In the event that a user is accessing the webapp but doesn't have IPFS downloaded we should provide a toggle to use the webviewer.

Open to other ideas, but for now I think is a good stop gap to prevent broken links for new users.

Screen Shot 2022-02-04 at 2 31 41 PM


packages/website/pages/files.js
```diff
@@ -302,7 +321,19 @@ export default function Files({ user }) {
...
+                <div className="flex-auto chicagoflf mv3">
+                  <h1>Files</h1>
+                  <label className="webViewer-toggle mt1">
+                    <input
+                      type="checkbox"
+                      name="toggleWebViewer"
+                      className="mr1"
+                      onChange={(e) => handleWebViewerToggle(e)}
+                    />
+                    Use IPFS web viewer instead of IPFS applications where
```
alanshaw, line 333: I'm not sure what "web viewer" means!? It means open on the gateway in your browser?

alanshaw (dismissed):

packages/website/pages/files.js
cmunns: Yes, not sure on the correct wording here. Open to your input...just felt important to have an opt-in for users that want to use https://ipfs.io as the viewer.

Codigo-Fuentes:
URL-fix

packages/website/pages/files.js
JeffLowe: @alanshaw is the concern the label used, or the underlying approach of giving user the choice?

dashcraft (approved):
LGTM

comment (non blocking): this is one of the files I've flagged as needing a refactor in the future

PR #1264, opened 2022-02-07
fix: gateway get durable object request function

vasco-santos:
Fixes two minor issues:
- get durable object URL
- when timeout we were not getting access to gateway URL nor properly handling abort error (code does not exist, at least in CF environment)

### DIFF
--- a/packages/gateway/src/gateway.js
+++ b/packages/gateway/src/gateway.js
@@ -14,6 +14,7 @@ import {
   CF_CACHE_MAX_OBJECT_SIZE,
   HTTP_STATUS_RATE_LIMITED,
   REQUEST_PREVENTED_RATE_LIMIT_CODE,
+  TIMEOUT_CODE,
 } from './constants.js'
 
 /**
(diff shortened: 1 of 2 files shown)


packages/gateway/src/gateway.js
```diff
@@ -193,6 +193,14 @@ async function gatewayFetch(
       signal: controller.signal,
       headers: getHeaders(request),
     })
+  } catch (error) {
+    if (controller.signal.aborted) {
+      return {
+        url: gwUrl,
+        aborted: true,
```
alanshaw, line 200: Can we please be specific that this is a timeout (`timeout: true`), or have `aborted: reason` - where reason is timeout.

We could maybe roll `requestPreventedCode` into this:

```js
{
  url: gwUrl,
  aborted: true,
  reason: REQUEST_PREVENTED_RATE_LIMIT_CODE
}
```

And for timeout:

```js
{
  url: gwUrl,
  aborted: true,
  reason: 'TIMEOUT'
}
```

...for all other successful you set `aborted: false`.

Just trying to get a standard return type out of this.

packages/gateway/src/gateway.js
```diff
@@ -288,6 +296,10 @@ async function updateGatewayMetrics(
   gwResponse,
   isWinner = false
 ) {
+  if (!gwResponse.response?.status && !gwResponse.requestPreventedCode) {
```
alanshaw, line 299: Should this be checking `aborted` instead?

alanshaw (approved):

PR #1265, opened 2022-02-07
feat: use Cluster statusAll method

alanshaw:
This PR switches to using the new `statusAll` method on cluster which now allows us to get the status for a given set of CIDs.

Due to nginx restrictions on request length we currently have to ask for 120 at a time. However it is significantly better than making 120 separate requests.

### DIFF
--- a/packages/cron/src/jobs/pins.js
+++ b/packages/cron/src/jobs/pins.js
@@ -3,6 +3,11 @@ import { consume, transform, pipeline } from 'streaming-iterables'
 
 const log = debug('pins:updatePinStatuses')
 const CONCURRENCY = 5
+/**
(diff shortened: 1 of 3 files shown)


packages/cron/src/jobs/pins.js
```diff
@@ -135,27 +140,26 @@ async function updatePinStatuses(config) {
         offset += limit
       }
     },
+    async function* (source) {
+      for await (let pins of source) {
+        // TODO: remove filter when nft2 has finished copying to nft3
```
vasco-santos, line 145: Can you please open an issue (maybe add to CIC) so that we don't forget?

vasco-santos (approved):
LGTM

PR #1282, opened 2022-02-08
fix: gateway get headers record

vasco-santos:
Fixes get headers function to use the `.get` function. While `['X-Forwarded-For']` was working, `['cf-connecting-ip']` was not working and it was resulting on  `undefined,...` https://github.com/protocol/bifrost-infra/issues/1609#issuecomment-1032018181

This fixes it, already tested


alanshaw (approved):
Yes, `get` does the lowercase I assume

PR #1289, opened 2022-02-08
fix: reset gateway durable objects migrations

vasco-santos:
Resets again all migrations to start a clean state to production


vasco-santos:
Going to merge this as the only code it touches is renaming a class name and is unblocking the release PR to be merged

Issue #1196, opened 2022-01-31
Gateway - Setup Worker Rate limits


vasco-santos:
We need to setup rate limits on the nft.storage gateway worker to avoid abuse.

We need to figure out what ranges we should have and document this

PR #1291, opened 2022-02-08
docs: add gateway usage and rate limit information

vasco-santos:
Adds usage section to the readme, as well as rate limit information

Closes https://github.com/nftstorage/nft.storage/issues/1196

### DIFF
--- a/packages/gateway/README.md
+++ b/packages/gateway/README.md
@@ -39,6 +39,25 @@ You only need to `npm start` for subsequent runs. PR your env config to the `wra
 
 The IPFS gateway for nft.storage is not "another gateway", but a caching layer for NFT’s that sits on top of existing IPFS public gateways.
 
+## Usage
+
+nft.storage Gateway provides subdomain style resolutions maintaining compliance with the [same-origin policy](https://en.wikipedia.org/wiki/Same-origin_policy). The canonical form of access `https://{CID}.ipfs.nft.storage/{optional path to resource}` causes the browser to interpret each returned file as being from a different origin.
+
+```
+curl https://bafkreidyeivj7adnnac6ljvzj2e3rd5xdw3revw4da7mx2ckrstapoupoq.ipfs.nft.storage
+Hello nft.storage! 😎
+```
+
+Please note that subdomain resolution is only supported with [CIDv1](https://docs.ipfs.io/concepts/content-addressing/#identifier-formats) in case-insensitive encoding such as Base32 or Base36. For CIDv0 support, you can use the Path style resolution using the nft.storage API as follows:
+
+```
+curl https://api.nft.storage/ipfs/QmT5NvUtoM5nWFfrQdVrFtvGfKFmG7AHE8P34isapyhCxX
+```
+
+### Rate limiting
+
+nft.storage Gateway is currently rate limited at 200 requests per minute to a given IP Address. In the event of a rate limit, the IP will be blocked for 30 seconds.
+
 ## Persistence
 
 Several metrics per gateway are persisted to track the performance of each public gateway over time. Moreover, the list of gateways that have previously fetched successfully a given CID are also persisted.
--- a/packages/gateway/README.md
+++ b/packages/gateway/README.md
@@ -51,7 +51,7 @@ Hello nft.storage! 😎
 Please note that subdomain resolution is only supported with [CIDv1](https://docs.ipfs.io/concepts/content-addressing/#identifier-formats) in case-insensitive encoding such as Base32 or Base36. For CIDv0 support, you can use the Path style resolution using the nft.storage API as follows:
 
 ```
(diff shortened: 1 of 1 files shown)


packages/gateway/README.md
```diff
@@ -39,6 +39,25 @@ You only need to `npm start` for subsequent runs. PR your env config to the `wra
...
+
+Please note that subdomain resolution is only supported with [CIDv1](https://docs.ipfs.io/concepts/content-addressing/#identifier-formats) in case-insensitive encoding such as Base32 or Base36. For CIDv0 support, you can use the Path style resolution using the nft.storage API as follows:
+
+```
+curl https://api.nft.storage/ipfs/QmT5NvUtoM5nWFfrQdVrFtvGfKFmG7AHE8P34isapyhCxX
+```
+
+### Rate limiting
+
+nft.storage Gateway is currently rate limited at 200 requests per minute to a given IP Address. In the event of a rate limit, the IP will be blocked for 30 seconds.
```
vasco-santos, line 59: 200 requests per minute seems a good default value for browser clients doing requests to nft.storage Gateway. There might be platforms that need more requests. If so, we probably should whitelist IPs that we want to accept more requests.
What do you think @alanshaw @dchoi27 ?

packages/gateway/README.md
```diff
@@ -39,6 +39,25 @@ You only need to `npm start` for subsequent runs. PR your env config to the `wra
...
+
+```
+curl https://bafkreidyeivj7adnnac6ljvzj2e3rd5xdw3revw4da7mx2ckrstapoupoq.ipfs.nft.storage
+Hello nft.storage! 😎
+```
+
+Please note that subdomain resolution is only supported with [CIDv1](https://docs.ipfs.io/concepts/content-addressing/#identifier-formats) in case-insensitive encoding such as Base32 or Base36. For CIDv0 support, you can use the Path style resolution using the nft.storage API as follows:
+
+```
+curl https://api.nft.storage/ipfs/QmT5NvUtoM5nWFfrQdVrFtvGfKFmG7AHE8P34isapyhCxX
```
alanshaw, line 54: Argh, so I was hoping that `https://nft.storage/ipfs/QmT5NvUtoM5nWFfrQdVrFtvGfKFmG7AHE8P34isapyhCxX` would work...is there a way?

alanshaw: We should not put this on the API...

packages/gateway/README.md
alanshaw: Sounds ok for starters.

alanshaw (changes requested):

packages/gateway/README.md
vasco-santos: I will see, likely I can do that! Need to play with DNS records + worker triggers, once I have https://github.com/nftstorage/nft.storage/pull/1064

I will not merge this PR until sort this out

packages/gateway/README.md
dchoi27: yeah - for instance, the opensea validate API hits the gateway hard (every time someone lists or mints an NFT on their website they use the gateway to grab the metadata). i'm sure they probably already rotate IPs. no opinions on 200/minute - if it's in the standard ballpark it works for me!

packages/gateway/README.md
alanshaw: Screenshot 2022-02-08 at 15 42 08

Should be able to do in the same way we do this for the legacy `/api/*` route.

packages/gateway/README.md
```diff
@@ -39,6 +39,25 @@ You only need to `npm start` for subsequent runs. PR your env config to the `wra
...
+## Usage
+
+nft.storage Gateway provides subdomain style resolutions maintaining compliance with the [same-origin policy](https://en.wikipedia.org/wiki/Same-origin_policy). The canonical form of access `https://{CID}.ipfs.nft.storage/{optional path to resource}` causes the browser to interpret each returned file as being from a different origin.
+
+```
+curl https://bafkreidyeivj7adnnac6ljvzj2e3rd5xdw3revw4da7mx2ckrstapoupoq.ipfs.nft.storage
+Hello nft.storage! 😎
+```
+
+Please note that subdomain resolution is only supported with [CIDv1](https://docs.ipfs.io/concepts/content-addressing/#identifier-formats) in case-insensitive encoding such as Base32 or Base36. For CIDv0 support, you can use the Path style resolution using the nft.storage API as follows:
```
alanshaw, line 51:
```suggestion
Please note that subdomain resolution is only supported with [CIDv1](https://docs.ipfs.io/concepts/content-addressing/#identifier-formats) in case-insensitive encoding such as Base32 or Base36. For CIDv0 support, you can use the Path style resolution as follows:
```

alanshaw (approved):

PR #1309, opened 2022-02-09
fix:add redirect rule for docs

dashcraft:
I added a redirect rule for the root of docs, the theory is that the server side router isn't picking up the index.md file when in production. This is only repro-able on staging or production, not locally.


packages/website/next.config.js
```diff
@@ -20,6 +20,25 @@ const nextConfig = withBundleAnalyzer({
   exportPathMap: async function () {
     return {
       '/ipfs-404.html': { page: '/404' },
+      '/docs/': { page: '/docs/quickstart', statusCode: 301 },
+      '/docs/why-nft-storage': { page: '/docs/why-nft-storage' },
+      '/docs/troubleshooting': { page: '/docs/troubleshooting' },
+      '/docs/quickstart': { page: '/docs/quickstart' },
+      '/docs/concepts/decentralized-storage': {
```
Codigo-Fuentes, line 27: **suggestion (non-blocking):** We could consider adding redirects from '/docs/concepts/' to '/docs/concepts/decentralized-storage' and similar for '/how-to' and '/client' to their first sub-topic as was done for '/docs/' to '/docs/quickstart

**rationale:** https://ff561a2e.nft-storage-1at.pages.dev/docs/concepts/ e.g., currently leads to the same broken state that navigating directly to /docs/ previously produced.

Codigo-Fuentes (approved):

PR #1311, opened 2022-02-09
fix: gateway metrics route with cached response

vasco-santos:
Add cache to metrics as we don't need to have requests to the Durable Objects in each request from Prometheus


alanshaw (approved):
LGTM

PR #1278, opened 2022-02-07
Fix/issue 1174 add tooltips table headers

cmunns:
Add table header tooltips for #1174 
Screen Shot 2022-02-07 at 11 46 34 AM

PR #1323, opened 2022-02-09
fix: fix /files table style issues

### LINKED ISSUE #1319: Minor style issues with /files table
- [x]  There is a double-border around the tooltip info box
- [x]  The question mark is not center aligned in "Storage Providers" column
tooltip-border
non-aligned-question-mark

Codigo-Fuentes:
Closes #1319 

fixed-border
aligned-mark


cmunns (commented):
Can we add these into #1278 . The question icon is already addressed there and the tooltip styles should retain the border-radius: 0 styles that are thicker. The second rounded border should be the only thing that has to get removed.

Issue #1313, opened 2022-02-09
Website page does not grow to fill the min height of the viewport


vasco-santos:
This has not been super problematic because most pages are long, but ToS has this issue

<img width="1760" alt="image" src="https://user-images.githubusercontent.com/7295071/153171062-658769ec-f5ad-4240-aa5f-a5eadcf99778.png">

PR #1338, opened 2022-02-10
fix: update layout component to remove potential space below footer (#1313)

e-schneid:


cmunns (approved):
This is a good solution, I decided to apply it to other pages as well instead of relying on 100vh

PR #1341, opened 2022-02-10
fix: mobile layout break faq page

### LINKED ISSUE #1340: FAQ Page Mobile Layout Break
Overflow issue on mobile FAQ Page

IMG_7665

Codigo-Fuentes:
Closes #1340


dashcraft (approved):
Approved: G2G

feedback:
Looks good, if it gets bigger than that, I think we should just create a class for it. Inline styles are fine for this!

PR #1339, opened 2022-02-10
Fixed broken link

dashcraft:
Fixes the broken link for javascript client lib, inside of quickstart.

PR #1342, opened 2022-02-11
Revert "Fixed broken link"

Codigo-Fuentes:
Revert, then new pr with correct format so it goes into website release as a fix

Reverts nftstorage/nft.storage#1339


dashcraft (approved):
revert

PR #1344, opened 2022-02-11
fix: gateway requested with custom header requested from

vasco-santos:


packages/gateway/src/gateway.js
```diff
@@ -236,6 +236,7 @@ function getHeaders(request) {
     'X-Forwarded-For': `${request.headers.get(
       'cf-connecting-ip'
     )}${existingProxies}`,
+    'X-Requested-From': 'nft.storage',
```
alanshaw, line 239: Perhaps we should use [`X-Forwarded-Host`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host)? WDYT?

vasco-santos, line 239:
```suggestion
    'X-Forwarded-Host': 'nft.storage',
```

alanshaw (approved):
TY

PR #1366, opened 2022-02-11
Blog post about Base URL APIs

jchris:
It would be great to have feedback on the content. It's a small idea so I wrote it quickly.


packages/website/posts/2022-02-15-base-url-apis.mdx
```diff
@@ -0,0 +1,24 @@
...
+
+It's little affordances like this which can make a big difference in the industry. If the easiest way to accept data from users is via a base URL, that is a thumb on the scale for user's data ownership, interoperability and reuse. Sites that accept a URL of uploads benefit from accepting similar data formats as competitive sites. By reusing the upload format, developers not only save time, but can utilize open source data handling libraries, etc. All this means developers can spend more time working on features that distinguish their app.
+
+All this has been possible for years. What’s new today is that with content addressable systems like IPFS you don’t have trust a third party, or control a data repository in the cloud, to upload verifiable data to the network and share the link with service providers. IPFS URLs make up a significant fraction of the assets and metadata in today’s NFTs, as the market sees the value in the verifiable storage that CIDs provide. Content-based identifiers are locked to their content, so any on-chain references can be cryptographically trusted. This makes a shared upload URL as good as sending the files directly.
```
olizilla, line 18:
```suggestion
All this has been possible for years. What’s new today is that with content addressable systems like IPFS you don’t have to trust a third party, or control a data repository in the cloud, to upload verifiable data to the network and share the link with service providers. IPFS URLs make up a significant fraction of the assets and metadata in today’s NFTs, as the market sees the value in the verifiable storage that CIDs provide. Content-based identifiers are locked to their content, so any on-chain references can be cryptographically trusted. This makes a shared upload URL as good as sending the files directly.
```

olizilla (approved):
based apis. I like it.

We could fix it so our blog posts are base URL friendly by co-locating our thumbnail image with the post in the same directory! 

The nft.storage site is not currently available on ipfs, so it's not an issue that would impact anyone today, but we should, seperate of this PR, fix that as well!

packages/website/posts/2022-02-15-base-url-apis.mdx
olizilla: plz to fix this wordo

olizilla:
👁️‍🗨️ 👁️‍🗨️

packages/website/posts/2022-02-15-base-url-apis.mdx
olizilla: @jchris plz. yah killing me.

PR #1371, opened 2022-02-14
chore: fix client-gen script and openapi schema

### LINKED ISSUE #1370: unable to generate client libs
When looking into https://github.com/nftstorage/python-client/issues/3, I realized that the current python client only has a `store` method, not `upload`, so it likely hasn't been generated in a while.

I noticed a few issues with the schema when trying to regenerate:

- The `client-gen.sh` script is slightly broken ATM - it's looking for `/website` instead of `/packages/website`
- The `store` operation in the schema is missing the `responses` block, so the openapi generator fails

I'll make a PR in a sec to fix.

yusefnapora:
This fixes the path to the schema file in `client-gen.sh`, and copies the `responses` section in the schema from the `upload` operation to `store`, to get rid of these errors:

```
Exception in thread "main" org.openapitools.codegen.SpecValidationException: There were issues with the specification. The option can be disabled via validateSpec (Maven/Gradle) or --skip-validate-spec (CLI).
 | Error count: 1, Warning count: 1
Errors: 
        -attribute paths.'/store'(post).responses is missing
Warnings: 
        -attribute paths.'/store'(post).responses is missing
```

closes #1370

### DIFF
--- a/client-gen.sh
+++ b/client-gen.sh
@@ -5,7 +5,7 @@
 
 docker run --rm \
   -v ${PWD}:/local openapitools/openapi-generator-cli generate \
-  -i /local/website/public/schema.yml \
+  -i /local/packages/website/public/schema.yml \
   -g ruby \
   --api-name-suffix=API \
   --git-user-id=nftstorage \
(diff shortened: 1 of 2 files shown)


vasco-santos (approved):
LGTM

PR #1373, opened 2022-02-14
fix: update ipfs car and client

hugomrdias:


vasco-santos (commented):
This looks good. We just need to figure out the CF Pages build issue with types

vasco-santos (approved):

PR #1384, opened 2022-02-15
Adding ACCOUNT_ENABLED user_tag_type to support blocking/unblocking a user

trigramdev9:
* Also adding a reason field on the user_tag table.

### DIFF
--- a/packages/api/db/tables.sql
+++ b/packages/api/db/tables.sql
@@ -11,6 +11,7 @@ CREATE TYPE auth_key_blocked_status_type AS ENUM (
 -- in the application.
 CREATE TYPE user_tag_type AS ENUM
 (
+  'ACCOUNT_ENABLED',
   'PSA_ENABLED',
   'STORAGE_LIMIT_BYTES'
 );
@@ -79,6 +80,7 @@ CREATE TABLE IF NOT EXISTS public.user
     inserted_at    TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL,
     updated_at     TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
 );
+CREATE INDEX IF NOT EXISTS user_updated_at_idx ON public.user (updated_at);
 
 CREATE TABLE IF NOT EXISTS public.user_tag
 (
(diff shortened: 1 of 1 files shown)


packages/api/db/tables.sql
```diff
@@ -87,6 +88,7 @@ CREATE TABLE IF NOT EXISTS public.user_tag
   tag             user_tag_type                                                 NOT NULL,
   value           TEXT                                                          NOT NULL,
   value_type      user_tag_value_type                                           NOT NULL,
+  reason          TEXT                                                          NOT NULL,
   inserted_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now())     NOT NULL,
   deleted_at  TIMESTAMP WITH TIME ZONE
 );
```
vasco-santos, line 94: I know is not related to this PR, but can we add a `UNIQUE (user_id, tag)` ?

trigramdev9: done!  great feedback!

vasco-santos: We can't do this because of the `deleted_at`. We need:

```
CREATE UNIQUE INDEX name_1_idx ON user_tag (user_id, tag, deleted_at)
WHERE deleted_at IS NOT NULL;

CREATE UNIQUE INDEX name_2_ix ON user_tag (user_id, tag)
WHERE deleted_at IS NULL;
```

Please rename the indexes per the name convention we use in the others

vasco-santos: apologies, I forgot the `deleted_at` 🤦🏼

packages/api/db/tables.sql
```diff
@@ -87,11 +89,14 @@ CREATE TABLE IF NOT EXISTS public.user_tag
...
-
-CREATE INDEX IF NOT EXISTS user_updated_at_idx ON public.user (updated_at);
+CREATE UNIQUE INDEX user_tag_is_deleted_idx ON user_tag (user_id, tag, deleted_at)
```
vasco-santos, line 96:
```suggestion
CREATE UNIQUE INDEX IF NOT EXISTS user_tag_is_deleted_idx ON user_tag (user_id, tag, deleted_at)
```

Otherwise multiple runs will create issues

packages/api/db/tables.sql
```diff
@@ -87,11 +89,14 @@ CREATE TABLE IF NOT EXISTS public.user_tag
...
+WHERE deleted_at IS NOT NULL;
+CREATE UNIQUE INDEX user_tag_is_not_deleted_idx ON user_tag (user_id, tag)
```
vasco-santos, line 98:
```suggestion
CREATE UNIQUE INDEX IF NOT EXISTS user_tag_is_not_deleted_idx ON user_tag (user_id, tag)
```

vasco-santos (changes requested):
one more detail in suggestions

vasco-santos (approved):
LGTM

Issue #1332, opened 2022-02-10
Gateway tracking wether requested content is in Database


vasco-santos:
We want to know if gateway requested CIDs are root CIDs stored in Content table (and also if they are Pinned).

Requirements:
- Keep state with counter of:
  - requested CIDs stored
  - requested CIDs pinned
  - requested CIDs pinQueued
  - requested CIDs not stored

vasco-santos:
@dchoi27 let me know if you have other thoughts/ideas of things we should look into in this context.

Probably a special case if we fail to request but content is in the DB?
Or track some kind of relationship on how "old" is the content that is being requested?  Maybe an histogram with like 0.5h, 1h, 2h, 4h, 12h, 24h, 3 days, ... + Inf

dchoi27:
Yes for sure how old the content is (when it was requested vs. when it was first uploaded)
Could you tell me more about "if we fail to request but content is in the DB"? Like if a user requests data we have but we can't fetch it?

Can we track the metrics around the response for each of the groups above? E.g. if it's pinQueued, does it take longer / less reliable to fetch?

vasco-santos:
> Could you tell me more about "if we fail to request but content is in the DB"? Like if a user requests data we have but we can't fetch it?

Yes, so this would be targeting the incomplete uploads.

> Can we track the metrics around the response for each of the groups above? E.g. if it's pinQueued, does it take longer / less reliable to fetch?

Yes, that's a good idea

dchoi27:
Awesome, SGTM

JeffLowe:
This sounds very similar to the needs and plans we have for niftysave (discussed as recently as today with @mikeal ).  I'm pulling in @the-simian here.  You two may sync up on roadmap to implement this to meet both needs.

olizilla:
@dchoi27 how important are these stats to us? In order to make this work #1386 adds logic to hit the nftstorage db for every single CID that is requested from the gateway. That seems like an amplification point where a spike in traffic to the gateway cause a spike in requests to the nftstorage db...  two systems that are currently isolated from each other become co-dependent.

in the worst case, a sustainable increase in gateway trafffic could be an unsustainable increase in nftstorage db reads... we can and will continue to optimse and grow that db, but I'd feel more comfortable if we ditched these metrics and kept the gateways sparate from the nft.storage api

Also notable adding these stats makes the current gateway impl less reusable / in need of more customisation to be used as a web3.storage gateway.

[...]

PR #1386, opened 2022-02-16
feat: gateway tracking requested cids in database

vasco-santos:
This PR adds metrics tracking whether content was stored in NFT.storage and its Pinning status characteristics. It uses nft.storage client to check the state of the content. We could go through the route of directly connecting to the DB here, but as we do this in the waitUntil it seems cleaner to rely on the client (specially as we can't access replica directly).

Prometheus metrics to be consumed are as follows:
```
# HELP nftgateway_errored_responses_with_known_content_total errored requests with known content in NFT.storage.
# TYPE nftgateway_errored_responses_with_known_content_total counter
nftgateway_errored_responses_with_known_content_total{env="test"} 0
# HELP nftgateway_responses_by_content_status_total total of responses by content status. Either stored or not stored.
# TYPE nftgateway_responses_by_content_status_total counter
nftgateway_responses_by_content_status_total{env="test",status="Stored"} 0
nftgateway_responses_by_content_status_total{env="test",status="NotStored"} 0
# HELP nftgateway_responses_by_pin_status_total total of responses by pin status.
# TYPE nftgateway_responses_by_pin_status_total counter
nftgateway_responses_by_pin_status_total{env="test",status="PinError"} 0
nftgateway_responses_by_pin_status_total{env="test",status="Pinned"} 0
nftgateway_responses_by_pin_status_total{env="test",status="Pinning"} 0
nftgateway_responses_by_pin_status_total{env="test",status="PinQueued"} 0
# HELP nftgateway_responses_per_time_by_status_total total of responses by status per response time bucket
# TYPE nftgateway_responses_per_time_by_status_total histogram
nftgateway_responses_per_time_by_status_total{status="PinError",le="0.05",env="test"} 0
nftgateway_responses_per_time_by_status_total{status="Pinned",le="0.05",env="test"} 1
nftgateway_responses_per_time_by_status_total{status="Pinning",le="0.05",env="test"} 0
nftgateway_responses_per_time_by_status_total{status="PinQueued",le="0.05",env="test"} 0
nftgateway_responses_per_time_by_status_total{status="PinError",le="0.1",env="test"} 0
nftgateway_responses_per_time_by_status_total{status="Pinned",le="0.1",env="test"} 2
nftgateway_responses_per_time_by_status_total{status="Pinning",le="0.1",env="test"} 0
nftgateway_responses_per_time_by_status_total{status="PinQueued",le="0.1",env="test"} 0
(...)
```

Follow up work (will create issue):
- improve testing by spinning up API locally with Miniflare

Should use #1391 cache in check endpoint

Closes https://github.com/nftstorage/nft.storage/issues/1332

### DIFF
--- a/packages/gateway/src/durable-objects/summary-metrics.js
+++ b/packages/gateway/src/durable-objects/summary-metrics.js
@@ -1,22 +1,33 @@
+import { NFTStorage } from 'nft.storage'
+import { PinStatusMap } from '../utils/pins.js'
 import {
   responseTimeHistogram,
   createResponseTimeHistogramObject,
 } from '../utils/histogram.js'
 
 /**
+ * @typedef {'Stored'|'NotStored'} ContentStatus
+ * @typedef {import('../utils/pins').PinStatus} PinStatus
+ *
  * @typedef {Object} SummaryMetrics
  * @property {number} totalWinnerResponseTime total response time of the requests
  * @property {number} totalWinnerSuccessfulRequests total number of successful requests
  * @property {number} totalCachedResponseTime total response time to forward cached responses
  * @property {number} totalCachedResponses total number of cached responses
+ * @property {number} totalErroredResponsesWithKnownContent total responses errored with content in NFT.storage
  * @property {BigInt} totalContentLengthBytes total content length of responses
  * @property {BigInt} totalCachedContentLengthBytes total content length of cached responses
+ * @property {Record<ContentStatus, number>} totalResponsesByContentStatus
+ * @property {Record<PinStatus, number>} totalResponsesByPinStatus
  * @property {Record<string, number>} contentLengthHistogram
  * @property {Record<string, number>} responseTimeHistogram
+ * @property {Record<PinStatus, Record<string, number>>} responseTimeHistogramByPinStatus
  *
  * @typedef {Object} FetchStats
- * @property {number} responseTime number of milliseconds to get response
- * @property {number} contentLength content length header content
+ * @property {string} cid fetched CID
+ * @property {boolean} errored fetched CID request errored
+ * @property {number} [responseTime] number of milliseconds to get response
+ * @property {number} [contentLength] content length header content
  */
 
 // Key to track total time for winner gateway to respond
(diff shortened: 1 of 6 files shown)


packages/gateway/src/durable-objects/summary-metrics.js
```diff
@@ -35,13 +49,22 @@ const TOTAL_CACHED_CONTENT_LENGTH_BYTES_ID = 'totalCachedContentLengthBytes'
...
+const TOTAL_RESPONSES_BY_PIN_STATUS_ID = 'totalResponsesByPinStatus'
 
 /**
  * Durable Object for keeping summary metrics of gateway.nft.storage
  */
 export class SummaryMetrics0 {
   constructor(state) {
     this.state = state
+    // @ts-ignore we don't need token just for check
+    this.nftStorageClient = new NFTStorage({})
```
vasco-santos, line 67: For just check, we don't need a token, while it is "required" in NFTstorage (but param called option)

@hugomrdias should we make it optional in the types?

hugomrdias: yes we should, but here if you just use it for check call might be better to not include the client.

vasco-santos: you mean,  just HTTP the endpoint?

olizilla: yep, I agree, if we just want to hit the public check endpoint, maybe just add a helper function around a fetch

olizilla:
notable that this PR is the only thing that ties the gateway impl to nftstorage. It would be neat to have a generic "racing gateway proxy" that anyone could deploy, and then have some mechinism for adding extra metrics for a given deployment.

We might want to reuse this for a web3.storage gateway

vasco-santos:
> notable that this PR is the only thing that ties the gateway impl to nftstorage. It would be neat to have a generic "racing gateway proxy" that anyone could deploy, and then have some mechinism for adding extra metrics for a given deployment.

We can have a check endpoint ENV Var to have this generic

vasco-santos:
Closing this and we will track on issue https://github.com/nftstorage/nft.storage/issues/1332 how to gather this

PR #1390, opened 2022-02-16
fix: add issue template and discord link to docs

cmunns:
Screen Shot 2022-02-16 at 2 22 03 PM


dashcraft (approved):
LGTM! 🚢

Codigo-Fuentes (commented):
**question**: The idea is that the direct link to the discord channel only works if you already have access to the discord? So we are changing the link to be an invite?
**suggestion**: If that is the case, then should we also update the main readme which currently links to the channel?

cmunns:
@JeffLowe shall we make a new issue for @Codigo-Fuentes suggestion here?

PR #1400, opened 2022-02-18
feat: add stats routes

dashcraft:
Working on stats routes, these are currently written while not having a local setup to test against.
I'll be testing over the next few days.


dashcraft:
This implements the materialized view implementation i discussed with @dchoi27. This _should_ be good enough for 10's of millions of records. Locally, on a postgres instance with 15 million records, it takes roughly 10-13 seconds, sometimes closer to 6. This is _much_ easier to refresh since it is all postgres side, but it is not the _fastest_ solution. This solution is exact though, so numbers are trustworthy.

As noted in my conversation with @dchoi27, the more robust solution should this one be rendered too slow or not robust enough, would be to cash the previous values and programmatically calculate the flow values from the 2 diffed values (previous count and count in past 7 days, added together as current total count).

dashcraft:
Fwiw on a timeseries database i have, this query takes 1.8s on 10 million records, so this is indicative that indexing plays a role in this query speed.

packages/api/db/cargo.testing.sql
```diff
@@ -50,3 +50,31 @@ INSERT INTO public."user" (magic_link_id, github_id, name, email, public_address
...
 
+
+-- Count is the most accurate representation of data
+-- Creating a materialized views gives us cacheing as well as accuracy for percent difference
+
+CREATE MATERIALIZED VIEW upload_stats AS
+	SELECT
+    (
+        SELECT
+        CAST("count"(*) AS float )
```
alanshaw, line 61: Why cast as float? - it will never have a fractional component.

packages/api/db/cargo.testing.sql
```diff
@@ -50,3 +50,31 @@ INSERT INTO public."user" (magic_link_id, github_id, name, email, public_address
 
 INSERT INTO public.auth_key (name, secret, user_id) VALUES ('main', 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJkaWQ6ZXRocjoweDY1MDA3QTczOWFiN0FDNWM1MzcxNjEyNDliODEyNTBFNDllMjg1M0MiLCJpc3MiOiJuZnQtc3RvcmFnZSIsImlhdCI6MTYzOTc1NDczNjYzOCwibmFtZSI6Im1haW4ifQ.wKwJIRXXHsgwVp8mOQp6r3_F4Lz5lnoAkgVP8wqwA_Y', 1);
 
+
+-- Count is the most accurate representation of data
+-- Creating a materialized views gives us cacheing as well as accuracy for percent difference
+
+CREATE MATERIALIZED VIEW upload_stats AS
```
alanshaw, line 57: How does this get updated?

packages/api/db/cargo.testing.sql
```diff
@@ -50,3 +50,31 @@ INSERT INTO public."user" (magic_link_id, github_id, name, email, public_address
...
+        SELECT
+        CAST("count"(*) AS float )
+        FROM upload
+    ) AS total_uploads
+	, 
+	(
+        SELECT 
+        CAST("count"(*) AS float )
+        FROM upload
+        WHERE inserted_at < CURRENT_DATE - 7
```
alanshaw, line 69: `CURRENT_DATE - 7` - does that definitely work as we expect...?

packages/api/db/tables.sql
```diff
@@ -216,4 +216,4 @@ CREATE TABLE IF NOT EXISTS backup
     UNIQUE (upload_id, url)
 );
 
-CREATE INDEX IF NOT EXISTS backup_upload_id_idx ON backup (upload_id);
+CREATE INDEX IF NOT EXISTS backup_upload_id_idx ON backup (upload_id);
```
alanshaw, line 219: No need for this change?

packages/api/pw-test.config.cjs
```diff
@@ -53,7 +53,7 @@ module.exports = {
     await execa(cli, ['db', '--start'])
     console.log('⚡️ Cluster and Postgres started.')
 
-    await execa(cli, ['db-sql', '--cargo', '--testing', '--reset'])
+    await execa(cli, ['db-sql', '--cargo', '--testing'])
```
alanshaw, line 56: What's the reason for this change please?

packages/api/src/routes/metrics.js
```diff
@@ -8,6 +9,19 @@ export async function metrics(_, { db }) {
...
+/** @type {import('../bindings').Handler} */
+export async function getUploadStats(_, { db }) {
+  let stats = await db.getUploadStats()
+
+  return new JSONResponse(
+    {
+      ok: true,
+      data: stats,
+    },
+    { status: 202 }
```
alanshaw, line 21: Why 202 not 200?

packages/api/src/utils/db-types.d.ts
```diff
@@ -857,6 +869,16 @@ export interface definitions {
...
   }
+  upload_stats: {
+    /**
+     * Format: bigint
+     * @description Note:
+     * This is a calculated value on a materialized view that represents a whole number percent
+     */
+    total_uploads: number
+    total_deals: number
+    growth_rate_percent: number
```
alanshaw, line 880: Where is this used?

packages/cron/src/jobs/metrics.js
```diff
@@ -85,6 +91,25 @@ async function updateUploadsCount(roPg, rwPg, type) {
...
 
+/**
+ * @param {Client} roPg
+ * @param {Client} rwPg
+ */
+async function updateUploadTotal(roPg, rwPg) {
+  const { rows } = await roPg.query(COUNT_TOTAL_UPLOADS)
+  if (!rows.length)
+    throw new Error(`no rows returned counting past weeks uploads`)
+  await rwPg.query(UPDATE_METRIC, [`uploads_all_types_total`, rows[0].total])
```
alanshaw, line 102: This is not used by the `/stats` route is it?

packages/cron/src/jobs/metrics.js
```diff
@@ -14,6 +14,10 @@ const COUNT_USERS = 'SELECT COUNT(*) AS total FROM public.user'
 
 const COUNT_UPLOADS = 'SELECT COUNT(*) AS total FROM upload WHERE type = $1'
 
+const COUNT_TOTAL_UPLOADS = 'SELECT COUNT(*) AS total FROM upload'
+// This is a rolling 7 day date window of total uploads
+const REFRESH_UPLOAD_STATS = 'REFRESH MATERIALIZED VIEW upload_stats;'
```
alanshaw, line 19: I think this needs `CONCURRENTLY` so that requests are not blocked while it is being updated.

packages/cron/src/jobs/metrics.js
```diff
@@ -85,6 +91,25 @@ async function updateUploadsCount(roPg, rwPg, type) {
...
+    throw new Error(`no rows returned counting past weeks uploads`)
+  await rwPg.query(UPDATE_METRIC, [`uploads_all_types_total`, rows[0].total])
+}
+
+/**
+ * @param {Client} roPg
+ * @param {Client} rwPg
+ */
+async function updateUploadStats(roPg, rwPg) {
+  await roPg.query(REFRESH_UPLOAD_STATS)
```
alanshaw, line 110: I'm not sure about having a separate materialized view here. Why not just calculate each metric concurrently and insert/update it into the metrics table like the rest? What do we gain by having a materialized view?

alanshaw (changes requested):

packages/api/db/cargo.testing.sql
alanshaw: Oh I see in the metrics cron.

packages/api/db/cargo.testing.sql
dashcraft: Previously i was doing a calculation in SQL that resulted in a decimal through division, this is no longer needed so I can remove!

packages/api/db/cargo.testing.sql
dashcraft: As long as it's a date, it should return 7 days minus the current date. If it's not a date, this won't work.
I'll double check this logic.

packages/api/db/tables.sql
dashcraft: Yeah this was a result of me shuffling files around, will remove

packages/api/pw-test.config.cjs
dashcraft: I am refactoring dev, this should've went over to that branch, I'll remove this change.

https://github.com/nftstorage/nft.storage/pull/1485

packages/api/src/routes/metrics.js
dashcraft: Good catch, i'll get this changed!

packages/api/src/utils/db-types.d.ts
dashcraft: This is the returned value from the stats page, growth_rate needs removed, I'll get it updated.

packages/cron/src/jobs/metrics.js
dashcraft: It is not, i'll remove this change, I shuffled some things around, I'll make sure there's nothing there that needs changed.

packages/cron/src/jobs/metrics.js
dashcraft: The logic for the materialized view is all held in SQL, it feels much easier to reason about, change, update and refresh _only_ the stats we return for the website, vs having separate individual metrics that get combined server side. There's no meaningful difference outside of not having to maintain several independent queries.

I'm super open, I've got a handful of alternative solutions, some of them more performant but less accurate, if your think adding them directly the metrics table with no materialized view is the way to go, I can do that.

dashcraft: I rewrote this to use the existing metrics table. We don't have to count the total again, if we already query each upload type individually, I can just sum that amount as the total and diff the past 7 days on the frontend.

I removed the materialized view and brought all the additional metrics queries into the metrics table + cron

alanshaw (changes requested):
This LGTM, just a minor request to follow the existing naming convention for the metric names.

Also, in main we're now [timing the queries](https://github.com/nftstorage/nft.storage/blob/3daefd54120c937084b5400a7ca3191005b8b6e6/packages/cron/src/jobs/metrics.js#L39) so we can debug more easily as the data size grows - can you please rebase/merge with main and add the `withTimeLog` function call around your queries? 🙏

packages/api/test/stats-show.spec.js
```diff
@@ -0,0 +1,34 @@
...
+    const getStats = await fetch('stats')
+    const statsResults = await getStats.json()
+    assert.deepStrictEqual(
+      statsResults,
+      {
+        ok: true,
+        data: {
+          total_uploads_past_7: 0,
+          total_deals: 1,
+          total_deal_size: 25515304172,
```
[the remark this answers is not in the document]
dashcraft: Will do!

alanshaw (approved):

Issue #1183, opened 2022-01-28
Augment gateway metrics collected


vasco-santos:
@alanshaw added a few suggestions 

:bar_chart: :bulb: Additional metrics to collect from the NFT gateway
- [ ] Histogram of file sizes (tracked in https://github.com/nftstorage/nft.storage/issues/1181)
(MB) <0.5 | <1 | <2 | <5 | <25 | <50 | <100 | <500 | <1,000 | <5,000 ...
...all the way up to >32,000
- [ ] Histogram of requests by CID version
- [ ] Requests per IPLD codec
- [ ] Requests by multihash hashing function
- [ ] Histogram of requests by file type:
  - image
  - video
  - text
  - html
  - script
  - directory
  - other (we should log these so we can add buckets over time)
- [ ] Direct CID requests vs CID+path
- [ ] Cache hits vs misses
- [ ] IPFS vs IPNS requests...also prerequisite IPNS support!? :laughing:

PR #1429, opened 2022-02-21
feat: augment metrics with cid and cid+path counters

vasco-santos:
As part of https://github.com/nftstorage/nft.storage/issues/1183 this PR augments metrics with query type counter. Prometheus metrics look as follows:

```

# HELP nftgateway_responses_by_query_type_total total of responses by query status. Either CID or CID+PATH.
# TYPE nftgateway_responses_by_query_type_total counter
nftgateway_responses_by_query_type_total{env="test",type="CID"} 1
nftgateway_responses_by_query_type_total{env="test",type="CID+PATH"} 1
```

### DIFF
--- a/packages/gateway/src/durable-objects/summary-metrics.js
+++ b/packages/gateway/src/durable-objects/summary-metrics.js
@@ -4,6 +4,8 @@ import {
 } from '../utils/histogram.js'
 
 /**
+ * @typedef {'CID'|'CID+PATH'} QueryType
+ *
  * @typedef {Object} SummaryMetrics
  * @property {number} totalWinnerResponseTime total response time of the requests
  * @property {number} totalWinnerSuccessfulRequests total number of successful requests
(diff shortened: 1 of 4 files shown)


vasco-santos:
Any suggestions for better naming for query type would be appreciated. Not super fan of it

olizilla:
📎 It looks like we are trying to use prometheus to do the job of http access logs We should explore https://www.cloudflare.com/en-gb/products/cloudflare-logs/ and a http access log summary tool.

vasco-santos:
> 📎 It looks like we are trying to use prometheus to do the job of http access logs

Thanks for the feedback. I actually understood that I could get part of this metrics with Cloudflare Analytics. Currently using Cloudflare Analytics to obtain content type using https://github.com/nftstorage/cloudflare-analytics-prometheus-exporter/ and this PR will simply focus on query type.

We already have a storage service for logs setup with Logtail and all the Logs go there. Sadly, they don't seem to offer any way of listening to the logtail source and parse inputs so that we can track metrics with a different app that exposes metrics to our Prometheus. They actually have a Grafana dashboard within Logtail, but we need to interact with the logs with SQL and it is quite complex to graph there rates of requests (for subdomain only) that are either CID or CID+Path format. Also, I would prefer to have everything in the same Grafana dashboard and just tracking here seems better than maintaining other service anyway.

packages/gateway/src/durable-objects/summary-metrics.js
```diff
@@ -188,57 +201,75 @@ export class SummaryMetrics0 {
...
+      this.responseTimeHistogram,
+      responseTimeHistogram,
+      stats.responseTime
     )
+  }
+}
 
-    histogramCandidates.forEach((candidate) => {
-      tmpHistogram[candidate] += 1
+function getUpdatedHistogram(histogramData, histogramBuckets, value) {
```
vasco-santos, line 243: Apologies for adding this refactor here 😬  I did the code because this PR already had MimeTypes histogram, but now I think we should keep it as it simplifies current histograms and if we add more in the future

PR #1430, opened 2022-02-21
fix: link to main branch from generated client docs

### LINKED ISSUE #1299: Links in autogenerated JS docs link don't link to the main branch
For instance: https://nftstorage.github.io/nft.storage/client/classes/lib.NFTStorage.html#storeCar-1
This link: 

Not sure if this is actually a bug, flagging in case it is

yusefnapora:
This sets the `gitRevision` typedoc option to `main` so that links to the source always target the main branch instead of the current git hash.

I also moved the typedoc options to a `typedoc.json` file, since they were getting kind of hard to maintain as cli args in the package.json script entry.

Closes #1299


vasco-santos (approved):
LGTM 👌🏼

PR #1446, opened 2022-02-22
fix: nextra scolling inline style

cmunns:
Removes the top offset in nextra fixed position. checked in Chrome, FF, Safari

Screen Shot 2022-02-22 at 10 50 49 AM

Screen Shot 2022-02-22 at 10 50 56 AM


dashcraft (approved):
lgtm

PR #1450, opened 2022-02-22
feat(docs): add overview page for js client & links to generated clients

yusefnapora:
This replaces the docs in the "Client" section of the nav with a page for the JS client and one for the OpenAPI generated clients. The JS page is an overview of installing the client, constructing an instance, and the various store methods available, plus a mention of `delete`. It should probably also include `check` and so on, but we can add that in a future PR.

Screen Shot 2022-02-22 at 5 30 08 PM


packages/website/pages/docs/client/generated.md
```diff
@@ -0,0 +1,19 @@
+# OpenAPI generated clients
+
+If you're integrating NFT.Storage with a JavaScript project, the [JavaScript client][reference-js-client] is a natural fit. If JavaScript isn't part of your stack, the NFT.Storage [HTTP API][reference-http-api] can be used from any programming language capable of making HTTP requests. 
+
+In addition to using the standard request libraries for your language, you can use client libraries generated from the [OpenAPI](https://www.openapis.org/) schema that describes the API.
```
vasco-santos, line 5: Should we point out here that JS client has a custom logic for chunking uploads and that generated clients will be limited to 100Mb uploads?

yusefnapora: Yeah, that's a good call out 👍

JeffLowe:
Does this negate the need for #1256 ?

yusefnapora:
@JeffLowe I think so, yes. Just closed that one, thanks for reminding me. I'll rebase this and fix conflicts in a sec

dashcraft (approved):
lgtm!

PR #1462, opened 2022-02-23
fix: database credentials for startup

dashcraft:
Updated docs and seed values to fit current user and current setup.

Linked to #1464 as part of solution (will not close the whole thing, but a nice quick win improvement)


JeffLowe:
@olizilla @vasco-santos @alanshaw could one of you give a quick review on this small improvement?  This originated from discussion between @dashcraft and @alanshaw

redaphid (approved):

Issue #1470, opened 2022-02-24
Add Tatum logo to website


dchoi27:
- [ ] add to homepage and /stats page
![Tatum_Logo (1)](https://user-images.githubusercontent.com/11778450/155575793-bde73d80-dfef-4cde-bdf4-207b860134d8.svg)

JeffLowe:
Will be closed by #1471

PR #1471, opened 2022-02-24
feat: add project galaxy logo

cmunns:
reworked grid spacing to accommodate new logo
Screen Shot 2022-02-24 at 9 49 12 AM
Screen Shot 2022-02-24 at 9 49 03 AM


cmunns:
@JeffLowe reverted the grid change and applied the logo swaps which includes Tatum #1470 

Screen Shot 2022-02-25 at 10 57 35 AM
.

Codigo-Fuentes:
**suggestion (non-blocking):** Make widowed items centered (when there are one or two items on the last row)

PR #1483, opened 2022-02-25
fix: Api developer experience improvements

dashcraft:
Updated dev experience to be more pleasant, doesn't require docker restart or database nuking when running dev

Doesn't require restarting or stopping docker, doesn't require nuking the database when stopping terminal.


redaphid (approved):
Looks good to me :)

PR #1485, opened 2022-02-28
fix: dev experience refactor

dashcraft:
This is a massive overhaul I'm experimenting with.

Issues:
Right now testing blows away the development DB.
Restarting the CLI blows away the development DB.
Localhost runs API interface on port 3000 which conflicts with a lot of other services, best to choose something different for dev purposes that isn't easy to guess.

Ideally, the dev DB cli is only used for initial setup, migrations, and testing, I don't think it should be tapped into as part of the normal dev process for the API. The DB situation unnecessarily increases dev time and is fairly inefficient, with docker-compose I'm working on standing up a test db that can be routinely blown away without affecting the other systems.


hugomrdias:
If you need to keep a DB around that the tests dont destroy you can just pass --project "some-string-you-like"

For the tests i think we should keep spawning a clean db.

Regarding the ports we can change those, just curious whats conflicting with 3000 ?

PR #1486, opened 2022-02-28
fix: metrics cron

alanshaw:
1. Switches to using `pg.Pool` so multiple SQL commands can be run concurrently!
1. Adds some logging to the metrics cron job to make debugging easier.
1. Increase the timeout by 5 mins!


vasco-santos (approved):
LGTM

PR #1487, opened 2022-02-28
fix: gateway sentry source maps

vasco-santos:
Copying  @olizilla 's lovely work from web3 https://github.com/web3-storage/web3.storage/pull/1024 and https://github.com/web3-storage/web3.storage/pull/1033


olizilla (approved):
👍

PR #1495, opened 2022-03-01
feat: authentication persistence refactor

dashcraft:
Same pr, reopened


redaphid (approved):

hugomrdias:
@dashcraft i dont really understand this PR, it lowers the security we had from magic.link to the local storage security (not good) until user logs out. 
The hook we had seems to be basically inlined.

what we should do is update magic and check if they already fixed theirs API to not do the flow KMS cycle in the isLoggedIn method.

PR #1499, opened 2022-03-01
fix: api assert dep

vasco-santos:
When https://github.com/nftstorage/nft.storage/commit/86e3a9c735eb9cc1a50565373b1fb8b92d8fe6d9 was merged, the dependencies of niftysave were uninstalled and api was using assert but was relying on niftysave's dep instead of its own...


alanshaw (approved):
Thanks 🙏

Issue #1502, opened 2022-03-01
Gateway html response directories links resolution


vasco-santos:
Let's consider https://bafybeiegsdpbk5bs73mtsj2cduyaq3hddra5c33t6vw2vcctxhhxmsd46u.ipfs-staging.nftstorage.link/https directory. This returns a directory with a file with a link https://bafybeiegsdpbk5bs73mtsj2cduyaq3hddra5c33t6vw2vcctxhhxmsd46u.ipfs-staging.nftstorage.link/ipfs/bafybeiegsdpbk5bs73mtsj2cduyaq3hddra5c33t6vw2vcctxhhxmsd46u/https/example , which is problematic.

The reason for this is that behind the scenes we request public gateways (ipfs.io, cf-ipfs.com and pinata dedicated gateway) using path resolution. The HTML response of the gateways will have a relative path (as expected) to its links `/ipfs/{rootCid}/path`.

We use path resolution within nftstorage.link gateway worker because Pinata dedicated gateway does not support subdomain resolution. So, it is not an option to change this.

PR #1503, opened 2022-03-01
fix: gateway directories links resolution

vasco-santos:
Please read problem in https://github.com/nftstorage/nft.storage/issues/1502 for better understanding of the issue.

This PR fixes directory links resolution. Given we can't use change the worker to perform subdomain resolution to all gateways in the race, in this PR a simple solution is implemented, where basically we will accept URLs like `https://{cid}.ipfs.nftstorage.link/ipfs/{cid}/path` and ignoring subdomain CID we will redirect to subdomain route `https://nftstorage.link/ipfs/{cid}/path`.

Alternative?
We want to proxy responses from the gateway (which can have large streams of data). However, we could potentially look into content type, and if content type is HTML we could read the HTML content, parse it and modify all  . I did not go this route because it is error prone (there is nothing against calling a directory within a directory `ipfs` for instance) and also we should not guess as it can also be a user website instead of a render of the directory listing.

Closes https://github.com/nftstorage/nft.storage/issues/1502


packages/gateway/src/ipfs.js
```diff
@@ -8,7 +8,7 @@ import { normalizeCid } from './utils/cid.js'
  */
 export async function ipfsGet(request, env) {
   const cid = request.params.cid
-  const path = request.url.split(cid)[1] || ''
+  const path = request.url.split(`/ipfs/${cid}`)[1] || ''
```
alanshaw, line 11: Can we instead be more explicit about what part of the URL we're considering? 

```js
new URL(request.url).pathname.split(cid)[1]
```

There's nothing to say the reason for the longer matcher string is because the CID could appear in the domain of the URL. Arguably using `pathname` is no better...perhaps a comment here could help.

...also what if the CID happens to appear in the path multiple times?

...can we also preserve querystring?

alanshaw (changes requested):

alanshaw (approved):

PR #1520, opened 2022-03-01
fix: client docs markdown fixes

### LINKED ISSUE #1517: Correct markdown issues in /docs
- [ ] correct the Markdown error in the [storeDirectory method] 
- [ ] “The storeCar method”, having “storeCar” be an inline-code-snippet creates a break in the middle of the link because it has padding. 

One fix could be to only hyperlink the first word "storeCar" and not the 2nd word "method"

Codigo-Fuentes:
Closes #1517 

- Adds missing link
- updates format of links so that inline code not mixed with non-code, fixing space in middle of link and improved readability


JeffLowe (approved):
Reviewed visually.  The link issues are resolved.

PR #1526, opened 2022-03-01
feat: Added support for full path external redirects

### LINKED ISSUE #1491: Include more prominent link to http api docs in /docs
There are a few direct links to this in the docs page content throughout.  However, ideally we would include the link in the side nav. 
https://nft.storage/api-docs/

One possible solution is to use a placeholder mds file (to auto-populate the side nav link) that redirects to the api docs.

dashcraft:
We had need for full path redirects in docs pages, this supports that behavior.
Added raw HTTP client redirect

Fixes #1491


dashcraft:
image

the-simian (approved):

PR #1540, opened 2022-03-03
fix: visibility issue conflict with nextra themes

dashcraft:
Utils was updated to have more tailwind style CSS, but nextra expects hidden to be a display property.


the-simian (approved):
visibility:hidden - invisible but takes up its original space
display:none- It is hidden and takes no space

yep, we want thing 2. no space. - approved.

Issue #1195, opened 2022-01-31
Gateway - Rate Limiting improvements with public IPFS gateways


vasco-santos:
# Problem Statement

Public IPFS gateways have rate limiting characteristics that will affect our Cloudflare worker nft.storage Gateway. A CF Worker has a unique outgoing IP, no matter what region it will be running.

When an IPFS gateway is rate limited, we will likely not be able to use a gateway that can already have the content cached. This will cause impact on response time, specially when gateway is blocked for a long period.

Finally, we can get to a state where all our public gateways are rate limited. In this case, we will simply redirect traffic to one gateway.

# Gateways characteristics

## Gloudflare Gateway

- by IP when 100 requests/min
- blocks IP during 1 Hour
    - further requests will not reset the timer, i.e 1H blocked from first block

## Pinata Gateway

- Currently N/A for Dedicated Gateways

## ipfs.io Gateway

- More difficult to predict rate limiting characteristics, considering its infrastructure setup
- ipfs.io public gateway uses several different techniques on load balancing and rate limiting, which makes it difficult to predict the actual rate limit value.
- ipfs.io gateway starts by geo routing requests, followed by load balancing. In our use case, I predict that all requests will be routed to the same geo area.
- The rate limit is set per load balancer. For instance, same {IP_ADDR, URI} are limited at1/second or 15/minute to a particular load balancer. Then there is a global 800/s on each Load Balancer.
- It also uses [bursting techniques](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req) that will start by delaying responses before it actually fails. 

# Scale

- IPFS gateway req/s are over 1000 (metrics from end of 2021)
- TODO: @vasco-santos  to find out data transfer metrics with PL Gateway Team

# Solutions

We will need to work with the teams responsible for each of the gateways mentioned and create plans on how we can improve the rate limiting situation.

Per previous conversations, having improvements on the Cloudflare Worker "client" to for example have multiple output IPs is something that will likely not be possible to achieve.

## Gloudflare Gateway

_TBD_

Multiple options on the table:
- have a different endpoint (with a special token / or some other way of whitelist our worker) that we can have more control on rate limiting
- support “do you have this locally” — we can rely on this with CF cache API. The limitation is the user would need to be within the cf-ipfs gateway region. Looking with CF team on how we could have global cache API support to the Cloudflare IPFS Gateway cache region
- just have bigger limits on Workers

## ipfs.io Gateway

State: ✅ 

[...]

PR #1546, opened 2022-03-03
feat: gateway with dedicated cf gateway

vasco-santos:
This PR adds our dedicated CF gateway as the gateway to be used 🎉 

I had our dedicated CF gateway `cf.nftstorage.link` setup with Cloudflare team, this will be an offering from them in the future and we had a manual process with them to have the DNS setup and proxying to ipfs.cloudflare.com.

With this dedicated gateway we have more freedom and powers with custom rate limits. With this in mind, a set of rules were put in place to avoid abuse and hammering this domain:
- [Firewall rule](https://dash.cloudflare.com/fffa4b4363a7e5250af8357087263b3a/nftstorage.link/firewall/firewall-rules) blocking all traffic matching `cf.nftstorage.link/*` from the outside world (only our worker can access it, same CF zone ID)
- [Rate limit rule](https://dash.cloudflare.com/fffa4b4363a7e5250af8357087263b3a/nftstorage.link/firewall/custom-rules) limiting traffic fo `cf.nftstorage.link/*` per `X-Forwarded-For`
  - wondering if we should use `cf-connecting-ip` instead, given clients can easily manipulate `X-Forwarded-For` to add fake proxies just to circumvent limits. If so, we will need to add a new header to the request to the gateways to use it. Thoughts?

Closes https://github.com/nftstorage/nft.storage/issues/1195

### DIFF
--- a/packages/gateway/src/durable-objects/gateway-rate-limits.js
+++ b/packages/gateway/src/durable-objects/gateway-rate-limits.js
@@ -16,7 +16,7 @@ export class GatewayRateLimits2 {
     this.state = state
     this.id = this.state.id.name
     /** @type {RateLimitCharacteristics} */
-    this.rateLimitCharacteristics = getRateLimitingPreventionCharacteristics(
+    this.rateLimitCharacteristics = getRateLimitConfig(
       this.id
     )
 
@@ -84,7 +84,7 @@ const MINUTE = SECOND * 60
  * @param {string} gatewayUrl
  * @return {RateLimitCharacteristics}
  */
-function getRateLimitingPreventionCharacteristics(gatewayUrl) {
+function getRateLimitConfig(gatewayUrl) {
   switch (gatewayUrl) {
     case 'https://ipfs.io':
       return {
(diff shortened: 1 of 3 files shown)


packages/gateway/src/durable-objects/gateway-rate-limits.js
```diff
@@ -82,16 +84,16 @@ const MINUTE = SECOND * 60
  * @param {string} gatewayUrl
  * @return {RateLimitCharacteristics}
  */
-function getRateLimitingCharacteristics(gatewayUrl) {
+function getRateLimitingPreventionCharacteristics(gatewayUrl) {
```
alanshaw, line 87: I'd settle for `getRateLimitConfig`...

alanshaw (approved):
LGTM

packages/gateway/src/durable-objects/gateway-rate-limits.js
```diff
@@ -11,12 +11,14 @@
...
  * State: number[]
  */
-export class GatewayRateLimits1 {
+export class GatewayRateLimits2 {
   constructor(state) {
     this.state = state
     this.id = this.state.id.name
     /** @type {RateLimitCharacteristics} */
-    this.rateLimitCharacteristics = getRateLimitingCharacteristics(this.id)
+    this.rateLimitCharacteristics = getRateLimitingPreventionCharacteristics(
```
vasco-santos, line 19:
```suggestion
    this.rateLimitCharacteristics = getRateLimitConfig(
```

PR #1568, opened 2022-03-07
fix: add json+ld markup for search results

cmunns:
Adds json+ld meta for rich snippet logo support in search results


packages/website/components/layout.js
```diff
@@ -103,6 +113,10 @@ export default function Layout({
         <meta name="twitter:card" content="summary_large_image" />
         <meta name="twitter:site" content="@protocollabs" />
         <meta name="twitter:creator" content="@protocollabs" />
+        <script
+          type="application/ld+json"
+          dangerouslySetInnerHTML={{ __html: jsonLD }}
```
dashcraft, line 118: comment (for future, non blocking):
If we end up needing more granular control over SEO than this, we might look into helmet or react-meta-tags so we can declaratively update them on docs page, blog page, etc.

Just ld json on 1 page is fine, no big deal.

dashcraft (approved):

PR #1569, opened 2022-03-07
fix: ios z-index and fixed position glitch

cmunns:
This should do it.
https://user-images.githubusercontent.com/1189523/157088689-57b21a42-8c91-42a7-a8d9-e2eec6dc260c.mov


dashcraft (approved):
lgtm 🚢

PR #1574, opened 2022-03-08
fix: metrics cron permissions and increase timeout

vasco-santos:
With https://github.com/nftstorage/nft.storage/pull/1400 in and new metrics being calculated, there were a few of new issues:
- `stats` user did not have access to cargo schema https://github.com/nftstorage/nft.storage/runs/5460080791?check_suite_focus=true#step:5:119
- timeout is not likely enough https://github.com/nftstorage/nft.storage/runs/5462071587?check_suite_focus=true#step:5:90
- sum content dag size is the clear bottleneck (should have an index like we have in web3s)

The queries taking longer are:

```
2022-03-08T09:36:15.852Z metrics:updateMetrics updateTotalUploadPast7 took: 827061ms
```
and `updateContentRootDagSizeSum` which is the long lasting one

Question: Should we consider `inserted_at` idx for upload considering the execution time of `updateTotalUploadPast7`?

### DIFF
--- a/.github/workflows/cron-metrics.yml
+++ b/.github/workflows/cron-metrics.yml
@@ -2,7 +2,7 @@ name: Cron Metrics
 
 on:
   schedule:
-    - cron: '*/15 * * * *'
+    - cron: '*/20 * * * *'
   workflow_dispatch:
 
 jobs:
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         env: ['staging', 'production']
-    timeout-minutes: 15
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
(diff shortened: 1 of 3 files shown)


vasco-santos:
The queries taking longer are:

```
2022-03-08T09:36:15.852Z metrics:updateMetrics updateTotalUploadPast7 took: 827061ms
```

and `updateContentRootDagSizeSum` which is the long lasting one

alanshaw:
> Question: Should we consider inserted_at idx for upload considering the execution time of updateTotalUploadPast7?

Yes, probably a good idea now. I will switch to using it in https://github.com/nftstorage/backup/blob/27e4c9972404daa20b1a583983c850abc5a0d69c/candidate.js#L15-L16 also then.

vasco-santos:
Cool, I'll merge this to have new timeout and will add it in new PR

Issue #1484, opened 2022-02-25
Reports of uploads ending in "Failed" status


dchoi27:
There are cases where uploads are successful (both directly to NFT.Storage and via PSA) but the status eventually ends up going from Queued to Failed. Because there's a few of these anecdotes around the same time, they might be related.

Some more specifics on the cases:
- Case 1: Someone said they successfully uploaded directly to NFT.Storage, and it was a brand new CID (since it was their first time using that specific chunk size). But it immediately returned as Failed.
- Case 2: Used pinning services API to pin a 2+ GiB directory to NFT.Storage using https://github.com/nftstorage/nft.storage-tools - their upload showed up as Failed initially, they retried and it seemed to work, and running `node status.js pinlist.txt` from nft.storage-tools gave them the screenshot below (so seems successful). however, pin status is still showing up as Failed using /check. Keeping the CID private in case it's a drop, but folks with access to the private Slack channel can grab it here: https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1645805030810139
![image.png](https://images.zenhubusercontent.com/5fa04bd64afb77249c77235b/0368cdb9-7e5f-4637-8de7-8969258c9ce3)
- Case 3: Someone with general intermittent login and upload problems (where usually when they upload, they get `"Failed to fetch"`) finally managed to upload some files, but the pin status behind those came back as `Failed` (https://discord.com/channels/806902334369824788/831502708082081822/946802019992240148). It sounds like they were still able to access the files through the gateway (though some intermittent timeouts did occur). I thought that anything that got successfully uploaded should never show pin status "Failed."

There's been some other mentions of this but we don't have much detail on those cases.

It sounds like we have all the data (good) but maybe something's up with how we're updating pin status. Think a lot of this might be user error / uploads not completing, but something feels like it's changed. Previously, in cases where the whole upload didn't finish, Cluster would report Queued for a really long time, but potentially now they update to Failed much more quickly (in which case, it sounds like it's a net improvement, and we could start using Cluster for things like reporting incomplete uploads. I tried running `SELECT date_trunc('day', inserted_at), status, count(id) FROM public.pin WHERE inserted_at > '2022-02-26' GROUP BY 1, 2` on our production DB stats replica but that timed out, but some light digging here would be helpful.

dchoi27:
[...]

PR #1577, opened 2022-03-08
feat: grace period for new pins found in error state

alanshaw:
This PR introduces a grace period for _new_ pins that are reported by cluster to be in `PinError`.

Pins will not transition to `PinError` until they are >24h old and will instead be set to `Pinning`.

Please see [this comment](https://github.com/nftstorage/nft.storage/issues/1484#issuecomment-1061685596) for the context behind this change.

resolves https://github.com/nftstorage/nft.storage/issues/1484


vasco-santos (approved):
LGTM

hugomrdias (approved):

PR #1585, opened 2022-03-08
Fix: remove localstorage, only use magic link api for auth status

### LINKED ISSUE #1556: User Auth - Remove Localstorage in favor of magic api
Storing in localstorage is unnecessary, @hugomrdias pointed out there the `isLoggedIn` method handles essentially
the same functionality as putting information in localstorage, we can very likely remove it and just use magic links
method to accomplish the same functionality without the duplicated implementation details.

https://magic.link/docs/api-reference/client-side-sdks/web#isloggedin

dashcraft:
Fixes #1556, removes localstorage and _always_ calls magic link to verify user is authenticated.

### DIFF
--- a/packages/website/components/navbar.js
+++ b/packages/website/components/navbar.js
@@ -31,7 +31,6 @@ export default function Navbar({ bgColor = 'bg-nsorange', logo, user }) {
 
   const logout = useCallback(async () => {
     await getMagic().user.logout()
-    await localStorage.removeItem('nft-user')
     await queryClient.invalidateQueries('magic-user')
     handleClearUser()
     Router.push({ pathname: '/', query: version ? { version } : null })
--- a/packages/website/components/navbar.js
+++ b/packages/website/components/navbar.js
@@ -31,7 +31,6 @@ export default function Navbar({ bgColor = 'bg-nsorange', logo, user }) {
 
   const logout = useCallback(async () => {
     await getMagic().user.logout()
-    await queryClient.invalidateQueries('magic-user')
     handleClearUser()
     Router.push({ pathname: '/', query: version ? { version } : null })
   }, [queryClient, version])
--- a/packages/website/lib/magic.js
+++ b/packages/website/lib/magic.js
@@ -49,11 +49,7 @@ export async function login(token, type = 'magic', data = {}) {
 
 export async function isLoggedIn() {
   try {
-    const meta = await getMagic().user.getMetadata()
-
-    return {
-      ...meta,
-    }
+    return await getMagic().user.getMetadata()
   } catch {
     // do nothing
   }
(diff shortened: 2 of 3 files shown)


packages/website/components/navbar.js
```diff
@@ -31,7 +31,6 @@ export default function Navbar({ bgColor = 'bg-nsorange', logo, user }) {
 
   const logout = useCallback(async () => {
     await getMagic().user.logout()
-    await localStorage.removeItem('nft-user')
     await queryClient.invalidateQueries('magic-user')
```
the-simian: can this be removed?

packages/website/lib/magic.js
```diff
@@ -49,17 +49,11 @@ export async function login(token, type = 'magic', data = {}) {
 
 export async function isLoggedIn() {
   try {
-    const exists = localStorage.getItem('nft-user')
-    if (exists) {
-      return JSON.parse(exists)
-    }
     const meta = await getMagic().user.getMetadata()
```
the-simian: return await getMagic().user.getMetadata() ?

the-simian: (nonblocking suggestion btw)

packages/website/pages/_app.js
```diff
@@ -33,6 +33,7 @@ export default function App({ Component, pageProps }) {
       // @ts-ignore
       Sentry.setUser(user)
     }
+    // @ts-ignore
```
the-simian, line 36: honestly, we should disable this rule, you shouldn't always need to observe what you call, sometimes you just want to run something once and do []

there are blogs suggesting you move the entire function into useEffect, but I don't personally agree. It can get really big and maybe you want the fn to be transportable.  I think this is just a nonsense warning with no true benefit, IMO.

the-simian: nonblocking, just my opinion

the-simian (approved):
This looks good, thank you!

the-simian:
This works but, tbh this is a bit slower than before. Its 'as fast as it used to be' but... It does seem to be calling more than once. I am assuming it was already doing this beforehand. I think we should investigate that in a followup issue, it might be a red herring or not.

I don't think we should block this, it fixes the problem, its a minor optimization if its calling more than once. Maybe magic link just does that? Worth looking at.

another thing could be if a component is unmounted and remounted, and calls magic link and its inside of a use Effect with [] it could call more than once. So we could look for that too.

dashcraft:
> This works but, tbh this is a bit slower than before. Its 'as fast as it used to be' but... It does seem to be calling more than once. I am assuming it was already doing this beforehand. I think we should investigate that in a followup issue, it might be a red herring or not.
> 
> I don't think we should block this, it fixes the problem, its a minor optimization if its calling more than once. Maybe magic link just does that? Worth looking at.
> 
> another thing could be if a component is unmounted and remounted, and calls magic link and its inside of a use Effect with [] it could call more than once. So we could look for that too.

100%, we probably need to create a control pattern or something, need to investigate in separate ticket.

PR #1598, opened 2022-03-09
feat: quick fix for logo issue

dashcraft:
There is an issue with logos on latest stats branch, refactored trusted by logos to container root dir.

### DIFF
--- a/packages/website/pages/stats.js
+++ b/packages/website/pages/stats.js
@@ -13,10 +13,11 @@ import { abbreviateNumber } from 'js-abbreviation-number'
 export function getStaticProps() {
   const logos = fs.readdirSync('public/images/marketplace-logos/home')
   // make opensea be the first logo
-  logos.sort((a, b) =>
-    a.includes('opensea') ? -1 : b.includes('opensea') ? 1 : 0
-  )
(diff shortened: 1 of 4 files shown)


packages/website/pages/stats.js
```diff
@@ -11,18 +11,20 @@ import { abbreviateNumber } from 'js-abbreviation-number'
  * @returns {{ props: import('../components/types.js').LayoutProps}}
  */
 export function getStaticProps() {
-  const logos = fs.readdirSync('public/images/marketplace-logos')
+  const logos = fs.readdirSync('public/images/marketplace-logos/home')
```
the-simian, line 14: image
this wont break but might tweak this tiny thing, I won't block but heads up

the-simian (approved):
The changes I'd like to see shouldn't block this
1. remove the // in the path. It will still work but better to have it just right
2. might have better alt text per-logo, but that's also not blocking

packages/website/pages/stats.js
the-simian: Otherwise its all showking up just fine here:
https://38990119.nft-storage-1at.pages.dev/stats/
image

PR #1612, opened 2022-03-10
Fix:  Stats page punctuation and minor refactoring

cmunns:


cmunns:
Making new clean PR...

PR #1614, opened 2022-03-10
fix: reapply stats format function and add loading

cmunns:
- fixes the safari 1px white border
- normalizes the bytes format function
- adds loading spinner while data is getting fetched


dashcraft (approved):
Lgtm!

PR #1627, opened 2022-03-12
fix: stats requires data attr

dashcraft:
Super quick, stats needs to use stats.data to pull data from API.


the-simian (approved):

the-simian:
that fixed it! thank you!

PR #1629, opened 2022-03-12
fix: upload stat isnt a byte count, but a number count

dashcraft:
Upload stat was formatted as bytes, but it's a total count not a byte count, so MB should be M for millions.


the-simian (approved):
looks good

Issue #1490, opened 2022-02-28
PSA pins changed from being pinned to cluster by source CID to content CID


hugomrdias:
Currently we use `content_cid` to pin to cluster since this [commit](https://github.com/nftstorage/nft.storage/commit/22c4e507c527d20d9a0587bee0380ea3471f45fe#diff-3c308d44c29879dc83b7cd5e55b7c3b2158a75fa948f3d8efd37c2ff3bc41679R50) but we should use `source_cid` because car uploads are pinned directly by root cid which is the `source_cid` in the db.

The issue is, when we run the cron to check pin status we dont know if we should use `content_cid` or `source_cid` to call cluster.

Possible solution is to re-pin everything that as a `source_cid` different from `content_cid`

PR #1634, opened 2022-03-14
fix: pin by source CID

alanshaw:
We need to pin by source CID to be consistent with all other uploads.

...also a fix to ensure the pin metadata is filtered correctly.

refs https://github.com/nftstorage/nft.storage/issues/1490

Part 2 is to pin all v0 source CIDs.


vasco-santos (approved):
LGTM

PR #1635, opened 2022-03-14
fix: decouple rate limit do

vasco-santos:


packages/gateway/src/gateway.js
```diff
@@ -279,6 +279,12 @@ async function updateGatewayRedirectCounter(request, env) {
  * @param {string} gwUrl
  */
 async function getGatewayRateLimitState(request, env, gwUrl) {
+  if (!shouldCheckRateLimitPrevention(gwUrl)) {
```
vasco-santos, line 282: We can make DO object being one per gateway here, Keeping the noop requests to ipfs.io and cf-ipfs. I don't think that it should be needed and we should really look into decommissioning this DO

alanshaw:
> Each request to worker means 3 requests to same DO of Rate Limit Prevention.

Is this right? We have an _instance_ per gateway. I'm not sure I understand why this fix worked? Do we not still get overloaded on the pinata gateway DO instance?

We already have an DO instance per gateway so I think we just need to do less work in the DO to improve throughput.

Ideas to improve throughput we should try:

1. `allowUnconfirmed: true` on `put`
1. Do not `await` on `put` (reading [the docs](https://developers.cloudflare.com/workers/learning/using-durable-objects/#accessing-persistent-storage-from-a-durable-object) it sounds like this is ok)
1. Not writing to `storage` at all and simply relying on the in-memory state? If evicted then we lose state but this should happen only on deploy.

I would maybe start with the first and move on to the next if throughput is still bad. It would be good for all gateways to use the same code path to make the race fair.

---

Also, as an aside, [reading this](https://developers.cloudflare.com/workers/learning/using-durable-objects/#accessing-persistent-storage-from-a-durable-object), it sounds like we don't need the in-memory instance variable because `storage` has an in memory cache.

vasco-santos:
> Is this right? We have an instance per gateway. I'm not sure I understand why this fix worked? Do we not still get overloaded on the pinata gateway DO instance?

Yes! In GatewayMetrics DO we have one per gateway, but for GatewayRateLimits DO we have one for all. This means that same DO was receiving 3 requests per user request

So, if we create 3 independent ones that will help. But I am still concerned on long term problems as we scale. I feel we should work towards not depending on DO interactions previous to return response to user. This way, we can at least not affect users if DOs get overwhelmed...

vasco-santos:
I think we can definitely try to do 1 + independent gateway as an improvement. The downsides I see are we still be prone to these errors as we scale (specially on requests not on waitUntil) and paying for all the read/write requests to DO. On the other side, it makes the race more fair (which was the initial point to keep them still performing the requests).

@alanshaw Let me know what you think

PR #1643, opened 2022-03-15
Making token_id text in the admin_search view

trigramdev9:
its already in staging, can you please validate because we need to drop the view completely and recreate


hugomrdias (approved):

trigramdev9:
@hugomrdias looks like my email doesn't have access to staging anymore.  web3 has the token_id field as text currently though, and it's working on my local.  i think it's safe to move forward with this in prod.

PR #1648, opened 2022-03-16
update holaplex logo

cmunns:
Screen Shot 2022-03-16 at 10 17 03 AM


dashcraft (approved):
comment (approved, non blocking):
Straightforward change, might make sure it fits into the new asset size requirements/compressions requirements since we're making a move towards optimizations. You're much more experienced with this than i am though, so if this is it, just ignore me.

PR #1649, opened 2022-03-16
feat: add time format function/util

cmunns:
timestamp still exists on hover, added localeDateString  formatting for local timezones. It's the exact function used on web3.storage.
Screen Shot 2022-03-16 at 10 41 18 AM


dashcraft (approved):
simple, lgtm!

PR #1389, opened 2022-02-16
feat: refactor pinning authorization logic to use user_tag table

e-schneid:
Closes #1381

Issue #1381, opened 2022-02-15
Refactor pinning authorization logic away from global CloudFlare worker variable to user_tag table


JeffLowe:
user_tag table will be managed by admin.storage app which will allow admin to manage users and give them pinning access vs updating a static array list of strings in CloudFlare worker global variables.

PR #1654, opened 2022-03-17
fix: Refactor pinning authorization logic to use user_tag table

trigramdev9:
* see #1389 #1381
* Adding a missing comma.
* Filtering tags on deleted_at is null.
* Removing some PSA_ALLOW references.

### DIFF
--- a/packages/api/src/utils/db-client.js
+++ b/packages/api/src/utils/db-client.js
@@ -83,13 +83,15 @@ export class DBClient {
     magic_link_id,
     github_id,
     did,
-    keys:auth_key_user_id_fkey(user_id,id,name,secret)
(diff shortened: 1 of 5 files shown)


e-schneid (approved):

packages/api/src/utils/db-client.js
```diff
@@ -83,13 +83,15 @@ export class DBClient {
     magic_link_id,
     github_id,
     did,
-    keys:auth_key_user_id_fkey(user_id,id,name,secret)
+    keys:auth_key_user_id_fkey(user_id,id,name,secret),
```
the-simian, line 86: some kind of editor or linter support for this would be nice. No idea if that exists, you you need an eagle eye to catch this

the-simian (approved):
Looks good, congrats on locating that

PR #1658, opened 2022-03-17
feat: issue 1642 pin req email hover

cmunns:
Screen Shot 2022-03-17 at 10 32 04 AM


dashcraft (approved):
Approved, i would double check that we always want to remove the border or square the radius on tooltips in the global sheet. If we do, this is fine, if we don't, we might want to use CSS modules or something in the meantime.

PR #1661, opened 2022-03-17
feat(docs): add how-to for checking upload status

### LINKED ISSUE #1489: Docs page on what pin statuses mean
Important points to hit:
- Pin status is just reporting what Cluster is reporting on. Just because the data doesn't say `Pinned` yet (e.g., `Queuing`) doesn't mean it's not safe
- Cluster is constantly trying to pin `Queuing` and `Failed` pins in the background
- If your data is retrievable, then Cluster has the data and will soon finish pinning
- If your data's pin status says `Failed`, then the full set of data you tried to upload wasn't available the last time NFT.Storage tried to update pin status from the Cluster
- If you used the Pinning Services API to try and pin data from somewhere else and it reports as `Failed`, then this is almost certainly because the full data was not available - this is really hard to debug because it's so situational (e.g., if you originally uploaded the content via your local IPFS node, there could be local network issues preventing connectivity, etc.). If the data was originally uploaded to Pinata or another pinning service, then it's far more likely that we should be able to grab this content, so feel free to file an issue
- If you uploaded directly to NFT.Storage, pin status reporting as `Failed` is usually because the service did not get your full upload

Since the pin status that Cluster and that NFT.Storage report back aren't always in-sync, sometimes it's tricky to debug whether a pin status `Failed` really means that your full upload wasn't gotten by NFT.Storage. We're working on new infrastructure to make this clearer, but in the meantime, the best thing to do is check whether the `Failed` status is true is to check if your upload was successful
- To check, you can try grabbing your data from an IPFS gateway. If you successfully can, then the pin status being reported by NFT.Storage just hasn't been updated
- If you uploaded a large directory, you should check whether a large sample of individual files in the directory are retrievable (e.g., `ipfs.io/ipfs/{CID of directory}/filename.png`)
- For uploads, we highly recommend using the JS client library or other tooling that programmatically uploads content. Because uploads often come across multiple requests, if there is latency in the full upload finishing, then it might report back pin status as `Failed` at some point
- NFT.Storage checks if `Failed` pins have been found and pinned by Cluster once a week for a month after initial pin status has been updated. If you find your file is retrievable but pin status still says `Failed`, then file an issue here with the relevant CIDs (or DM one of our devs on Discord if you'd like to keep the CIDs private), and we'll take a look for you

yusefnapora:
This adds a new how-to doc for checking and understanding the status of an upload. Closes #1489. 

That ticket is specifically about "pin status", but I tossed in a section about storage providers as well, since it seemed to make sense to document them both in one place.


packages/website/pages/docs/how-to/get-status.md
```diff
@@ -0,0 +1,82 @@
...
+
+The information in the file listing page is also available using the [JavaScript client's `check` method][client-js-check], and via the [HTTP API][client-http] by sending a `GET` request to the `/check/{cid}` endpoint.
+
+## Pin status
+
+When you upload data to NFT.Storage, it is ["pinned"][ipfs-docs-pinning] to several [IPFS clusters][ipfs-cluster], which are groups of machines that coordinate to provide content-addressed data to the [IPFS](https://ipfs.io) peer-to-peer network.
+
+The **Pin Status** column shows the status of the pinning operation. There are three possible values for the Pin Status column: `Queuing`, `Pinned`, and `Failed`.
+
+The `Queuing` status indicates that a request to pin the data has been submitted to a cluster node, and the node has added the request to a queue for pinning. Depending on how busy the cluster is, an upload may stay in the `Queuing` state for some time before transitioning to `Pinned` or `Failed`.
```
dchoi27, line 21: Let's add a note saying that even in the Queuing status, any data you successfully uploaded is still available. You can test this by retrieving your data from an HTTP gateway (if it's a directory, sample a set of files in the directory).

packages/website/pages/docs/how-to/get-status.md
```diff
@@ -0,0 +1,82 @@
...
+
+A `Pinned` status means that the cluster has successfully pinned the data and will retain a copy until instructed otherwise.
```
dchoi27, line 23: Would take out "until instructed otherwise"

packages/website/pages/docs/how-to/get-status.md
```diff
@@ -0,0 +1,82 @@
...
+
+A `Failed` status indicates that an operation failed while attempting to pin the data. Note that a `Failed` status may be temporary, as the cluster will continually retry `Failed` pins in the background and will update the status to `Pinned` if a retry attempt is successful.
```
dchoi27, line 25: Would leave a note that a Failed pin indicates that the initial upload attempt failed, and it will resolve back to Pinned if the data with the same CID was later successfully uploaded (though it might take a bit of time to update). If you think your upload is in this state, you can test this by retrieving your data from an HTTP gateway (if it's a directory, sample a set of files in the directory).

packages/website/pages/docs/how-to/get-status.md
```diff
@@ -0,0 +1,82 @@
...
+
+Please note that the status reported by the NFT.Storage service may not always be in sync with the actual status as reported by the IPFS cluster. The NFT.Storage service checks if `Failed` pins have been found and pinned by the cluster once per week for up to one month after the initial upload. 
+
+If you find that your file is retrievable but the pin status still says `Failed`, please [file an issue][new-issue] with the relevant CIDs and we'll take a look. If you'd prefer to keep the CIDs private, reach out to one of the devs on the `#nft-storage` [Discord channel][support-discord] in a direct message.
```
dchoi27, line 37: Wouldn't tell them to DM right away if they don't want to share the CID - they can file an issue letting us know they don't want to share the CID and we'll let them know how to proceed.

packages/website/pages/docs/how-to/get-status.md
dchoi27: you do cover some of this below, but would still float some of it up to the summary paragraph on failed pins here

packages/website/pages/docs/how-to/get-status.md
```diff
@@ -0,0 +1,82 @@
...
+
+Once enough data has been aggregated, the NFT.Storage service will propose several Filecoin [storage deals][fil-docs-deals] with different Filecoin storage providers. These will initially be shown in a "pending" status, while the provider accepts the deal and posts its commitment to the Filecoin blockchain.
+
+When using the HTTP API or client library to check the upload status, information about Filecoin storage is provided in the `deals` array field within the status response object. The `deals` array will be empty while the upload is in the `Queuing` state. 
+
+Once storage deals have been proposed, the `deals` array will contain JSON objects describing each deal. Each object has a `status` field, which will have a value of `"queued"` when the deal is pending. When the deal has been confirmed and published to the Filecoin blockchain, the `status` field will have a value of `"active"`.
+
+Here's an example JSON object that describes an active storage deal:
```
dchoi27, line 51: Would also add: "You can verify the status of your upload itself on the Filecoin blockchain at any time after the initial storage deals are made."

dchoi27 (changes requested):
Thanks for doing this! Think it's close, but think we need to be fairly prescriptive on how they can debug things themselves.

dchoi27 (approved):
LGTM!

dchoi27:
cc @mbommerez

PR #1664, opened 2022-03-17
feat: db optimizations, backups optimizations

yocontra:
Partial of #1126 (does not include partitioning).

Included in this PR:
- Improves the docker-compose for the database by specifying CPU and RAM constraints
- Updates the postgres dockerfile to use the latest PG (13) to match what we run in production/staging
  - Also improves it by adding additional logging, which is useful when working with the DB on local dev
- Improve the DB configuration
  - Many other settings aren't able to be configured like this (Heroku doesn't expose most flags), noted in #782 so I did what I could within the constraints of what we can do inside the DB
- Improves the "backup" system - instead of having its own table with a many to one relationship, add a new key `backup_urls` for each upload. This removes a redundant large table (12M uploads = > 12M backup rows), and speeds up create_upload by removing more inserts.

### DIFF
--- a/packages/api/db/config.sql
+++ b/packages/api/db/config.sql
@@ -1,10 +1,16 @@
-ALTER DATABASE postgres SET default_statistics_target = 1000;
-ALTER DATABASE postgres SET enable_partitionwise_aggregate = on;
-ALTER DATABASE postgres SET enable_partitionwise_join = on;
-ALTER DATABASE postgres SET max_parallel_workers_per_gather = 8;
-ALTER DATABASE postgres SET max_parallel_workers = 16;
-ALTER DATABASE postgres SET max_parallel_maintenance_workers = 8;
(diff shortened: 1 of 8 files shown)


hugomrdias (approved):
LGTM

yocontra:
Not optimized or tested yet against staging but migration:

```sql
ALTER TABLE "upload" ADD COLUMN "backup_urls" TEXT[];

UPDATE "upload" SET "backup_urls" = (SELECT array_agg("url") FROM "backup" WHERE "upload_id" = "uploads"."id");

DROP TABLE "backup";
```

yocontra:
@mikeal @hugomrdias This is ready to deploy - it has been run against staging and tested against a clone of the production database. Please let me know what next steps are.

hugomrdias (approved):

PR #1671, opened 2022-03-20
fix: link to Getting Started

intelliot:
In the Website README, this fixes the link to the Getting Started section of the README, which was previously broken. Since GitHub uses `blob` for files, we need to link to the file (instead of the directory).

### DIFF
--- a/packages/website/README.md
+++ b/packages/website/README.md
@@ -4,7 +4,7 @@ Frontend build for the NFT.Storage website.
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


intelliot:
The link in the GitHub diff doesn't work because it points to this repo instead of my fork. You can confirm that these links work:

https://github.com/intelliot/nft.storage/blob/patch-1/README.md#getting-started

https://github.com/nftstorage/nft.storage/blob/main/README.md#getting-started

alanshaw (approved):
LGTM - thanks!

PR #1679, opened 2022-03-21
fix: update /files tooltip copy

cmunns:
Screen Shot 2022-03-21 at 11 12 27 AM
Screen Shot 2022-03-21 at 11 12 24 AM


dashcraft (approved):
I assume verbiage is copied from a copy sheet/scope, so this is approved under that assumption

PR #1681, opened 2022-03-21
fix: swap stats image, img already exists in assets folder

cmunns:
Image was already present.
Screen Shot 2022-03-21 at 12 04 26 PM


dashcraft (approved):
lgtm!

Issue #1653, opened 2022-03-17
Website loading improvements


hugomrdias:
Our current ligthhouse perf score is 63 we can improve this with minimal effort.

# Images
We should optimise the size of these images

<img width="606" alt="Screen Shot 2022-03-17 at 11 41 05" src="https://user-images.githubusercontent.com/314190/158801172-ccc20304-c956-421d-876b-dad5fd6fbbfd.png">

Manually run them through https://squoosh.app/ or automatic using `next/image` and cloudflare images.
Example:

```js
// image.js
import Image from 'next/image'

const cloudflareImageLoader = ({ src, width, quality }) => {
  if (!quality) {
    quality = 75
  }
  return `https://nft.storage/cdn-cgi/image/format=auto,width=${width},quality=${quality}${src}`
}

/**
 * @param {import('next/image').ImageProps} props
 */
export default function Img(props) {
  if (process.env.NODE_ENV === 'development') {
    return <Image unoptimized={true} {...props} />
  } else {
    return <Image {...props} loader={cloudflareImageLoader} />
  }
}

// page.js

<Img
  src="/images/getting-started.png"
  alt="getting-started"
  width={1412}
  height={784}
  layout="responsive"
/>

// next.config.js - workaround for next export

images: {
  loader: 'imgix',
  path: '',
},



```

This second option gives us the ability to use next-gen formats like avif and webp.

We should also make sure that ALL images have explicit width and height.

# Lodash

Full lodash.min is being included in a chunk bundle and loaded in the landing page.

We should remove it completely from the code base (its also not listed as a dependency) and find another way to implement the feature in plain js. https://github.com/nftstorage/nft.storage/blob/4ab48ce19657f5f6a03ea2fe0a98d2d10f0e5509/packages/website/lib/statsUtils.js#L1

Ligthhouse reports other minor issues that we should improve.

PR #1683, opened 2022-03-21
fix: lots of small optimizations led to big gains

dashcraft:
I made some changes, the original starting point for me locally was 63% for performance.

Naturally, these changes may not entirely propagate the same results in prod, but this is a decent measurement that performance.

Need to optimize images more though.

image


the-simian:
Unfortunately the cloudflare build broke but I can still provide some feedback in some places that tI think will still help this PR

the-simian:
Honestly I'm amazed at how much the next Image component really helped performance, good work.

hugomrdias:
can you please remove the sitemap.xml? probably should go into the git ignore to avoid this in the future.

```bash
Error: Image Optimization using Next.js' default loader is not compatible with `next export`.
--
23:07:17.811 | Possible solutions:
23:07:17.811 | - Use `next start` to run a server, which includes the Image Optimization API.
23:07:17.811 | - Use any provider which supports Image Optimization (like Vercel).
23:07:17.811 | - Configure a third-party loader in `next.config.js`.
23:07:17.811 | - Use the `loader` prop for `next/image`.
23:07:17.811 | Read more: https://nextjs.org/docs/messages/export-image-api
```

FYI this fails in the same way locally when you run `yarn build`

You can't use `next/image` like this, it does not work with static exports. You should do something like i described in https://github.com/nftstorage/nft.storage/issues/1653.

hugomrdias:
you can use something like `image-size` inside a `getStaticProps` to get the original sizes when needed

the-simian (approved):
1. I see you got the cloudflare image in here: https://github.com/nftstorage/nft.storage/blob/2ad0687f67b6c093211b7ef89a04a861cde586a5/packages/website/components/cloudflareImage.js
2. and you implemented the code changes to get lodash out.  
3. I also ran a perf check:
Screen Shot 2022-03-28 at 10 37 03 AM

So I think we're good!

Issue #1672, opened 2022-03-20
Uploading error, please help!


M1lfsh4k3:
I have uploaded entire directories before using nftstorage so I know that this should work.
The following error shows when executing "node storeDirectory.mjs "path""

PS C:\Users\alors\Desktop\folder\NFT engine\nft.storage-main> node storeDirectory.mjs "C:\Users\alors\Desktop\folder\folder1\Generator\build\images"
storing 10750 file(s) from C:\Users\alors\Desktop\folder\folder1\Generator\build\images
node:events:498
      throw er; // Unhandled 'error' event
      ^

Error: EMFILE: too many open files, open 'C:\Users\alors\Desktop\folder\folder1\Generator\build\images\7693.png'
Emitted 'error' event on ReadStream instance at:
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -4066,
  code: 'EMFILE',
  syscall: 'open',
  path: 'C:\\Users\\alors\\Desktop\\folder\\folder1\\Generator\\build\\images\\7693.png'
}

Is it because its too many files? In that case how do I resolve this problem because all images has to be under the same CID folder when it's done. Thank you in advance!

M1lfsh4k3:
Please can someone help me with this. Is it possible to upload in two batches but to the same CID folder?

alanshaw:
@vasco-santos looking into this we might be able to fix by switching to https://www.npmjs.com/package/graceful-fs in [files-from-path](https://github.com/web3-storage/files-from-path) because it:

> Queues up open and readdir calls, and retries them once something closes if there is an EMFILE error from too many file descriptors.

@M1lfsh4k3 I'll look into this tomorrow morning and hopefully get a fix out asap.

PR #1693, opened 2022-03-22
fix: create stream when needed

alanshaw:
Replaces the import candidate `content` property with a getter function that allows the stream to be created (and file descriptor opened) only when it is accessed.

The issue is that when there are many thousands of files in a directory, we'll attempt to open that many file descriptors eagerly, which can cause `EMFILE` errors on some systems.

refs https://github.com/nftstorage/nft.storage/issues/1672

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -392,7 +392,7 @@ class NFTStorage {
     if (blob.size === 0) {
       throw new Error('Content size is 0, make sure to provide some content')
     }
-    return packCar([{ path: 'blob', content: blob.stream() }], {
+    return packCar([toImportCandidate('blob', blob)], {
       blockstore,
       wrapWithDirectory: false,
     })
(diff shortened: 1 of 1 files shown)


packages/client/src/lib.js
```diff
@@ -392,7 +392,7 @@ class NFTStorage {
     if (blob.size === 0) {
       throw new Error('Content size is 0, make sure to provide some content')
     }
-    return packCar([{ path: 'blob', content: blob.stream() }], {
+    return packCar([toImportCandidate('blob', blob)], {
```
olizilla, line 395: Not an issue with this PR, but does having the path as 'blob' and using `pack` from ipfs-car with it's default `wrapWithDirectory:true` mean this has been creating lots of DAGs of a directory with a single entry of `blob`? If not, why not? is 'blob' used as a magic marker, or is `wrapWithDirectory` being set to `false` somewhere?

olizilla (approved):
Neat fix!

hugomrdias (approved):

PR #1694, opened 2022-03-22
feat: labeling uploads on files

cmunns:
Output ready.
Screen Shot 2022-03-22 at 3 00 26 PM


dashcraft (approved):
Should be good, deploying to staging for double check.

Issue #1242, opened 2022-02-04
skip running release-please on PRs from forks


yusefnapora:
This will avoid errors like [this one](https://github.com/nftstorage/nft.storage/runs/5071360929?check_suite_focus=true) on https://github.com/nftstorage/nft.storage/pull/1240 - they're not a blocker, since the action will succeed once the PR is merged. But it will be nice to get rid of the red X 😄

gobengo:
@yusefnapora Currently this website.yml 'release' job is run on all PRs into main
* Just checking... but are there times where we do expect this to work? i.e. releasing a feature-branch?
* Assuming yes, the smallest change here would be to keep running the release job on PRs to main, *unless* the source branch is on a repo not-owned by github.com/nftstorage (i.e. exclude forks)

gobengo:
candidate solution: https://github.com/nftstorage/nft.storage/pull/1698

PR #1695, opened 2022-03-23
Replace 'website' action with website_pr and website_cd

gobengo:
Motivation: https://github.com/nftstorage/nft.storage/issues/1242

* Get rid of old 'website.yml' gh action that would run both on PRs to main and push to main
* Add website_pr.yml action
  * runs tests on all PRs modifying website, not just those into main branch
* Add website_cd.yml (continuous delivery)
  * runs tests - seems worth the time-cost to re-test before release
  * runs release-please - my understanding is that this action creates a PR from main -> release branch (so it's fine to never run it on PRs)

### DIFF
--- a/.github/workflows/website_cd.yml
+++ b/.github/workflows/website_cd.yml
@@ -1,11 +1,6 @@
-name: Website
+name: Website Delivery
 on:
   push:
-    branches:
-      - main
-    paths:
-      - 'packages/website/**'
-  pull_request:
     branches:
       - main
     paths:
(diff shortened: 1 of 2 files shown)


gobengo:
Once something like this is merged, I think we can verify it by opening this PR https://github.com/gobengo/nft.storage/pull/new/issues/1242-testpr

yusefnapora (approved):
This looks like it should work, thanks @gobengo :)

gobengo:
Feedback from @hugomrdias: https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1648068052302369?thread_ts=1648065445.982479&cid=C02BZPRS9HP

PR #1698, opened 2022-03-23
ci: github actions dont run release-please unless push-to-main

gobengo:
Motivation:
* https://github.com/nftstorage/nft.storage/issues/1242
* second time (last attempt [reverted](https://github.com/nftstorage/nft.storage/pull/1697)) 

How to Test:
* once this is merged, it should prevent this other PR from breaking: https://github.com/nftstorage/nft.storage/pull/1723
  * 'proof' that that pr would work if not from fork: https://github.com/nftstorage/nft.storage/pull/1725


.github/workflows/website.yml
```diff
@@ -22,7 +22,9 @@ jobs:
           node-version: 16
       - uses: bahmutov/npm-install@v1
       - run: yarn test:website
+      - run: yarn build:website
```
gobengo, line 25: Motivation: https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1648068210775589?thread_ts=1648065445.982479&cid=C02BZPRS9HP

(this will fail faster than the status quo, which is to only fail downstream)

alanshaw: Build before test?

alanshaw: FYI cloudflare builds the site for the preview link and it will fail there also. However the logs for that build are not available to community contributors so we're surfacing the error here also.

@gobengo lets avoid linking to the private slack channel from PRs as context gets lost for folks without access 🙏

hugomrdias: i pointed to this in slack but alan is right better to but it before test

hugomrdias (changes requested):
LGTM just move the build before the tests

.github/workflows/website.yml
gobengo: > lets avoid linking to the private slack channel from PRs as context gets lost for folks without access 

@alanshaw - sure thing. forgot it was private!

gobengo: moved build before test in https://github.com/nftstorage/nft.storage/pull/1698/commits/ea179157a5d0a70102e1bb70ecfc37857dde0fcc

yusefnapora (approved):
This looks good to me, although I did say that last time too... 😛

It looks like you've covered all the bases here and addressed the change request from @hugomrdias, so I'd say this is safe to merge.

gobengo:
@hugomrdias any more feedback or change requests?

hugomrdias (approved):

.github/workflows/api.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           DATABASE_CONNECTION: postgresql://postgres:postgres@localhost:5432/postgres
   deploy-dev:
     name: Deploy Dev
-    if: github.event_name == 'pull_request' && github.ref != 'refs/heads/main'
+    if: github.event_name == 'pull_request' && github.ref != 'refs/heads/main' && secrets.CF_API_TOKEN != null
```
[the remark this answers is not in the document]
gobengo: Sorry! Fixed syntax on main. Will work on [this way](https://github.com/nftstorage/nft.storage/pull/1791) of supporting forks without that var

Issue #1717, opened 2022-03-29
Gateway CID blocklist


vasco-santos:
We will receive takedown requests for problematic CIDs over time and we should have a solution for this. At the moment, we are relying on Cloudflare WAF with hardcoded domain but it would be wise to get a KV store in place where we could have a status and reason for better management of blocked CIDs.

@alanshaw 's suggestion: Key is normalized CID and value is a JSON object with status and reason.

olizilla:
Suggestions:
- talk to Cloudflare about this and see if we can get an unlimited WAF rules.
- add problematic CIDs to the bad bits list and check that for inclusion https://badbits.dwebops.pub/

alanshaw:
Situation with Cloudflare:

> Can I purchase additional active rules?
> No. The number of active rules is fixed based on customer plan. Refer to [Entitlements](https://developers.cloudflare.com/firewall/cf-firewall-rules/#entitlements)

We have the max (1,000 rules) already.

Note we currently have 1 rule blocking 3 CIDs, but the expression [limit is 4KB](https://developers.cloudflare.com/firewall/cf-dashboard/create-edit-delete-rules/#expression-builder-and-editor) so it can grow a bit but not loads.

I was thinking originally that using the firewall would be the way to go but now I think KV is probably better/easier - we can add millions and millions of records here from various sources.

olizilla:
## Notes on Google Safe Browsing

We are looking at 72hrs until the "Deceptive site ahead" phishing protection warning clears...

> Phishing review processing time: Phishing reviews take about a day to process. If successful, the user-visible phishing warning will be removed and your page may again surface in search results.
>
> If Google finds that your site is clean, warnings from browsers and search results should be removed within 72 hours.
>
> – https://web.dev/request-a-review/#c.-phishing-or-social-engineering

olizilla:
Entries in the badbits list are
- Normalised to base32 cidv1
- Include a path and `/` is set as the default
- SHA-256 hashed

e.g

```js
sha256(cidv0 + "/")
```
see: https://github.com/protocol/badbits.dwebops.pub/blob/main/badbits-lambda/helpers.py#L7-L17

vasco-santos:
Created https://github.com/publicsuffix/list/pull/1548 for publicsuffix

olizilla:
Also worth fixing http -> https redirect rule for nftstorage.link as the `http` url currently turns up a domain parking link farm which doesn't help our look

[...]

PR #1721, opened 2022-03-29
feat: gateway denylist

alanshaw:
Adds a denylist to the gateway as a cloudflare workers KV store. Also CLI commands for syncing with our local `denylist.json` and the [badbits denylist](https://badbits.dwebops.pub/denylist.json).

Our `denylist.json` is the _same_ format as the badbits denylist but has optional properties `status` and `reason` which enables us to communicate a custom HTTP status and reason for blockage for a given CID. We _may_ want to send a 451 for court orders.

refs https://github.com/nftstorage/nft.storage/issues/1717


packages/gateway/scripts/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+```sh
+CF_API_TOKEN=<token> node scripts/cli.js denylist sync --env production
+```
+
+It reads from the local `denylist.json` as well as the [badbits denylist](https://badbits.dwebops.pub/denylist.json). Sources can be updated in `denylist.js`.
+
+# `denylist add`
+
+Add a CID (or CID + path) to the local deny list.
```
vasco-santos, line 19: Can we add instructions on how to do CID + path? I think an example would be good

packages/gateway/wrangler.toml
```diff
@@ -31,6 +31,7 @@ bindings = [
 account_id = "fffa4b4363a7e5250af8357087263b3a" # Protocol Labs CF account
 zone_id = "c7795a0adce7609a95d62fec04705aff"    # nftstorage.link zone
 route = "*.ipfs.nftstorage.link/*"
+kv_namespaces = [{ binding = "DENYLIST", id = "" }]
```
vasco-santos, line 34: TODO

vasco-santos: We also need for staging

packages/gateway/src/gateway.js
```diff
@@ -53,6 +56,16 @@ export async function gatewayGet(request, env, ctx) {
   const cid = getCidFromSubdomainUrl(reqUrl)
   const pathname = reqUrl.pathname
 
+  if (env.DENYLIST) {
+    const anchor = await toDenyListAnchor(`${cid}/`)
+    // TODO: in theory we should check each subcomponent of the pathname also.
```
vasco-santos, line 61: So, this means go through the path and test all wrapping directories if they exist?

packages/gateway/scripts/denylist.js
```diff
@@ -0,0 +1,149 @@
...
+const rootDir = path.dirname(path.dirname(import.meta.url))
+const wranglerConfigPath = `${rootDir}/wrangler.toml`
+const denyListPath = `${rootDir}/denylist.json`
+
+const DENY_LIST_SOURCES = [
+  'https://badbits.dwebops.pub/denylist.json',
+  denyListPath,
+]
+
+export async function denylistSyncCmd({ env }) {
```
vasco-santos, line 25: Should we trigger this in a daily cron job ?

packages/gateway/scripts/denylist.js
```diff
@@ -0,0 +1,149 @@
...
+ * @typedef {{ name: string, metadata: any }} Key
+ * @typedef {{ key: string, value: any, metadata?: any }} BulkWritePair
+ */
+
+const rootDir = path.dirname(path.dirname(import.meta.url))
+const wranglerConfigPath = `${rootDir}/wrangler.toml`
+const denyListPath = `${rootDir}/denylist.json`
+
+const DENY_LIST_SOURCES = [
+  'https://badbits.dwebops.pub/denylist.json',
```
vasco-santos, line 21: What do you think on an ENV var for this? We could have multiple ones splitted by "," and if we automate this script we can add new lists without needing to PR

packages/gateway/scripts/denylist.js
```diff
@@ -0,0 +1,149 @@
...
+    throw new Error('missing binding in kv_namespaces: DENYLIST')
+  }
+  console.log(`🪢 using KV binding: DENYLIST (${denyListKv.id})`)
+
+  for (const url of DENY_LIST_SOURCES) {
+    console.log(`🦴 fetching ${url}`)
+    const denyList = await getDenyList(url)
+    const kvs = denyList.map(({ anchor: key, status, reason }) => ({
+      key,
+      value: { status, reason },
```
vasco-santos, line 53: given we won't have status/reason in any place at the moment, wondering if we should store an origin property with where the CID+Path combo is registered, or add a default value to reason including it? what do you think?

packages/gateway/scripts/denylist.js
alanshaw: Hmm, maybe, I'd probably make it a command line parameter like `--additional-sources` though as it's not private. There's part of me that thinks if we're going to add a new source we should probably record that in VCS though - a PR to ask the team to sign off on using a new source sounds like a good thing.

packages/gateway/scripts/denylist.js
alanshaw: Yes, but I'll open an issue to get that done...

packages/gateway/scripts/denylist.js
alanshaw: I'd probably not have a default reason that includes the origin because in the case when we need to set a reason it would not be a reliable way to know the origin.

I'd be ok with putting `origin` as a separate property...but where would we use it?

packages/gateway/src/gateway.js
alanshaw: We decided on a call that we'd not support this right now as the vast majority of blocked items are CID only.

vasco-santos: Maybe worth rephrasing this comment to what we decided for now, and what are the options we can consider moving to in the future? (and an issue to track it)

vasco-santos (approved):
LGTM, great work

packages/gateway/src/gateway.js
alanshaw: I will open an issue for this as well and elaborate!

alanshaw: https://github.com/nftstorage/nft.storage/issues/1737

PR #1724, opened 2022-03-29
chore: add DEVELOPMENT.md for dev setup docs

### LINKED ISSUE #1691: add developer docs for local DB setup
There are some cool scripts in the API package for setting up and managing the local DB instance. We should document how to use them in the README.

Some use cases to cover:

CI / unit testing:
- [ ] resetting things if the test framework throws and fails to clean up
- [ ] intentionally not cleaning up after a test run, to inspect weird database states

Local development:
- [ ] setup stateful local env (same as test setup but without automatic resets)
- [ ] how to wipe / reset manually

@hugomrdias if there are any other cool tricks we should document, let me know :)

yusefnapora:
This pulls the environment setup instructions out of the README and into a new DEVELOPMENT.md file, so we can have a home for all the local setup info a dev might need. I also moved the release please info and commit style guide, but I'm having second thoughts about that... maybe it's best for it to stay front-and-center on the main README.

To avoid the need for copy/paste, I added a `.env.example` file you can just rename to `.env` to get a working setup, apart from the magic link secret key. 

I also added a `.env.development` file to `packages/website`, so the only thing you need to set for the frontend is the public magic link key (in `.env.local`).

Once we get rid of the hard dependency on magic.link, you should be able to just copy the example env over and run locally without any configuration.

I also added a list of all the env vars used in the project, or at least the ones that were in the old README instructions. That might not belong in DEVELOPMENT.md, but I don't have enough material for a DEPLOY.md yet 😄...

Leaving this as a draft for now since I'd like to edit a bit more tomorrow, but feel free to leave feedback if you have it.

TODO:
- [x] update READMEs for packages/api and packages/website to link to DEVELOPMENT.md for local setup info

I added some info about the DB cli scripts, to close #1691. Could still use some fleshing out with use case examples though.


hugomrdias (approved):

.env.example
```diff
@@ -0,0 +1,30 @@
+# needs to be real so create a personal magic.link account or use the staging key
```
vasco-santos, line 1: Could we use `.env.tpl` like we do in web3 side?

https://github.com/web3-storage/web3.storage/blob/main/.env.tpl

yusefnapora: sure, good call

Issue #1688, opened 2022-03-22
add automated screenshot capture to CI


yusefnapora:
To make front-end issues easier to diagnose, let's set up automatic screenshots in our CI flow to grab screenshots of our most important pages at different responsive breakpoints.

@hugomrdias recommends that we use [playwright's screenshots feature](https://playwright.dev/docs/screenshots), which seems to have everything we need.

hugomrdias:
https://playwright.dev/docs/test-snapshots

gobengo:
* [ ] POC PR curling website homepage for 200 ok
* [ ] POC PR taking screenshot of website homepage using playwright

gobengo:
https://github.com/playwright-community/playwright-jest-examples/pull/22/files

gobengo:
@hugomrdias comment is a good one for implications on how to use this screenshotting capability: to be alerted to unexpected visual regressions by checking-in a screenshot of what we expect it  to look like, and then asserting that the screenshot-from-ci is similar enough. Will be implementing towards this using `@playwright/test` and then the `.toMatchScreenshot()` assertion.

yusefnapora:
Great work on this @gobengo. Also, thanks for making an issue to track the web3 side. I think we can call this one done :)

PR #1725, opened 2022-03-29
ci: add @playwright/test and test homepage (not from fork)

gobengo:
Motivation:
* implementing https://github.com/nftstorage/nft.storage/issues/1688
* be same as https://github.com/nftstorage/nft.storage/pull/1723 , except source branch is not from my fork (so is not affected by #1242 (hopefully fixed by #1698)

Changes:
* .github/workflows/website.yml
  * is now triggered on pull_requests targeting branches beyond just 'main'
  * has new 'test-e2e' job, which matrixes by some things and invokes `yarn workspace website run test:e2e -- --project=${{ matrix.browser }}`
* in packages/website
  * add `test:e2e` script, which invokes https://playwright.dev/docs/test-assertions
  * add a simple e2e test that asserts the homepage describes the nft.storage product (more can come as followups once the ci/runner is merged)
  * ~add an e2e test that asserts visual screenshot of homepage: more here https://playwright.dev/docs/test-snapshots~ - [removed](https://github.com/nftstorage/nft.storage/pull/1725/commits/c07715d085852465fc821665b81b5c156bc87a23) because this may increase costs more than it's worth
    * especially take note on those docs of how to generate snapshots for linux (e.g. while using a mac host os) using docker

Potential followups:
* debug why 'webkit' matrix resulted in blank/white homepage.png ([example](https://github.com/nftstorage/nft.storage/actions/runs/2066549701)) and reenable it - @gobengo has no idea and it seems not-so-required to test webkit if we're also checking firefox/chromium


yusefnapora:
Neat!

I'm a little nervous about enabling assertions on screenshot diffs, and I know @dashcraft has similar reservations. I'm probably not the best judge, since I haven't used screenshot regression testing in the past; the closest analogue I've used is Jest snapshot tests, which assert on the virtual DOM output. IME those tend to be pretty "noisy", and after a while people just start updating the snapshots whenever they make a deliberate change without actually digging into the differences, which kind of defeats the purpose.

If we do enable the assertions, we should add docs somewhere on how to update the snapshots, and we should probably have an npm script to do the updating, since there's some Docker voodoo involved. If we're not going to assert, we should get the action to link out to the collection of screenshots (or post a comment on the PR), so humans can sanity check  them during the review process.

Tagging @alanshaw, @olizilla, @redaphid @funkytek to get some 👀 and opinions... how do you guys feel about CI enforcement of visual changes like this?

funkytek:
I think there are some use-cases for blocking on UI regressions (feature-locked production UI that will be stable for a long time).  Since I think we tend to be a bit more fluid on this project with changes I agree that sticking with automating screenshots for reviewers to check would be a great value add

gobengo:
* [ ] remove the playwright snapshot assertion
* [ ] investigate just taking a screenshot and making it available to the reviewer via upload artifacts

PR #1731, opened 2022-03-30
test: add homepage.e2e test that saves screenshot homepage.png

gobengo:
* note target branch is issues/1688-screenshots
  * because I want to test the gh action before merging into there while in review


gobengo:
See 'artifacts' here: https://github.com/nftstorage/nft.storage/actions/runs/2066549701
There is a zip file per matrix combination of os/browser, which contain a 'homepage.png' file.

PR #1755, opened 2022-04-03
feat: content_cid not id is used from nft data upload not uploads

dashcraft:
We had been using ID but id is not accurate, content_cid is the id that relates nft uploads to uploads.

### DIFF
--- a/packages/api/src/routes/nfts-upload.js
+++ b/packages/api/src/routes/nfts-upload.js
@@ -176,7 +176,7 @@ export async function nftUpdateUpload(event, ctx) {
 
     // id is required for updating
     if (!content_cid)
-      return new JSONResponse({ ok: false, value: 'ID is required' })
+      return new JSONResponse({ ok: false, value: 'upload CID is required' })
 
     const body = await event.request.json()
     const { name } = body
--- a/packages/api/src/routes/nfts-upload.js
+++ b/packages/api/src/routes/nfts-upload.js
@@ -172,17 +172,17 @@ export async function nftUpdateUpload(event, ctx) {
   const { params, db } = ctx
   try {
     const { user } = await validate(event, ctx)
-    const content_cid = params.id
+    const { cid } = params
 
     // id is required for updating
-    if (!content_cid)
+    if (!cid)
       return new JSONResponse({ ok: false, value: 'upload CID is required' })
 
     const body = await event.request.json()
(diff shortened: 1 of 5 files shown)


packages/api/src/utils/db-client.js
```diff
@@ -125,20 +125,20 @@ export class DBClient {
...
     }
 
     if (!upload) {
       throw new Error(
-        `Cannot update upload ${JSON.stringify(data.id)} ${JSON.stringify(
-          status
-        )}`
+        `Cannot update upload ${JSON.stringify(
+          data.content_cid
+        )} ${JSON.stringify(status)}`
```
alanshaw, line 141: Why are these strings/numbers being passed to `JSON.stringify`?

packages/api/src/routes/nfts-upload.js
```diff
@@ -172,15 +172,20 @@ export async function nftUpdateUpload(event, ctx) {
...
-    const id = params.id
+    const content_cid = params.id
 
     // id is required for updating
-    if (!id) return new JSONResponse({ ok: false, value: 'ID is required' })
+    if (!content_cid)
+      return new JSONResponse({ ok: false, value: 'ID is required' })
 
     const body = await event.request.json()
     const { name } = body
```
alanshaw, line 182: Can we please validate this is a string?

packages/api/src/routes/nfts-upload.js
```diff
@@ -172,15 +172,20 @@ export async function nftUpdateUpload(event, ctx) {
   const { params, db } = ctx
   try {
     const { user } = await validate(event, ctx)
-    const id = params.id
+    const content_cid = params.id
 
     // id is required for updating
-    if (!id) return new JSONResponse({ ok: false, value: 'ID is required' })
+    if (!content_cid)
+      return new JSONResponse({ ok: false, value: 'ID is required' })
```
alanshaw, line 179:
```suggestion
      return new JSONResponse({ ok: false, value: 'upload CID is required' })
```

alanshaw (changes requested):

packages/api/src/utils/db-client.js
dashcraft: Leftover cleanup, good catch!

alanshaw (approved):

Issue #1745, opened 2022-03-31
Failed remote pins


Eroo36:
For some weeks now everything i remote pin to nft.storage shows failed after queue in the web ui. But i think they are pinned fine. I already mentioned this in discord and it was indeed the case. Here are some example files to check status:
QmfQX2FPDSXAh8Hj2sXRGYPXFcxMF1kdTK2n63uzXtjXzc
QmV8VauacrAZNLGWj5e8YEDqK581jhwSkBP3am6wFBybhL
QmQQWuUAysJicwMgmxanLEHZCKktZAcaXSjFyZ8t2QB719

![Screenshot_2022-03-31-20-54-11-592_com.android.chrome.jpg](https://user-images.githubusercontent.com/13486466/161119155-32aadd90-7a89-4bfd-bb17-e12b5a264023.jpg)

OnahProsperity:
This same issue brought me here.

PR #1757, opened 2022-04-04
fix: use source_cid in pin cron job

alanshaw:
TLDR; we pin to cluster using `source_cid`, but the pins cron job checks status by `content_cid` so it is not found. For most data this is not a problem but when v0 CIDs are pinned their status never gets updated to `Pinned`.

This is a long standing bug that was masked by a regression that got fixed recently here: https://github.com/nftstorage/nft.storage/pull/1634

In Web3.Storage this query is harder so we actually downgrade `content_cid` to v0 when found to be unpinned everywhere and re-check with cluster: https://github.com/web3-storage/web3.storage/blob/3231807f24190de3e5be810ab25ed9567fda41cb/packages/cron/src/jobs/pins.js#L73-L75

Fixes https://github.com/nftstorage/nft.storage/issues/1745


vasco-santos (approved):
LGTM

PR #1759, opened 2022-04-04
feat: add deny list sync cron job

### LINKED ISSUE #1738: Trigger deny list sync on cron job
This will allow us to automatically deny serving content that has been added to other lists we trust.

alanshaw:
resolves #1738

### DIFF
--- a/.github/workflows/cron-denylist.yml
+++ b/.github/workflows/cron-denylist.yml
@@ -1,0 +1,25 @@
+name: Cron sync deny list
+
+on:
+  schedule:
+    - cron: '13 8 * * *'
+  workflow_dispatch:
+
+jobs:
+  update:
+    name: Sync deny list with KV store
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        env: ['staging', 'production']
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/cron-denylist.yml
```diff
@@ -0,0 +1,25 @@
...
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - uses: bahmutov/npm-install@v1
+      - name: Run job
+        env:
+          CF_API_TOKEN: ${{ secrets.CF_API_TOKEN }}
```
vasco-santos, line 24:
```suggestion
          CF_API_TOKEN: ${{ secrets.CF_GATEWAY_TOKEN }}
```

vasco-santos (changes requested):

.github/workflows/cron-denylist.yml
vasco-santos: We use a new token since we moved to nftstorage.link zone

alanshaw: Ah yes, ty

vasco-santos (approved):

PR #1761, opened 2022-04-04
refactor: deny list sync schedule

alanshaw:
Runs every 5h and on merge our denylist.json to main.

### DIFF
--- a/.github/workflows/cron-denylist.yml
+++ b/.github/workflows/cron-denylist.yml
@@ -2,8 +2,13 @@ name: Cron sync deny list
 
 on:
   schedule:
-    - cron: '13 8 * * *'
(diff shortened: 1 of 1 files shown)


.github/workflows/cron-denylist.yml
```diff
@@ -2,8 +2,13 @@ name: Cron sync deny list
...
 on:
   schedule:
-    - cron: '13 8 * * *'
+    - cron: '13 0,5,10,15,20 * * *'
   workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - 'packages/gateway/denylist.json'
```
vasco-santos, line 11: 👍🏼

vasco-santos (approved):

PR #1768, opened 2022-04-05
fix: four fe bugs

Codigo-Fuentes:


cmunns (approved):
Looks good.

packages/website/pages/files.js
```diff
@@ -555,50 +555,50 @@ export default function Files({ user }) {
...
+                  id="files-next"
+                  tracking={{
+                    event: countly.events.FILES_NAVIGATION_CLICK,
+                    ui: countly.ui.FILES,
+                    action: 'Next',
+                  }}
+                >
+                  Next →
+                </Button>
+              </div>
```
alanshaw, line 601: Why did these move out of the `When` block?

Codigo-Fuentes: So that they don't scroll horizontally with the table

PR #1770, opened 2022-04-05
fix: temporarily hide upload name column

alanshaw:
This allows us to release the blog posts. API can then be deployed separately, and then the frontend (with this change reverted).


dashcraft (approved):
Looks good on cloudflare preview!

PR #1773, opened 2022-04-05
fix: lift pg statement_timeout

yocontra:
- Raises the database statement timeout to 3 minutes, this should accommodate the existing cron jobs that take ~30s to run
- Adds a localized 30s timeout to uploads
- A future improvement would be to isolate the cron jobs and postgrest users in the DB, and each user can have a statement_timeout config aligned with their actual expected times


alanshaw (approved):

hugomrdias (approved):

hugomrdias:
@yocontra did you apply this in staging and prod ?

yocontra:
@hugomrdias It is applied on production

JeffLowe:
Status check here @hugomrdias @alanshaw 
Does this need to be merged?

Issue #1736, opened 2022-03-31
Upgrade ipfs-cluster to v1


dchoi27:
Sister issue to: https://github.com/web3-storage/web3.storage/issues/1184

~~...but we want to do it first for NFT.Storage~~

olizilla:
For this, we will need

- Update dep on ipfs-cluster client to v5.0.1
- Merge to main and update **staging** env cloudflare worker secrets for `CLUSTER_SERVICE` to `IpfsCluster` & `CLUSTER_BASIC_AUTH_TOKEN` from 1pass - to use the new staging cluster running ipfs-cluster@1.0.0-rc.4
- Test on staging
- Wait till maintenance window: https://status.nft.storage/incidents/m54cpnvpm3zz
- Enable read-only mode
- Deploy ipfs-cluster@1.0.0-rc.4 binaries to production cluster
- Merge API release PR to deploy api to prod with support for ipfs-cluster@1.0.0-rc.4
- Merge cron pin update for cluster@1.0 PR to update cron jobs with support for ipfs-cluster@1.0.0-rc.4
- Monitor
- FIN

PR #1789, opened 2022-04-08
feat: api support for ipfs-cluster@1.0

olizilla:
The IPFS Cluster v1 release has much improved perf, and some breaking API changes.

In this PR we upgrade the ipfs-cluster client and local test environment to work with the v1 cluster API.

CI was [failing for the API](https://github.com/nftstorage/nft.storage/actions/runs/2115915629), so this PR includes a change to that workflow to fix that.

After merging this PR 
- Update staging env cloudflare worker secrets for CLUSTER_SERVICE to IpfsCluster & CLUSTER_BASIC_AUTH_TOKEN from 1pass - to use the new staging cluster running ipfs-cluster@1.0.0-rc.4
- open a follow PR to update the cron package dep on ipfs-cluster client.

see: https://github.com/nftstorage/nft.storage/issues/1736
see: https://github.com/nftstorage/ipfs-cluster/releases/tag/v5.0.0

License: (Apache-2.0 AND MIT)

### DIFF
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -28,7 +28,7 @@ jobs:
           DATABASE_CONNECTION: postgresql://postgres:postgres@localhost:5432/postgres
   deploy-dev:
     name: Deploy Dev
-    if: github.event_name == 'pull_request' && github.ref != 'refs/heads/main' && secrets.CF_API_TOKEN != null
+    if: github.event_name == 'pull_request' && github.ref != 'refs/heads/main'
     runs-on: ubuntu-latest
     needs: test
     steps:
--- a/.github/workflows/api.yml
+++ b/.github/workflows/api.yml
@@ -38,9 +38,11 @@ jobs:
           node-version: '16'
       - uses: bahmutov/npm-install@v1
       - name: Publish app
+        if: env.CF_API_TOKEN
         uses: cloudflare/wrangler-action@1.3.0
         env:
           SENTRY_TOKEN: ${{secrets.SENTRY_TOKEN}}
+          CF_API_TOKEN: ${{secrets.CF_API_TOKEN}}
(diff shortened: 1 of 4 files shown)


olizilla:
@gobengo CI is failing on this PR with

```
Invalid workflow file: .github/workflows/api.yml#L31
The workflow is not valid. .github/workflows/api.yml (Line: 31, Col: 9): Unrecognized named-value: 'secrets'. Located at position 75 within expression: github.event_name == 'pull_request' && github.ref != 'refs/heads/main' && secrets.CF_API_TOKEN != null
```

It looks like this was introduced with https://github.com/nftstorage/nft.storage/pull/1698

vasco-santos (commented):
Cluster related changes look good. secrets change still doesn't make as it looks like dev deploy fails. I will check and see if understand why this `if` is in use here

gobengo:
@olizilla big apologies! I should have tested all the workflows since I touched them. Is there a better way than briefly updating some file in each project?

Here is another approach that might work? https://github.com/nftstorage/nft.storage/pull/1790/files#diff-45219019a341aa9fbd159d7d59f842ce8871e83eba709153b2d2c53769673d39R41

@vasco-santos if you're still looking, wdyt of that?

gobengo:
> I will check and see if understand why this if is in use here

First, I think the existing secret.CF_API_TOKEN is defined in this repo but erroring against CF API. i.e. it's potentially invalid (or CF API is rejecting it, but that seems less likely).

Second, I added the conditional to make PRs pass this check even if coming from forks that don't have that variable defined.

gobengo:
@olizilla I think merging [That PR](https://github.com/nftstorage/nft.storage/pull/1793), which is into the source branch of this pr, might make it pass.

PR #1790, opened 2022-04-08
ci(api): remove secrets check in if: in github api workflow

gobengo:


packages/api/NONCE
```diff
@@ -0,0 +1 @@
+test
```
gobengo, line 1: Should remove this before merge, but it's required to trigger the api workflow path selector

PR #1802, opened 2022-04-11
fix: update data-stored image path

### LINKED ISSUE #1782: Fix regression on /stats page image
Design calls for 2 different images for the stat cards on the /stats page.  

- [ ] update the right card (Data Stored" with the following image (Be sure to minify before uploading)
data-stored-svgo-out

## Design for reference
Screen Shot 2022-03-18 at 8 10 54 AM

johnathonroach:
Fixes #1782
This PR updates the image path for the "Data Stored" section of the `/stats` page. The image asset was already in place so only a path update was needed.

### DIFF
--- a/packages/website/pages/stats.js
+++ b/packages/website/pages/stats.js
@@ -142,7 +142,7 @@ export default function Stats({ logos }) {
             <StatCard title="Data Stored">
               <div>
                 <Img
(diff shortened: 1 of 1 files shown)


alanshaw (approved):

Codigo-Fuentes:
**suggestion:** Update metadata so card generates with stats image.

PR #1828, opened 2022-04-14
feat: update 404 page with sweet graphic

### LINKED ISSUE #1729: Navigating to website path that doesn't exist does not return 404
Navigating to a part of the website that doesn't exist just takes you to the homepage, sometimes with some images broken. Let's get a 404 page up - can be a really simple design - I'll ask Protocol Labs design for one, but think we can start doing some digging on what is needed to get it up (something feels buggy about how the redirects are currently happening)

johnathonroach:
Fixes #1729 
This PR updates the 404 page with styles and a 404 image gif.


cmunns (changes requested):
I think we should try not to mix inline styles with classes. We can convert the inline styles to tailwind equivalent, even the arbitrary values as well:
https://tailwindcss.com/docs/max-height#arbitrary-values

dashcraft (changes requested):
The 404 page works but the router doesn't use the 404 page when a route falls through? Is this part of the ticket but the routing is coming later?

I can pop back in whenever this gets answered and approve

johnathonroach:
@dashcraft it appears from the original request that there were situations where the 404 was not showing "just takes you to the homepage" (https://github.com/nftstorage/nft.storage/issues/1729).

I was unable to get any routes to do that - seemed they all fell through to the 404 as expected.

johnathonroach:
@JeffLowe @dashcraft I've updated the latest on this PR removing any test routing settings used to troubleshoot the routing issues on staging. At this point, I cannot determine what is causing the route issues on the staging and prod environment.

This PR only addresses the styling of the 404 page with the graphic.

What I was able to find:
- All "not found" routes return a 200 from the server (not great for SEO)
- No oddities were found in the routing rules for Cloudflare
- We may need to look at the hosting server if that is a thing or have someone who is more familiar with Cloudflare take a look

JeffLowe:
Ok.  After @dashcraft or @cmunns rereviews, let's get this out.  
Also, let's ticket a "server routing investigation" ticket for the support initiative team.  @johnathonroach let me know what to put in the ticket and I'll make sure it gets into their hands.

dashcraft (approved):
Locally works fine, on server there are known issues we'll need to ticket out. 

G2G!

cmunns (approved):

PR #1840, opened 2022-04-20
fix: api ci

vasco-santos:
API CI is currently broken. This PR reverts back to `yarn.lock` previous to #1701 and reinstalls deps to update only changed ones in package.json from website

Adds yarn lock changes to PR changes to guarantee it is working. Proposal to extend this in follow up PR https://github.com/nftstorage/nft.storage/pull/1839


hugomrdias:
this PR https://github.com/nftstorage/nft.storage/pull/1841 should fix the issue

PR #1845, opened 2022-04-22
feat: issue 1827 storybook init and convert inline styles and classes to tailwind

cmunns:
Implemented base SB configuration, but bulk of the work here was removing named classes and inline styles wherever possible–normalizing Tailwind as much as possible.


johnathonroach (approved):
Great and important work! LGTM!

PR #1867, opened 2022-04-29
Correcting pack import path in token.js

rkgudboy:
Updating import path of pack of ipfs-car in token.js


alanshaw:
@rkgudboy please see https://nft.storage/docs/troubleshooting/#why-am-i-seeing-module-not-found-errors for dealing with this error.

PR #1888, opened 2022-05-03
Allow developers to run API in docker, and select between mutable and immutable modes

redaphid:


yusefnapora:
@redaphid I think I fixed the thing where playwright wants to download chromium every time - there was just a version mismatch between the docker image and our code.

I also got the test to sort of run by ripping out parts of the `pw-test.config.cjs` file, and copying `api/scripts` into the container, since the playwright config was trying to load things from there. The playwright config also tries to load the env file from `../../.env`, which I think doesn't exist in the container, so the tests aren't finding the right vars... That's as far as I got, so I didn't bother pushing up the mess :)

Anyway, it seems like it will be kind of a hassle to write a playwright config that could run both in the container and on the host... maybe we could get away with always running in docker? That might simplify things a bit.

packages/api/package.json
```diff
@@ -7,11 +7,11 @@
...
-    "predev": "./scripts/cli.js db --start && ./scripts/cli.js db-sql --cargo --testing --reset",
-    "dev": "miniflare --watch --debug --env ../../.env",
-    "clean": "./scripts/cli.js db --clean && rm -rf docker/compose",
-    "build": "scripts/cli.js build",
-    "test": "tsc && playwright-test \"test/**/*.spec.js\" --sw src/index.js",
+    "dev": "./scripts/start-dev.sh",
+    "dev:persist": "./scripts/start-dev-persist.sh",
+    "build": "./scripts/cli.js build",
+    "test": "./scripts/start-test.sh ",
+    "test:watch": "nodemon --watch ./src --watch ./test --watch ./scripts --exec ./scripts/start-test.sh",
```
yusefnapora, line 14: `test:watch` is super cool 👍

packages/api/pw-test.config.cjs
```diff
@@ -11,54 +11,34 @@ dotenv.config({
...
 /** @type {import('esbuild').Plugin} */
 const nodeBuiltinsPlugin = {
   name: 'node builtins',
   setup(build) {
     build.onResolve({ filter: /^stream$/ }, () => {
       return { path: require.resolve('readable-stream') }
     })
-
-    build.onResolve({ filter: /^cross-fetch$/ }, () => {
-      return { path: path.resolve(__dirname, 'scripts/fetch.js') }
```
yusefnapora, line 24: @redaphid do we not need this `cross-fetch` thing any more?

packages/api/pw-test.config.cjs
```diff
@@ -11,54 +11,34 @@ dotenv.config({
...
   },
   buildSWConfig: {
-    inject: [
-      path.join(__dirname, './scripts/node-globals.js'),
-      path.join(__dirname, './test/scripts/worker-globals.js'),
-    ],
     plugins: [nodeBuiltinsPlugin],
     define: {
-      DATABASE_URL: JSON.stringify(process.env.DATABASE_URL),
+      DATABASE_URL: JSON.stringify(process.env.PW_DATABASE_URL),
```
yusefnapora, line 36: @redaphid I don't see `PW_DATABASE_URL` defined anywhere. maybe this is left over from local debugging?

yusefnapora (commented):
This is so great 💯 

I took a minute to read over the diff and made a couple comments, but I'll let someone else review "for real" since I might be too close to it to catch stuff at this point 😄

packages/api/pw-test.config.cjs
redaphid: Nope! later playwrights use the native fetch when possible. Just a side effect of going through the code.

packages/api/pw-test.config.cjs
redaphid: ah. Yeah we can remove all these defines now. Good catch!

packages/api/docker/docker-compose.yml
```diff
@@ -1,57 +1,111 @@
...
-    ports:
-      - '127.0.0.1:9094:9094'
-    volumes:
-      - ./compose/cluster0:/data/ipfs-cluster
+  playwright:
+    profiles:
+      - test
+    depends_on:
+      api:
+        condition: service_healthy
```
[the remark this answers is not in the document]
redaphid: It's in the new(ish) `docker compose` cmd :). 

It was useful for making sure the 'sidecar'-esque db init container had exited successfully before continuing.

packages/api/pw-test.config.cjs
hugomrdias: this is not related to playwright, its related to [postgrest-js](https://github.com/supabase/postgrest-js) and cross-fetch is still in there (https://github.com/supabase/postgrest-js/blob/master/src/lib/types.ts#L1) even if the code may use web fetch so this is still use full because it remove cross-fetch from the bundle.

hugomrdias:
When we talked about this, we discussed improving the current docker compose situation and the database persistence for dev and/or tests. 
This PR goes away beyond that by running miniflare and playwright-test inside docker i would like to not do that, its way easier to run, debug and test miniflare/pw-test outside docker. This would make pw-test --debug impossible.

On future PRs like this, its probably better to split into smaller scopes and smaller PRs so we can all review step by step, smaller improvements.

redaphid:
@hugomrdias We incorporated your feedback :). Mind looking this over again when you have the chance?

packages/api/docker/docker-compose.yml
```diff
@@ -1,57 +1,126 @@
...
-      PGRST_DB_URI: postgres://postgres:postgres@db:5432/postgres
+      PGRST_DB_URI: $DATABASE_CONNECTION
       PGRST_DB_SCHEMAS: public,cargo
       PGRST_DB_ANON_ROLE: postgres
       PGRST_JWT_SECRET: super-secret-jwt-token-with-at-least-32-characters-long
+  db-init:
+    build:
+      dockerfile: './docker/Dockerfile'
+      context: '../'
+    entrypoint: 'psql $DATABASE_CONNECTION -v "ON_ERROR_STOP=1" -f ./db/config.sql -f ./db/tables.sql -f ./db/cargo.testing.sql -f ./db/functions.sql'
```
vasco-santos, line 15: I think we should have a init.sql script here. If we add a new sql file that needs to run, it will be difficult for the developer to understand the multiple places where it should be added. So, having a init.sql that would execute the other files would be rad. Not super sure how to do it, but found [this](https://stackoverflow.com/a/40667907)

packages/api/docker/postgres/Dockerfile
```diff
@@ -1,13 +1,7 @@
-FROM supabase/postgres:13.3.0
+FROM supabase/postgres:12.5.0
```
vasco-santos, line 1: Why are we downgrading?

packages/api/pw-test.config.cjs
vasco-santos: let's keep it to remove cross-fetch from bundle. And probably nice to add a comment on why that is needed aka postgrest-js to remove from the bundle

packages/api/package.json
```diff
@@ -7,11 +7,13 @@
...
-    "dev": "miniflare --watch --debug --env ../../.env",
-    "clean": "./scripts/cli.js db --clean && rm -rf docker/compose",
-    "build": "scripts/cli.js build",
-    "test": "tsc && playwright-test \"test/**/*.spec.js\" --sw src/index.js",
+    "dev": "./scripts/start-dev.sh",
+    "dev:persist": "./scripts/start-dev-persist.sh",
+    "dev:clean": "./scripts/clean-dev-persist.sh",
+    "build": "./scripts/cli.js build",
+    "test": "./scripts/start-test-local.sh",
+    "test:docker": "./scripts/start-test.sh ",
```
vasco-santos, line 15:
```suggestion
    "test": "./scripts/start-test.sh",
    "test:docker": "./scripts/start-test-docker.sh ",
```

Can we keep same namings to avoid confusion?  We would need shell script names renamed

vasco-santos (commented):
See that we had a lot of back and forth here. Apologies if I added some questions in comments that were already discussed :)

I like the overall direction here, I like to have both options of using docker and local environment for testing. Left some minor suggestions so that we can get this merged in

packages/api/docker/docker-compose.yml
yusefnapora: That's a good idea, especially if we start having fancier migrations - we can just hook them into the init script.

.github/workflows/api.yml
```diff
@@ -21,33 +21,8 @@ jobs:
...
-        if: env.CF_API_TOKEN
-        uses: cloudflare/wrangler-action@1.3.0
-        env:
-          SENTRY_TOKEN: ${{secrets.SENTRY_TOKEN}}
-          CF_API_TOKEN: ${{secrets.CF_API_TOKEN}}
-        with:
-          apiToken: ${{secrets.CF_API_TOKEN }}
-          workingDirectory: 'packages/api'
+      - run: 'cp .env.tpl .env'
+      - run: yarn test:api:docker
```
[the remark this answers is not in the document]
yusefnapora: It's possible that @redaphid and I were confused about deploy-dev. We both thought it was a remnant of the local-tunnel days, where you had to deploy to cloudflare in order to run the dev environment at all.

If it's still useful as an ad-hoc "not quite staging" env we can bring it back.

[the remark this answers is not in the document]
vasco-santos: > If it's still useful as an ad-hoc "not quite staging" env we can bring it back.

It is, but it will be re-written with new commits in other PRs which is annoying. Would love to be able to trigger it in CI for a specific branch, but I think we won't be able to provide a branch on a manual workflow trigger

redaphid:
@vasco-santos I've incorporated all your suggestions! Lmk if this works for you.

packages/api/docker/postgres/Dockerfile
```diff
@@ -1,14 +1,8 @@
-FROM supabase/postgres:13.3.0
+FROM  supabase/postgres:13.3.0
 
 COPY 00-initial-schema.sql /docker-entrypoint-initdb.d/00-initial-schema.sql
```
hugomrdias, line 3: we should add the other sql files here

.github/workflows/api.yml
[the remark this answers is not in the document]
hugomrdias: yes please keep the deploy-dev job and also this yaml is not valid at all i think.

hugomrdias (changes requested):
Can we please remove all the playwright and miniflare out of docker and focus on the DB.
Also please avoid any formatting not enforced by the repo config to make PRs simpler to review.

vasco-santos (changes requested):
We should update this PR to:

1. Scope only development setup, let's have test setup from https://github.com/nftstorage/nft.storage/pull/2065
2. In the API package we are building the API, so let's not dockerize the API (either Miniflare or Playwright) and dockerize only external services, i.e Cluster, DB, ...

PR #1898, opened 2022-05-09
feat: return unique error message when using a blocked key

### LINKED ISSUE #1382: API users with blocked API key should receive custom error message
When users with blocked keys use API they should get a specific error message instead of a generic message. It should inform them the API key is blocked, and to use another api key.

e-schneid:
When an API key is blocked in admin.storage, it sets deleted_at to the current time, and then sets the last auth_key_history item to have status 'Blocked.'  This PR just updates the code to look at the status so it can send a unique error to the client when an API key is deleted via a block.

Closes #1382


JeffLowe:
@jsdevel this one is ready to review

jsdevel (approved):

PR #1901, opened 2022-05-10
fix: invalid multipart data error

alanshaw:
[We require a `Content-Type` to be set for each part of the multipart request.](https://nft.storage/api-docs/#operations-NFT_Storage-upload)

If users don't set this value `FormData#getAll()` returns a `string` not a `File`.

This PR prevents Sentry errors for invalid multipart (by our standards).

Screenshot 2022-05-10 at 10 38 41


vasco-santos (approved):
lgtm

PR #1914, opened 2022-05-16
chore: refactor service configuration loading

yusefnapora:
@gobengo, @redaphid & I were talking a few days ago about the configuration loading in the api and how we all find `constants.js` to be kind of awkward. 

So this is an attempt to formalize the configuration loading a bit and build up a `ServiceConfiguration` object by looking for globals and falling back to a default for e.g. unit testing.

I recorded a little [screencast of me talking through the code](https://bafybeiefuwx4cejp3ifniqoomlmotcrx2ycnn67ao2x3athr3sziiza7da.ipfs.nftstorage.link/2022-05-16%20-%20nft%20storage%20config%20object.mp4), which might be helpful to get an idea of what's changed.

Key changes:

- constants.js is replaced with config.js
- config.js exports `getServiceConfiguration`, which lazily loads the config into a JS `ServiceConfiguration` object
- default values for unit tests are defined in code, in config.js
- runtime values are dynamically looked up on `globalThis`, to avoid undefined reference errors if a var is missing.
- if a config variable doesn't exist at runtime, we log a warning and use the default value

All existing tests pass, but the config code itself isn't tested yet.

Leaving as a draft until tests & discussion :)


packages/api/src/config.js
```diff
@@ -0,0 +1,231 @@
...
+}
+
+/**
+ * Returns `true` if the string `s` is (case insensitive) equal to `"true"`, or `false` for any other value.
+ *
+ * @param {string} s
+ * @returns {boolean}
+ */
+function boolValue(s) {
+  return s.toLowerCase() === 'true'
```
gobengo, line 230: wdyt?

```suggestion
  return  Boolean(JSON.parse(s.toLowerCase()))
```

gobengo:
lgtm

packages/api/src/config.js
yusefnapora: oh nice :)

yusefnapora:
I think this is ready to review, although the `VERSION` workaround is not great. If anyone has a less-bad workaround we could do that, or else we can change ucan-storage to stop setting the global.

yusefnapora:
Made a tiny PR to rename the constant in ucan.storage, so we can get rid of the hack here: https://github.com/nftstorage/ucan.storage/pull/28

packages/api/src/config.js
```diff
@@ -0,0 +1,321 @@
...
+  // to have a defined value. In dev & test, we allow falling back to default config values.
+  const env = runtimeEnvFromString(vars['ENV'])
+  if (allowDefaultConfigValues(env)) {
+    mergeWithDefaultValues(vars, env)
+
+    // special case for VERSION, because ucan-storage defines a VERSION global
+    // that seems to be clobbering ours if no VERSION var is defined by cloudflare.
+    if (vars['VERSION'] === '0.8.0') {
+      vars['VERSION'] = DEFAULT_CONFIG_VALUES.VERSION
+    }
```
hugomrdias, line 224: this happens because of this https://github.com/nftstorage/nft.storage/blob/main/packages/api/scripts/cli.js#L61-L63 we should really namespace the those `NFT_STORAGE_VERSION` etc

packages/api/src/config.js
```diff
@@ -0,0 +1,321 @@
...
+
+  /** @type Record<string, unknown> */
+  const globals = globalThis
+
+  for (const name of Object.keys(DEFAULT_CONFIG_VALUES)) {
+    const val = globals[name]
+    if (typeof val === 'string') {
+      vars[name] = val
+    }
+  }
```
hugomrdias, line 197: I think we should do more in this for loop and remove some of the other functions.

Merging should happen right here for `test` and `dev`, fail for the other env. Basically the some logic overall just in one place, makes it easier to flow.

hugomrdias: Also we avoid iterating multiple times over the config keys and we can make it all static and return one single object from this file.

packages/api/src/config.js
```diff
@@ -0,0 +1,321 @@
...
+  let clusterUrl = configVars.CLUSTER_API_URL
+  if (configVars.CLUSTER_SERVICE) {
+    const serviceUrl = CLUSTER_SERVICE_URLS[configVars.CLUSTER_SERVICE]
+    if (!serviceUrl) {
+      throw new Error(`unknown cluster service: ${configVars.CLUSTER_SERVICE}`)
+    }
+    clusterUrl = serviceUrl
+  }
+
+  return {
```
hugomrdias, line 122: since we are changing things, this object structure always felt wrong for config constants with all the nesting etc... we had reasons for this but not anymore.

my suggestion would be flatten it just like the DEFAULT_CONFIG_VALUES object

packages/api/src/index.js
```diff
@@ -25,19 +25,19 @@ import { userDIDRegister } from './routes/user-did-register.js'
...
   setMaintenanceModeGetter,
 } from './middleware/maintenance.js'
 import { getContext } from './utils/context.js'
 import { withAuth } from './middleware/auth.js'
 
-const getMaintenanceMode = () =>
-  typeof MAINTENANCE_MODE !== 'undefined' ? MAINTENANCE_MODE : DEFAULT_MODE
+const getMaintenanceMode = () => getServiceConfig().maintenanceMode
+
 setMaintenanceModeGetter(getMaintenanceMode)
```
hugomrdias, line 41: We dont need this workaround of getter and setter, this was due to a bug in `wrangler dev` that didnt load the injected globals properly. 

The maintenance middleware can access the config directly just like any other injected global.

hugomrdias: This is just a thing we never clean up not super related to this PR, so you can ignore it if you dont feel like changing it.

packages/api/src/config.js
hugomrdias: Also it would be easier to spot config globals, if we could import config and do `config.S3_ENDPOINT` instead of `config.s3.endpoint` or whatever we do now.

hugomrdias (changes requested):
Love the general direction of this PR !! Left some comments and suggestions.

packages/api/src/config.js
yusefnapora: Sweet, that makes things quite a bit simpler 😄It sounds like the final config object will look just like `DEFAULT_CONFIG_VALUES`, plus a little bit of type validation (string to bool, validating maintenance mode, etc).

packages/api/src/index.js
yusefnapora: Oh sweet, I was a bit confused by that and meant to ask you but forgot :) I'll pull it out.

packages/api/src/config.js
yusefnapora: That's what I thought too, but that would set version to a more complex string: `${pkg.name}@${pkg.version}-${opts.env}+${git.short(__dirname)}`

Here's what shows up in `dist.js`:

image

Also, if I manually apply the edit from [here](https://github.com/nftstorage/ucan.storage/pull/28) to the copy `ucan-storage` inside of `node_modules` (or yarn link that branch), the version test passes again.

yusefnapora:
Thanks for the review, @hugomrdias.  I left in the validation and boolean parsing, but I took out all the nested structure from the config object. It does feel much simpler now, and I like that you don't have to keep two names in your head (`runtimeEnvironment` vs `ENV`).

I took out the hard-coded `VERSION` hack, so that test is failing again (see comment above). It works with my ucan-storage PR branch, but I still don't understand why the collision is happening in the first place... 😕

hugomrdias:
> Thanks for the review, @hugomrdias. I left in the validation and boolean parsing, but I took out all the nested structure from the config object. It does feel much simpler now, and I like that you don't have to keep two names in your head (`runtimeEnvironment` vs `ENV`).
> 
> I took out the hard-coded `VERSION` hack, so that test is failing again (see comment above). It works with my ucan-storage PR branch, but I still don't understand why the collision is happening in the first place... 😕

Just change this https://github.com/nftstorage/nft.storage/blob/main/packages/api/scripts/cli.js#L61-L63 to : 

```js
 define: {
    NFT_STORAGE_VERSION: JSON.stringify(version),
    NFT_STORAGE_COMMITHASH: JSON.stringify(git.long(__dirname)),
    NFT_STORAGE_BRANCH: JSON.stringify(git.branch(__dirname)),
    global: 'globalThis',
  },
```
And do the necessary changes in this PR for the new global var names

The problem is here https://github.com/nftstorage/ucan.storage/blob/main/src/ucan.js#L82 not https://github.com/nftstorage/ucan.storage/blob/main/src/ucan.js#L12 .

packages/api/src/config.js
```diff
@@ -0,0 +1,237 @@
...
+export const getServiceConfig = () => {
+  if (_globalConfig) {
+    return _globalConfig
+  }
+  _globalConfig = loadServiceConfig()
+  return _globalConfig
+}
+
+/** @type ServiceConfiguration|undefined */
+let _globalConfig
```
hugomrdias, line 75: can we move this up before the `getServiceConfig` ?

hugomrdias (changes requested):
lets looks perfect !!! love it so much better
left one comment inline 

after namespacing the git globals this can be merged IMO

yusefnapora:
Sweet, thanks @hugomrdias. Should be all good now if you want to take another look.

hugomrdias (changes requested):
LGTM just some leftover commented code

packages/api/src/bindings.d.ts
```diff
@@ -7,32 +7,130 @@ import { UserOutput, UserOutputKey } from './utils/db-client-types.js'
...
+//     sentry: {
+//       dsn: string
+//     }
+
+//     mailchimp: {
+//       apiKey: string
+//     }
+//   }
+// }
+
```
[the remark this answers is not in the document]
yusefnapora: ah thanks for catching that!

hugomrdias (approved):

vasco-santos:
We are having an issue from this PR on the Environment variables provided to the API
https://github.com/nftstorage/nft.storage/runs/6694458301?check_suite_focus=true#step:6:25

Issue #1525, opened 2022-03-01
Docs storeDirectory example using async iterator filesFromPath


dchoi27:
Mentioned in #1523, but it'd be great to add an example in docs for storing a directory using async iterator filesFromPath

redaphid:
This would be suuuuuper helpful for me. I was trying to do this for days a few months back.

gobengo:
IMO the docs of this would be prettier if we allowed passing not just an `Iterable<File>` but `AsyncIterable<File>`. https://github.com/nftstorage/nft.storage/blob/3003926c399be163fd09b3e4788699ac8f850f61/packages/client/src/lib.js#L458

gobengo:
but for huge directories/files, this will also lead to huge CAR files. In the future it may be desirable to support emitting not just one car but an `AsyncIterable<Car>` that also tries to keep each CAR below a max size.

gobengo:
I see this live on https://nft.storage/docs/how-to/store-directory/#adding-the-upload-code

I think this is done. @dchoi27 agree?

gobengo:
Considering this done

PR #1918, opened 2022-05-17
Fix typo in how-to/store-directory code snippet

gobengo:
Motivation:
* I was diving into https://github.com/nftstorage/nft.storage/issues/1525 and started by following the docs as-is. When I ran the script, I got an error:
```
TypeError: paths is not async iterable
    at filesFromPath (file:///Users/bengo/protocol.ai/store-directory/node_modules/files-from-path/esm/src/index.js:27:28)
```
  * it's because the code snippet was passing the node stdlib 'path' module where I believe the function wanted one or more filesystem paths as strings


yusefnapora (approved):
thanks @gobengo!

PR #1920, opened 2022-05-18
feat: storeDirectory accepts files as AsyncIterable<File>

gobengo:
Motivation:
* https://github.com/nftstorage/nft.storage/issues/1525#issuecomment-1129357723
* this allows end-users to defer buffering of large sets of files (e.g. large dirs), and just pass us an `AsyncIterable<File>`. If I'm reading the `packCar` code right, I think even we avoid buffering the whole file list in an array (though Blockstore may still do the equivalent)


yusefnapora (approved):
This looks great to me, and makes it way nicer to actually implement the docs change.

I think you're right about `packCar` being fine with an async iterable input, so this seems like it should work.

PR #1921, opened 2022-05-18
docs: Update docs/how-to/store-directory to pass files as AsyncIterable<File>

gobengo:
Depends on
  * [x] https://github.com/nftstorage/nft.storage/pull/1920
  * [x] release of `nft.storage` client to npm so these docs work on latest

Motivation
* make the documented way of doing this not buffer big dirs in a way that could crash https://github.com/nftstorage/nft.storage/issues/1525

### DIFF
--- a/packages/website/pages/docs/how-to/store-directory.md
+++ b/packages/website/pages/docs/how-to/store-directory.md
@@ -65,9 +65,9 @@ main()
 
 The `filesFromPath` function is provided by the [`files-from-path` package][npm-files-from-path]. It will read the contents of a directory into `File` objects that can be passed into the NFT.Storage client.
 
-The `pathPrefix` option tells `getFilesFromPath` to remove the input `path` from the filenames of the `File` objects it creates. For example, if you're reading in files from a directory called `example`, calling `getFilesFromPath` _without_ the `pathPrefix` argument would result in `File` objects with filenames like `example/file1.txt`, `example/file2.txt`, and so on. If you set the `pathPrefix` option to `example`, you'll get `file1.txt`, `file2.txt`, etc. instead. This results in a final IPFS URI of `ipfs://<directory-cid>/file1.txt` instead of `ipfs://<directory-cid>/example/file1.txt`.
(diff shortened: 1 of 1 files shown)


packages/website/pages/docs/how-to/store-directory.md
```diff
@@ -42,14 +42,18 @@ async function main() {
...
   const directoryPath = process.argv[2]
-  const files = await getFilesFromPath(path, {
+  const files = (async function* () {
+    for await (const file of filesFromPath(directoryPath, {
       pathPrefix: path.resolve(directoryPath), // see the note about pathPrefix below
-      hidden: true // use the default of false if you want to ignore files that start with '.'
-  })
+      hidden: true, // use the default of false if you want to ignore files that start with '.'
+    })) {
+      yield new File([file.stream()], file.name);
```
gobengo, line 50: constructing this explicitly isn't required for the js to work, but it is required for ts auto inference to work.

If we want to simplify this further, we could update `NFTStorage.storeDirectory` to accept `IterableOrAsyncIterable<FileObject|File>` (not just `File` but `FileObject` too).

It can also look similar by just not caring about ts auto inference, but IMO it's nice to be ready for a dev to not want to have to manually grok type errors

yusefnapora: If `storeDirectory` accepted `FileObject`, can we just feed it the output of `filesFromPath` directly? If so, that seems worth it. I've been confused by this same type error before, and it would be nice to sort it out for good.

gobengo: Yes

gobengo: @yusefnapora implemented in https://github.com/nftstorage/nft.storage/pull/1924

gobengo:
i tested the code snippet against the newly released nft.storage@6.3.0

yusefnapora (approved):
This looks awesome, thanks @gobengo

PR #1924, opened 2022-05-18
feat: storeDirectory accepts iter of file objects

gobengo:
Motivation:
* https://github.com/nftstorage/nft.storage/pull/1921#discussion_r876243994
* this makes it so storeDirectory can accept POJO, not just dom File objects, which can make our docs code snippets even simpler when using `files-from-path`

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -675,7 +675,7 @@ class NFTStorage {
 
 /**
  * Cast an iterable to an asyncIterable
- * @template {File|FileObject} T
+ * @template T
  * @param {Iterable<T>} iterable
  * @returns {AsyncIterable<T>}
  */
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -25,6 +25,7 @@ import { fetch, File, Blob, FormData, Blockstore } from './platform.js'
 import { toGatewayURL } from './gateway.js'
 import { BlockstoreCarReader } from './bs-car-reader.js'
 import pipe from 'it-pipe'
+import all from 'it-all'
 
 const MAX_STORE_RETRIES = 5
 const MAX_CONCURRENT_UPLOADS = 3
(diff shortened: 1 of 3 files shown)


packages/client/src/lib.js
```diff
@@ -216,13 +219,28 @@ class NFTStorage {
...
+      async function* () {
+        yield* filesSource
+      },
+      async function* (files) {
+        for await (const fileObject of files) {
+          /** @type {File} */
+          const file =
+            fileObject instanceof File
+              ? fileObject
+              : new File(await all(fileObject.stream()), fileObject.name)
```
yusefnapora, line 238: Do we need to `await` the stream here? It would be cool if we could just pass in `[fileObject.stream()]`

yusefnapora: I think that `toImportCandidate` inside of `encodeDirectory` will accept anything with a `stream()` method that returns an async iterable. So you may be able to skip the coercion to `File` here and just pass everything through.

Unless you tried that already and it didn't work, lol

gobengo: it wasn't so simple. I didn't want to augment more functions than needed, but insofar as somethings were typed as wanting `File`, this required buffering. afaict constructing a dom `File` requires not just a stream but a fully-read `Uint8Array`.

I just pushed a change though that allows the `FileObject` to `encodeDirectory` and makes this code less ugly.

packages/client/src/lib.js
```diff
@@ -758,10 +747,11 @@ const decodePin = (pin) => ({ ...pin, created: new Date(pin.created) })
  * the stream is created only when needed.
  *
  * @param {string} path
- * @param {Blob} blob
+ * @param {Pick<Blob, 'stream'>|{ stream: () => AsyncIterable<Uint8Array> }} blob
```
yusefnapora, line 750: I really like this type definition 😄 - I should start using `Pick` and `Omit` more...

packages/client/src/lib.js
```diff
@@ -758,10 +747,11 @@ const decodePin = (pin) => ({ ...pin, created: new Date(pin.created) })
...
+ * @returns {import('ipfs-core-types/src/utils.js').ImportCandidate}
  */
 function toImportCandidate(path, blob) {
-  /** @type {ReadableStream} */
+  /** @type {AsyncIterable<Uint8Array>} */
```
yusefnapora, line 754: nice

yusefnapora (approved):
beautiful 💯

packages/client/src/lib.js
gobengo: yeah :) it would be cool if there was something like an eslint rule that helped you specify your inputs as 'narrowly'/minimally as possible. e.g. in this case we didn't really *need* a whole `Blob`, just a subset

packages/client/src/lib.js
gobengo: yeah its so cool how node made stream.Readables implement AsyncIterable

PR #1932, opened 2022-05-20
feat: cron release workflow

### LINKED ISSUE #1931: Cron Package - Continuous Deployment
## Context
Cron code that lands into main is automatically in production.
It's unexpected since that is not the case for the rest of the packages in the repo and that leads to nasty bugs in prod.

ie, when code in cron jobs requires changes in the DB, or updated APIs
This [PR](https://github.com/web3-storage/web3.storage/pull/1088) is a good example

## Possible solution
We could add a release step for the cron package, leveraging release-please as we do in the other packages is likely a good way forward.

Reporting some initial thoughts from different places: 
> We could maybe do something where cron is included in a release please process and the yaml is copied (and committed) from the project dir to the root on release?

flea89:
Fixes #1931 

Port of web3.storage [one](https://github.com/web3-storage/web3.storage/pull/1281/files)

### DIFF
--- a/.github/workflows/cron-dagcargo-sizes.yml
+++ b/.github/workflows/cron-dagcargo-sizes.yml
@@ -15,6 +15,12 @@ jobs:
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Checkout latest cron release tag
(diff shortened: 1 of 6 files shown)


alanshaw (approved):
🎉

PR #1936, opened 2022-05-23
fix: default PSA pin status filter

### LINKED ISSUE #1933: Pinning Services API 'GET' example is broken
Not sure if this is a documentation or code issue, but running the [Pinning Services API 'GET' example](https://staging.nft.storage/docs/how-to/pinning-service/#list-successful-pins) to list successful pins fails with the following result:

`{"reason":"VALIDATION_ERROR","details":[{"instanceLocation":"#","keyword":"required","keywordLocation":"#/required","error":"Instance does not have required property \"status\"."}]}`

alanshaw:
`status` filter is not a required param, it should be implicity set to 'pinned' when not set by the client.

resolves #1933

cc @flea89

### DIFF
--- a/packages/api/src/routes/pins-list.js
+++ b/packages/api/src/routes/pins-list.js
@@ -123,6 +123,10 @@ function parseSearchParams(params) {
     out.status = /** @type {ListUploadsOptions["status"]}*/ (
       statusParam.split(',').map(toDbPinStatus)
     )
+  } else {
+    // "when no filter is provided, only successful pins are returned"
+    // https://ipfs.github.io/pinning-services-api-spec/#tag/pins/paths/~1pins/get
+    out.status = ['Pinned']
   }
 
   const afterParam = params.get('after')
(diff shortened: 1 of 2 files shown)


flea89:
@alanshaw I remember some discussion happened on this exact topic (I wasn't directly involved).

In this [PR](https://github.com/web3-storage/web3.storage/pull/937), specifically in [this comment](https://github.com/web3-storage/web3.storage/pull/937#discussion_r804708026) we were advised to make it required.

alanshaw:
I checked this with @lidel - this is the required behaviour.

alanshaw:
> @alanshaw I remember some discussion happened on this exact topic (I wasn't directly involved).
> 
> In this [PR](https://github.com/web3-storage/web3.storage/pull/937), specifically in [this comment](https://github.com/web3-storage/web3.storage/pull/937#discussion_r804708026) we were advised to make it required.
> 
> That said, what you're suggesting here is what we assumed by reading the [documentantion](https://ipfs.github.io/pinning-services-api-spec/#tag/pins/paths/~1pins/get), and IIRC how we implemented it in the first place. By having a quick look to the spec as well it doesn't specify the field as `required` and I can't find what is been referred to as `non-empty`, but I might be looking in the wrong place.

@flea89 yeah...reading that old comment it seems I might have misinterpreted what was meant. I will double check.

Issue #1939, opened 2022-05-24
Backup URLs not properly stored


vasco-santos:
We have a regression in nftstorage backups from postgres optimization PR https://github.com/nftstorage/nft.storage/pull/1664

Given we do chunked uploads, we used to rely on multiple calls to createUpload to add new rows to Backup table. We moved to a `text[]` column in uploads table, which is not updated with previous URLs and just contains last call to upload.

We aligned on a solution for this at https://github.com/web3-storage/web3.storage/pull/1305 and we will need to:

1. Port fix from web3.storage implementation
2. Create a script that uses S3 API to get all the URLs for each rootCid/userId path and add them to the Uploads
  - We need to validate if we only need this after postgres optimization PR was merged, or if the migration script also not included all the backup URLs

PR #1940, opened 2022-05-24
fix: preserve all backup_urls for chunked uploads

yusefnapora:
This stops the bleeding on #1939 by making sure we append to `upload.backup_urls` instead of overwriting when updating an existing entry.

Once this is in, we'll need to do the migration to backfill the missing backup_urls from the time after #1664 was merged. I've started working on that but likely won't finish today.


packages/api/test/nfts-upload.spec.js
```diff
@@ -375,22 +379,75 @@ describe('NFT Upload ', () => {
...
+        path: `/dir/file-${i}.bin`,
+        content: getRandomBytes(chunkSize),
+      })
+    }
+
+    const { root, car } = await packToBlob({
+      input: files,
+      maxChunkSize: chunkSize,
+    })
+    console.log('chunked car root cid:', root.toString())
```
vasco-santos, line 406:
```suggestion
```

vasco-santos (changes requested):
Overall looking good, thanks @yusefnapora 
I guess we don't need `s3-scrape-backup-urls.js` file right?

vasco-santos (approved):
LGTM

hugomrdias (approved):

PR #1941, opened 2022-05-24
feat: use application/vnd.ipld.car as preferred content-type

### LINKED ISSUE #1758: Use IANA media type application/vnd.ipld.car
See https://github.com/ipfs/in-web-browsers/issues/192#issuecomment-1087506324

yusefnapora:
This closes #1758 by making `application/vnd.ipld.car` the main Content-Type for CAR uploads. The API will still accept `application/car`, but the client will send `application/vnd.ipld.car`.

**Important**: since this changes the client behavior, we need to deploy the API before publishing a new client version.

### DIFF
--- a/packages/api/src/routes/metaplex-upload.js
+++ b/packages/api/src/routes/metaplex-upload.js
@@ -52,7 +52,7 @@ export async function metaplexUpload(event, ctx) {
       user,
       key,
       car: blob,
-      mimeType: 'application/car',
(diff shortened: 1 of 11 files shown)


lidel (approved):
lgtm, https://www.iana.org/assignments/media-types/application/vnd.ipld.car

@yusefnapora anything blocking you from merging this?

Issue #1943, opened 2022-05-24
continuously measure 'time to retrievability' to establish a baseline for future improvement experiments


gobengo:
Motivation:
* answer Q I've been asked: "How long after writing data to nft.storage can I retrieve it via IPFS?"
* answer Q I always worry about: "How did this code change I'm proposing/deploying affect the important KPIs?" (e.g. did I just release a regression in behavior?)

Inspiration:
* https://sre.google/sre-book/monitoring-distributed-systems/
* https://sre.google/sre-book/testing-reliability/

Overview
* add a test that does
  * generate `> carMaxFileSize` bytes (e.g. 11Mb) of image data to upload - this is so that it triggers the client to split the upload to >1 car
  * measure upload speed during
    * use `nft.storage` client to upload these files to prod nft.storage
  * for each of knownGateways
    * measure this as time to retrievability
      * loop trying to retrieve the uploaded cid from the knownGateway
* create github action to measure and report these results on a regular interval

Decisions
* code where?
  * https://github.com/nftstorage/nft.storage/tree/main/packages/cron

Parts
* [ ] uploading and measuring without fancy data generation
* [ ] generate data to be uploaded so its large enough to trigger a carbites split in the client

alanshaw:
Some thoughts (not requirements):

* It would be nice to be uploading some useful data rather than just junk
* A typical NFT use case 10k drops
    * Folks typically upload a directory with 10k JSON files and also 10k of images (the files must not be the same or deduplication will occur)
* We don't _have_ to upload 10MB every time. In the Web3.Storage client we expose the DAG splitting size as a param to `put`. In NFT.Storage you could use carbites to split the DAG by a smaller size and then call `storeCar` multiple times.
* When available ([coming in v0.13](https://github.com/ipfs/go-ipfs/issues/8640)), you might want to use the `?format=car` parameter on the gateways to verify ALL the blocks of the DAG are accessible.

gobengo:
#status I've been looping trying to debug and fix why stats aren't showing up in grafana. Digging in more today and may need pairing to help brainstorm why it's not working if this latest theory doesn't pan out

gobengo:
This is working now.

next https://github.com/nftstorage/nft.storage/issues/1994

PR #1945, opened 2022-05-25
feat: create cron nft-ttr to measure nft time to retrievability

gobengo:
Motivation:
* https://github.com/nftstorage/nft.storage/issues/1943

Plan

1. note all feedback
    * dont use jest for cron tests. use ava
    * dont use yargs for argv parsing, use `"sade": "^1.7.4"`
    * use less fields on nft-ttr metric [via](https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1654044830889049?thread_ts=1653585353.059339&cid=C02BZPRS9HP)
    * make nft-ttr retrieval metric a histogram instead of a gauge [via](https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1654055779880559?thread_ts=1653585353.059339&cid=C02BZPRS9HP)
    * alan [thoughts](https://github.com/nftstorage/nft.storage/issues/1943#issuecomment-1143650993)
        * It would be nice to be uploading some useful data rather than just junk
        * A typical NFT use case 10k drops
            * Folks typically upload a directory with 10k JSON files and also 10k of images (the files must not be the same or deduplication will occur)
        * We don't have to upload 10MB every time. In the Web3.Storage client we expose the DAG splitting size as a param to put. In NFT.Storage you could use carbites to split the DAG by a smaller size and then call storeCar multiple times.
        * When available (https://github.com/ipfs/go-ipfs/issues/8640), you might want to use the ?format=car parameter on the gateways to verify ALL the blocks of the DAG are accessible.
4. updates based on review/blocking feedback
1. merge this - whatever is required to get approval
2. prioritized other tasks to do while awaiting review approval
    * [x] update this PR to use sade for arg parsing
    * [x] update this PR such that cron package uses ava for test running (no jest)
    * [x] test retrieving from more than one gateway
6. verify once merged
    * crons are running by gh actions once the workflow is described in main branch
    * crons result in metrics successfully pushed to pushgateway
7. followup pr with
    * fixes if the aforementioned verifications fail

Possible Futures
* @alanshaw [requirement 'We should also record the time it took to upload the data.'](https://github.com/nftstorage/nft.storage/pull/1945#discussion_r892340411)
  * https://github.com/nftstorage/nft.storage/issues/1994
* upgrade tsc https://github.com/nftstorage/nft.storage/pull/1945#issuecomment-1155533297
* eslint rest of files and simplify eslint config exceptions https://github.com/nftstorage/nft.storage/pull/1945#discussion_r892271164
* [ ] change nft-ttr so it doesn't generate 10Mb every time, it just generates ~1Mb but triggers at least one car split by manually configuring the client
  * [ ] blocked by https://github.com/nftstorage/nft.storage/issues/1962 
* [ ] propose [ADRs](https://adr.github.io/)
* [ ] propose ADR in nft.storage for using sade for arg parsing
* [ ] propose ADR in nft.storage for using ava as test runner
* add test coverage requirements to cron usage of c8 (e.g. thresholds from [ucanto](https://github.com/web3-storage/ucanto/blob/main/package.json#L26))
* make this job have the nft.storage-specific parts be injectable so same job can be used to observe web3.storage
* disable eslint overrides (which were added to avoid unnecessarily modifying files otherwise not touched by this PR) in packages/cron/package.json `.eslintConfig.overrides`
* add another cronjob to continuously test @alanshaw's [thought](https://github.com/nftstorage/nft.storage/issues/1943#issuecomment-1143650993)
  > * A typical NFT use case 10k drops
  >   * Folks typically upload a directory with 10k JSON files and also 10k of images (the files must not be the same or deduplication will occur)


packages/cron/src/jobs/measureNftTimeToRetrievability.js
```diff
@@ -0,0 +1,215 @@
...
+      type: 'store',
+      image: imageId,
+      startTime: storeStartedDate,
+      duration: new Milliseconds(
+        storeEndAt.toNumber() - storeStartedAt.toNumber()
+      ),
+    }
+    config.log('info', storeLog)
+    const retrievalUrl = new URL(
+      `https://${metadata.ipnft}.ipfs.nftstorage.link/image/blob`
```
gobengo, line 112: eventually should test a handful of gateways, not just this one. perhaps as separate job invocations (e.g. using github actions matrix)

gobengo: https://github.com/nftstorage/nft.storage/pull/1945/commits/176ebeab8fd1a665010290404db039f1b8644661

.github/workflows/cron-nft-ttr.yml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - uses: bahmutov/npm-install@v1
+      - name: Run job
+        env:
+          DEBUG: '*'
+          ENV: ${{ matrix.env }}
+          NFT_STORAGE_API_KEY: ${{ secrets.NFT_STORAGE_API_KEY }}
+          PUSHGATEWAY_BASIC_AUTH: ${{ secrets.PUSHGATEWAY_BASIC_AUTH }}
```
gobengo, line 27: I haven't been able to test the pathway when this is truthy because I don't have a sample secret

gobengo: nvm I found a value in secrets vault and tested locally. then made small fixes until the push metrics request responded with status 200

gobengo:
looks like some of my sample metrics pushes show up [here](https://protocollabs.grafana.net/explore?left=%7B%22datasource%22:%22Prometheus-k8s%22,%22queries%22:%5B%7B%22exemplar%22:true,%22expr%22:%22nftstorage_retrieval_duration_seconds%22,%22interval%22:%22%22,%22queryType%22:%22randomWalk%22,%22refId%22:%22A%22,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22000000044%22%7D%7D%5D,%22range%22:%7B%22from%22:%22now-24h%22,%22to%22:%22now%22%7D%7D&orgId=1)

alanshaw (changes requested):

.github/workflows/cron-nft-ttr.yml
```diff
@@ -0,0 +1,38 @@
...
+          ENV: ${{ matrix.env }}
+          NFT_STORAGE_API_KEY: ${{ secrets.NFT_STORAGE_API_KEY }}
+          PUSHGATEWAY_URL: https://pushgateway.k8s.locotorp.info/metrics/job/nftstorage_ci/instance/github_action
+          PUSHGATEWAY_BASIC_AUTH: ${{ secrets.PUSHGATEWAY_BASIC_AUTH }}
+        run: |
+          yarn workspace --silent cron run --silent start:nft-ttr measure --logConfigAndExit
+          yarn workspace --silent cron run --silent start:nft-ttr measure \
+            --min-image-size-bytes=10000000 \
+            --gateways https://nftstorage.link https://dweb.link \
+            --metricsPushGateway $PUSHGATEWAY_URL \
```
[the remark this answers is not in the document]
gobengo: no reason, but I changed it to be consistent: https://github.com/nftstorage/nft.storage/pull/1945/commits/b7d958b35fd4f7b1b4e727cf57fdc02e1dd33794

packages/cron/src/bin/nft-ttr.spec.js
```diff
@@ -0,0 +1,43 @@
+import { describe, testIf, testsCanUseNetwork } from '../lib/testing'
+import { main as binNftTtr } from './nft-ttr'
+import { recordedLog } from '../lib/log'
+import * as assert from 'node:assert'
+
+const BYTES_10_MB = 10 * 1e6
```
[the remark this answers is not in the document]
gobengo: ok https://github.com/nftstorage/nft.storage/pull/1945/commits/c27d2299b236ba84d0bf142644f5471c04e4fd74

packages/cron/src/jobs/measureNftTimeToRetrievability.js
```diff
@@ -0,0 +1,278 @@
...
+ * @returns {AsyncIterable<Blob>}
+ */
+export function UrlImages(...urls) {
+  return (async function* () {
+    for (const url of urls) {
+      const resp = await fetch(url.toString())
+      yield resp.blob()
+    }
+  })()
+}
```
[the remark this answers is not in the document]
gobengo: > How come these functions begin with uppercase? We usually reserve for classes or react components.

no good reason. I'll use that convention going forward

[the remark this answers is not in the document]
gobengo: also added `@typescript-eslint/naming-convention` warning so lint catches this

packages/cron/src/jobs/measureNftTimeToRetrievability.js
```diff
@@ -0,0 +1,278 @@
...
+ * @param {string} [config.url] - URL to nft.storage to measure
+ * @param {boolean} [config.logConfigAndExit] - if true, log config and exit
+ * @param {string} [config.metricsPushGateway] - Server to send metrics to. should be a https://github.com/prometheus/pushgateway
+ * @param {URL[]} config.gateways - IPFS Gateway to test retrieval from
+ * @param {import('../lib/log.js').LogFunction} config.log - logger
+ * @param {object} secrets
+ * @param {string} secrets.nftStorageToken - API Token for nft.storage
+ * @param {HttpAuthorization} secrets.metricsPushGatewayAuthorization - authorization header value for metricsPushGateway
+ */
+export async function measureNftTimeToRetrievability(config, secrets) {
```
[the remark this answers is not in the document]
gobengo: I like to try to isolate secrets early so they dont actually get logged to ci stdout for all to see.
e.g. I didn't want to afford for users of the script to pass secrets via cli flags, which would then show up in process command. I wanted them to pass using a separate channel from 'config' (e.g. env vars).
Not that attached to having them separate at this layer.
I did something to merge them into a shared options object.

packages/cron/src/jobs/measureNftTimeToRetrievability.js
```diff
@@ -0,0 +1,278 @@
...
+    const nft = {
+      image: /** @type {Blob} */ (image), // use image Blob as `image` field
+      name: 'Example NFT',
+      description: 'Example NFT used by nft.storage cron nft-ttr',
+    }
+    const client = new NFTStorage({
+      token: secrets.nftStorageToken,
+    })
+    const storeStartedDate = new Date()
+    const storeStartedAt = now()
```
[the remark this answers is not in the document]
gobengo: imho passing around an untyped int is more complexity than this because then whoever's dealing with the code later will have to remember the units instead of the compiler doing it for us.

packages/cron/src/jobs/measureNftTimeToRetrievability.js
```diff
@@ -0,0 +1,278 @@
...
+ * @param {RetrieveLog} retrieval
+ * @returns {string} prometheus metrics string
+ */
+function formatRetrievalMetrics(retrieval) {
+  const durationSeconds = retrieval.duration.size / 1000
+  return (
+    [
+      '# TYPE nftstorage_retrieval_duration_seconds gauge',
+      '# HELP nftstorage_retrieval_duration_seconds How long it took to retrieve an nft image',
+      `nftstorage_retrieval_duration_seconds{image="${retrieval.image}",url="${retrieval.url}",size="${retrieval.contentLength}"} ${durationSeconds}`,
```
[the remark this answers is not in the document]
gobengo: I switched to prom-client

packages/cron/src/jobs/measureNftTimeToRetrievability.js
```diff
@@ -0,0 +1,278 @@
...
+      headers: {
+        authorization: config.metricsPushGatewayAuthorization.authorization,
+        'content-type': 'application/octet-stream',
+      },
+      body: formatRetrievalMetrics(retrieval),
+    },
+  ]
+  const pushResponse = await fetch.apply(null, pushRequest)
+  const pushSuccessfulStatusCodes = [200, 201]
+  if (!pushSuccessfulStatusCodes.includes(pushResponse.status)) {
```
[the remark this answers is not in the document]
gobengo: done

packages/cron/src/jobs/measureNftTimeToRetrievability.js
```diff
@@ -0,0 +1,380 @@
...
+ * @param {RetrieveLog} retrieval
+ * @returns {string} prometheus metrics string
+ */
+function formatRetrievalMetrics(retrieval) {
+  const durationSeconds = retrieval.duration.size / 1000
+  return (
+    [
+      '# TYPE nftstorage_retrieval_duration_seconds gauge',
+      '# HELP nftstorage_retrieval_duration_seconds How long it took to retrieve an nft image',
+      `nftstorage_retrieval_duration_seconds{byteLength="${retrieval.contentLength}"} ${durationSeconds}`,
```
lanzafame, line 377: @gobengo You need to move to a histogram. A gauge is the wrong measurement type for this metric. I outlined my recommendation in the slack thread on this PR but for convenience, tldr, remove the url and image field, `size` should be a set of buckets, i.e. 10MB, 25MB, 50MB, 100MB and the metric should be a histogram (https://www.npmjs.com/package/prom-client#Histogram)

gobengo: @lanzafame Thanks for the review. Helped me realize this code is now unused, as I switched to using npm `prom-client` per @alanshaw suggestion, and in doing so I switched to a [Histogram](https://github.com/nftstorage/nft.storage/pull/1945/files#diff-5823c1dd943cae9a4760e898922c5e7a0344a231d51560db2541882f2a9beb41R15).

> size should be a set of buckets, i.e. 10MB, 25MB, 50MB, 100MB

By size do you mean `byteLength`? fwiw I removed all mention of the fields. I think we're mostly interested in measuring the retrieval duration in seconds. The byteLength/size was a nice to have.

Thanks for educating me about the perf et al tradeoffs in storing extra fields and not using buckets.

packages/cron/src/jobs/measureNftTimeToRetrievability.spec.js
```diff
@@ -0,0 +1,58 @@
...
+        gateways: [new URL('https://nftstorage.link')],
+        store: (n) => storer.store(n),
+      },
+      {
+        nftStorageToken: 'TODO',
+        metricsPushGatewayAuthorization: { authorization: 'bearer todo' },
+      }
+    )
+    assert.equal(storeSpy.mock.calls.length, 1)
+    const start = info.find((logs) => logs[0]?.type === 'start')[0]
```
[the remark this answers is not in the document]
gobengo: Good point. I switched to AsyncIterable

gobengo:
@alanshaw I believe I've addressed much of your feedback from last review. wdyt?

packages/cron/src/jobs/measureNftTimeToRetrievability.js
lanzafame: If the bytelength is still variable (as in the size of the file uploaded does change) then it should be recorded against the duration value so that the histogram can be presented correctly on the grafana end. Otherwise we will have uploads that take longer due to the file size but appear to be because the service is being slow.

packages/cron/package.json
```diff
@@ -11,29 +11,170 @@
...
+        ],
+        "rules": {
+          "@typescript-eslint/naming-convention": "off"
+        }
+      }
+    ]
+  },
+  "eslintIgnore": [
+    "node_modules"
+  ]
```
alanshaw, line 179: Please can you follow up with a PR to fix the linter issues and remove all this configuration? 🙏🙏🙏

alanshaw (changes requested):

packages/cron/src/lib/metrics.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+/**
+ * @param {Registry} registry
+ * @returns {RetrievalDurationSecondsMetric}
+ */
+export function createRetrievalDurationSecondsMetric(registry) {
+  /** @type RetrievalDurationSecondsMetric} */
+  const metric = new Histogram({
+    name: 'retrieval_duration_seconds',
```
alanshaw, line 16: We should also record the time it took to upload the data.

packages/cron/src/bin/nft-ttr.js
```diff
@@ -0,0 +1,280 @@
...
+      1000
+    )
+    .action((opts) => {
+      measure = measureNftTimeToRetrievability({
+        secrets,
+        ...createMeasureOptionsFromSade(opts, secrets),
+        ...options,
+      })
+    })
+  sadeParseWithoutExit(argParser, argv)
```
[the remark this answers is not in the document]
gobengo: I tried that but it still exited :/

packages/cron/package.json
gobengo: sure thing. didn't want to lint those old files prematurely until you weighed in on the lint rules at hand

packages/cron/src/bin/nft-ttr.js
```diff
@@ -0,0 +1,280 @@
...
+    yield* measure
+  }
+}
+
+/**
+ * Utility to use a sade instance to parse argv, but avoiding the behavior where sade may call process.exit
+ * @param {sade.Sade} prog
+ * @param {string[]} argv
+ */
+function sadeParseWithoutExit(prog, argv) {
```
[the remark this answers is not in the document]
gobengo: will do

> It's also a bit unexpected to have a "main" function in a binary that's an async iterable...

I can rename the function

packages/cron/src/lib/metrics.js
gobengo: I agree. Can we do it as fast-follow after merge or should I do before?

packages/cron/src/lib/metrics.js
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @exports
+ * @typedef {Histogram<"byteLength">} RetrievalDurationSecondsMetric
+ */
+
+/**
+ * @param {Registry} registry
+ * @returns {RetrievalDurationSecondsMetric}
+ */
+export function createRetrievalDurationSecondsMetric(registry) {
```
lanzafame, line 13: This should be milliseconds instead of seconds, unless I missed a sneaky conversion somewhere, milliseconds is the unit that is used everywhere else. Also, the conversion is handled automatically in grafana.

gobengo: from https://prometheus.io/docs/practices/naming/

> ...should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers). See below for a list of base units.

There was a conversion, but using the prom-client `Histogram` directly as a metric wouldn't have included very good type-checking so I added some types to help. https://github.com/nftstorage/nft.storage/pull/1945/commits/440e07e9885bf9544b59b99ac62aebc466ba7ec3#diff-5823c1dd943cae9a4760e898922c5e7a0344a231d51560db2541882f2a9beb41R12

packages/cron/src/bin/nft-ttr.js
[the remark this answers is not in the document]
gobengo: I made a 'cli' function that returns the iterable, and the `main` function returns a promise of its doneness.

packages/cron/src/lib/log.js
```diff
@@ -0,0 +1,56 @@
...
+ * @template {string} LogLevel
+ * @param {LogFunction<LogLevel>} log;
+ * @returns {LogFunction<LogLevel>}
+ */
+export const createJSONLogger = (log) => {
+  return (level, ...loggables) => {
+    const stringifiedLoggables = loggables.map((o) => JSON.stringify(o))
+    return log(level, ...stringifiedLoggables)
+  }
+}
```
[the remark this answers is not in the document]
gobengo: nope i removed it

.github/workflows/cron-nft-ttr.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Run job
+        env:
+          DEBUG: '*'
+          ENV: ${{ matrix.env }}
+          NFT_STORAGE_API_KEY: ${{ secrets.NFT_STORAGE_API_KEY }}
+          PUSHGATEWAY_URL: https://pushgateway.k8s.locotorp.info/metrics/job/nftstorage_ci/instance/github_action
+          PUSHGATEWAY_BASIC_AUTH: ${{ secrets.PUSHGATEWAY_BASIC_AUTH }}
+        run: |
+          yarn workspace --silent cron run --silent start:nft-ttr measure --logConfigAndExit
+          yarn workspace --silent cron run --silent start:nft-ttr measure \
```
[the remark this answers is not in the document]
gobengo: got rid of --silent and ndjson support

packages/cron/package.json
gobengo: @alanshaw this would do it. Not sure if you want to review in here or out. https://github.com/gobengo/nft.storage/pull/1

gobengo:
@alanshaw Should I merge this diff that edits unrelated modules to pass new lint rules into this PR? Or do as followup to this one? I'd slightly prefer the latter.
https://github.com/gobengo/nft.storage/pull/1/files

package.json
```diff
@@ -1,8 +1,17 @@
...
+    "packages": [
+      "packages/*"
+    ],
+    "nohoist": [
+      "@cloudflare/workers-types",
+      "@types/jest",
+      "@types/mocha",
+      "@types/node",
+      "@types/webpack-env"
+    ]
```
alanshaw, line 13: Why this `nohoist` change? Unless it causing an issue, can we move to a separate PR?

packages/cron/src/lib/metrics.js
alanshaw: Yes can be follow up

alanshaw (approved):
Lets get this merged and do the follow up items.

Note: needs a rebase (cron jobs are now integrated with release please so have to be release in the same way as other packages)

package.json
gobengo: It was causing an issue I believe. I will verify it's really required now.

gobengo: it wasn't required. :/ ty

PR #1952, opened 2022-05-30
feat: Adding HasDeleteRestriction user_tag

jsdevel:
* Adding the type and failing HTTP DELETE operations if this tag is set.
* See nftstorage/admin.storage#66

- [ ] will need migration run


vasco-santos (changes requested):
Thanks @jsdevel to add this.

Looking good, can we just update `packages/src/routes/user-tags.js` with this tag?

PR #1955, opened 2022-06-02
fix: config loading

hugomrdias:
Should fix config loading issue reported here https://github.com/nftstorage/nft.storage/pull/1914#issuecomment-1144590892


vasco-santos (approved):
LGTM

yusefnapora (approved):
Very nice, thanks @hugomrdias. I like that you moved the default config values out of the code and into the template env. In hindsight its simpler to have just the one source of truth.

yusefnapora:
@hugomrdias I'll go ahead and merge, since this was blocking production deploys.

vasco-santos:
Awesome thanks @hugomrdias and @yusefnapora . Deploys are now working 🎉

PR #1964, opened 2022-06-06
feat: add copy_upload_history db function to migrate user data

### LINKED ISSUE #1950: Allow migration of upload history from one user account to another
We want to support transferring the upload history from one account to another. 

Possible use cases are when the original account is no longer accessible, or potentially converting from a Github/oauth login to an email-based login (although this probably requires other changes as well).

Decisions to make:
- should API keys be migrated to the new user also?
- what's our policy for allowing this kind of migration?
  - e.g. how can the user verify that they are the authorized user for the original account?
- should we expose this feature through the admin interface? 
  - might be tricky, since it involves a database migration.

yusefnapora:
This adds a new postgres function `copy_user_data` to copy the upload history (and optionally auth keys and user tags) from one user id to another, to close #1950 

I added a `copyUserData` function to `DBClient` and added tests in `db-client.spec.js`, but I haven't wired this up to anything that could invoke it yet.

@vasco-santos / @hugomrdias do either of you have an opinion about how this should be invoked, given that it's currently scoped as a manual admin-only process?

I could add a script somewhere that takes a `DATABASE_CONNECTION` from the environment, and we could just run it on a dev box somewhere. 

Or. if it would be easy to add a new thing to the admin interface, we could move the `copyUserData` function out of `DBClient` here and put it into the admin backend. That's probably cleaner, but it seemed easier to test out the implementation here.


yusefnapora:
Also, we could potentially add something to the `meta` json object in the uploads during the copy operation, e.g. to add something like `"copied_from_user": "old-user-id"`.

vasco-santos:
I am not sure on how the admin user access is currently scoped. If it only has access to the view table https://github.com/nftstorage/nft.storage/blob/main/packages/api/db/tables.sql#L222 and to user tags table, or a broader scope. If so, we need to figure out if we want it to have more permissions, or have it in a different place. cc @hugomrdias @jsdevel as I am not aware of the decisions made there.

From the DB Client it will be tricky. pgrest only allows requests of up to 30s, and will timeout if more. For big users like marketplaces, I suspect this limit will be easily reached. This way, we need it to run in a server using a SQL connection and node-pg.

It looks like admin website does not have a backend and just rely on a DB client with postgrest https://github.com/nftstorage/admin.storage/blob/main/packages/website/src/lib/db-client.ts . I think we will need something else running for this. Probably a Digital Ocean machine exposing a REST API (we will need JWT Auth with a credential from the admin website to perform requests), and this server will receive HTTP requests from the website and run the function with node-pg.

packages/api/db/functions.sql
```diff
@@ -117,3 +117,66 @@ FROM cargo.aggregate_entries ae
...
+      reason,
+      inserted_at,
+      deleted_at
+  )
+  SELECT new_user_id, t.tag, t.value, t.reason, t.inserted_at, t.deleted_at
+  FROM user_tag as t
+  WHERE t.user_id = old_user_id;
+END IF;
+
+IF copy_auth_keys THEN
```
vasco-santos, line 167: Thinking more about this, and we have a gap here. Uploads will be not be associated with the auth key anymore, which limits future use cases like deleting/listing uploads by auth key. However, given the path to Uploads v2, this will probably be different with ucan (would love input from someone with more context there). 

Probably we don't want to just copy every time the auth keys, because then previous ones will continue to work, which might not be desired by the user.

Thoughts?

packages/api/src/utils/db-client.js
```diff
@@ -637,6 +637,20 @@ export class DBClient {
...
+   * upload entries.
+   *
+   * @param {import('./db-client-types').CopyUserDataInput} input
+   */
+  async copyUserData(input) {
+    const rsp = await this.client.rpc('copy_user_data', input)
+    if (rsp.error) {
+      throw new DBError(rsp.error)
+    }
+  }
```
vasco-santos, line 653: I think that we should not expose this per https://github.com/nftstorage/nft.storage/pull/1964#issuecomment-1148349005 as this will probably take more than 30s for some users

vasco-santos:
> How do you feel about those changes?

I think that is reasonable. Having a new auth key is better as previous ones should be revoked. In theory, we could even not require auth key as it is optional and leave it empty. But, I would like to keep it as we can in the future leverage it as a way to authenticate a copy history by the user without admin to act.

vasco-santos (approved):
LGTM!
@yusefnapora can you test this in staging to confirm it is working?

yusefnapora:
@vasco-santos - I just applied the migration script to the staging DB, and I did a test run to copy uploads from one account to another:

<details>
<summary>manual SQL test transcript</summary>

```
de0fa7gfjmn2ku=> SELECT id FROM public.user where EMAIL = 'yusef@protocol.ai';
 id  
-----
 124
(1 row)

de0fa7gfjmn2ku=> SELECT id FROM public.user where EMAIL = 'yusef@napora.org';
 id  
-----
 125
(1 row)

de0fa7gfjmn2ku=> SELECT id from auth_key WHERE user_id = 125;
 id 
----
 49
(1 row)

de0fa7gfjmn2ku=> SELECT COUNT(id) FROM upload WHERE user_id = 124;
 count 
-------
     2
(1 row)

de0fa7gfjmn2ku=> SELECT COUNT(id) FROM upload WHERE user_id = 125;
 count 
-------
     0
(1 row)

de0fa7gfjmn2ku=> SELECT * FROM copy_upload_history(124, 125, 49);
 copy_upload_history 
---------------------
 
(1 row)

de0fa7gfjmn2ku=> SELECT COUNT(id) FROM upload WHERE user_id = 124;
 count 
-------
     2
(1 row)

de0fa7gfjmn2ku=> SELECT COUNT(id) FROM upload WHERE user_id = 125;
 count 
-------
     2
(1 row)
```

</details>

And I can see the uploads in the web UI for both user accounts. Also the new user's uploads have the correct key_id if you query them.

I'll go ahead and merge this in, since it seems to not hurt anything :smile: Thanks again for all the help.

Issue #1692, opened 2022-03-22
research & discuss database migration strategies


yusefnapora:
This issue is just to track the discussion around DB migrations - not the implementation, which will happen in a few weeks once we've decided what we want to do.

## Where we are

DB schemas are [checked into git](https://github.com/nftstorage/nft.storage/tree/main/packages/api/db), but schema changes need to be manually applied to deployed environments, which is stressful and scary. 

## Where we want to be

Schema changes are automatically applied in CI, with tests to validate the change. The same migrations that run in local development are applied on staging and production databases, and we're confident that success in local and staging envs correlates with success in production.

## Possible solutions

So far we're looking at two ends of the "magic spectrum":

- [Sqitch](https://sqitch.org/) is unopinionated and "non-magical"... it bills itself as a "change management system," where you supply the SQL (or whatever) that needs to be applied, and it makes sure it happens.
- [Prisma](https://www.prisma.io/) is a full ORM, so it's all about the magic. You define your data model in a custom DSL, and it derives DB migrations from it. It also includes type safe DB "clients" for querying and so on. It can introspect an existing DB schema to produce a prisma data model for the initial migration. Future migrations happen by changing the prisma model.

There are other tools that fit somewhere in between these two philosophies... if anyone has a favorite, please chime in.

vasco-santos:
There are 5 requirements I could think of:
- apply latest migration script version from repo path
- apply migration script on staging DB on merge to main
- apply migration script on production DB on release on API
- support rollbacks
- support writes disabled

For the first three, we need to hook up github action workflows to perform them with one of the mentioned tools. [Flyway](https://flywaydb.org/documentation/concepts/migrations#sql-based-migrations) also seems a good candidate and they already have a [github action](https://github.com/marketplace/actions/flyway-migration).

(probably not needed for first iteration) Supporting rollbacks also seems important, both for staging and production. We will merge to main and get the migration to run in Staging DB. In the event of something not working properly, we should be able to roll it back. Same for production, as production environment has a different scale that sometimes might raise issues that we don't replicate in staging. For this part, we should probably have a manual github action workflow job that could be triggered via Github Actions UI to go back.

[...]

PR #1965, opened 2022-06-06
feat: use flyway for db migrations

yusefnapora:
Hey everyone, as I was distracting myself from something else, I started looking into using [Flyway](https://flywaydb.org) for database migrations, using the [node-flywaydb](https://github.com/markgardner/node-flywaydb) package to pull down and run the flyway cli.

This is related to #1692, but will need discussion before we can call the matter resolved. I just figured it would be easier to talk about if we had some code we could point to 😄 

Anyway, here's what's going on in the PR. The `db/migrations` directory now has flyway-compatible migrations that set up the DB schema as it currently is in `main`.

It uses Flyway's [placeholders](https://flywaydb.org/documentation/configuration/placeholder) to set the `DAG_CARGO_*` constants when setting up the foreign data wrapper.

It also makes the assumption that if `DAG_CARGO_HOST` and friends aren't in the environment, you want to use the testing dag-cargo config instead of the foreign data wrapper. To make that work, if the vars for the FDW aren't there it will set the placeholder `DAG_CARGO_TEST_MODE` to `true`, and the cargo-related migrations are wrapped in `IF` statements to run the test mode SQL if that var is true, and the FDW setup otherwise. 

On reflection, that might be too much magic - maybe it's better to read `DAG_CARGO_TEST_MODE` from the environment instead, and always set it in our dev and test setups.

All the placeholder stuff is in `db/flyway-config.cjs`.

I haven't thought through integrating with CI yet, since I wanted to get feedback and make sure this is what we actually want, etc.


yusefnapora:
Stray observations about flyway:

- "fancy migrations" that involve custom logic can be implemented in Java
- if we pay for the "Flyway Teams" thing (~3k USD/year), we can [write migrations with arbitrary scripting languages](https://flywaydb.org/documentation/concepts/migrations#script-migrations) instead

Either of those could be used to set the `MAINTENANCE_MODE` flag for migrations that require disabling writes.

It doesn't seem to support automatic rollbacks, but in the paid version you can write ["undo migrations"](https://flywaydb.org/documentation/concepts/migrations#undo-migrations) to reverse a versioned migration.

yusefnapora:
I'm curious about the custom alternative also. There are some things we'd like to do (like staging backup / rollback) that would be hard to fit into the Flyway scheme of things. It feels like Flyway wants to "own" the whole migration process, but will let you hook into it if you pay them 💸

PR #1967, opened 2022-06-08
chore: add jobs key

flea89:
Cron workflow setup file is missing the job key.

Error [here](https://github.com/nftstorage/nft.storage/actions/runs/2460571231)


adamalton (approved):
LGTM

Issue #1870, opened 2022-04-29
Max pending/unpinned quota


alanshaw:
We've recently had a user pin hundreds of thousands of CIDs that they do not have access to and can never retrieve.

This adds load to our cluster and dagcargo that is currently never resolved.

The idea here is to add a quota for maximum "in flight" pins. This effectively puts a per user cap on the number of pins that can be submitted that never become pinned. The user then starts to receive 4xx errors for pin requests, which should prompt them to delete bad pins. It also serves as a crude rate limiting mechanism.

Alternatively an adequete cleanup job could take care of this, but the idea here is to reduce that load by encouraging users to clean up after themselves.

vasco-santos:
Looking into this, I think the best approach is to add a route middleware to routes `POST /pins` and `/pins/:requestId`, augmenting the authorization layer to also check the on going requests. We can have ENV variable with max amount

PR #1974, opened 2022-06-09
feat: restrict in flight psa requests with max quota

vasco-santos:
This PR adds a route middleware to routes POST /pins and /pins/:requestId, augmenting the authorization layer to also check the on going requests. We also have a ENV var that we can add to override the current default of 100 in flight requests.

100 was chosen arbitrarily given I don't have any nice metrics to get a better number, so open for suggestions here.

Closes https://github.com/nftstorage/nft.storage/issues/1870


alanshaw (commented):
The issue is that pin status is updated asynchronously every 10 minutes, so folks might reach their quota quickly and then have to wait a long time to continue...

vasco-santos:
> The issue is that pin status is updated asynchronously every 10 minutes, so folks might reach their quota quickly and then have to wait a long time to continue...

Should we look for Pins created older than 10 minutes to start with? I think there is no great solution for this while we have Pin Status as is

PR #1979, opened 2022-06-13
docs: add store from URL example

alanshaw:
Adds an example showing how to store ERC-721 compatible metadata with image data retrieved from a URL on the internet.

Adapted from https://github.com/nftstorage/nft.storage/blob/main/examples/client/browser/store.html


vasco-santos (approved):
LGTM

PR #1993, opened 2022-06-16
retroactively eslint other cron files

gobengo:
Motivation:
* in #1945 I added some eslint rules (e.g. to avoid bugs due to dangling promises and stuff). Nothing too custom, just typescript-eslint/recommended. But I added exceptions for files I didn't touch as part of the functional goals of that PR. During review, @alanshaw encouraged me to lint those other files too. I wanted to do it as a followup PR since the other one was already pretty big.


packages/cron/src/jobs/dagcargo.js
```diff
@@ -47,7 +49,12 @@ export async function updateDagSizes({ pg, after }) {
   const countRes = await pg.query(COUNT_CONTENT_WITHOUT_SIZE, [
     after.toISOString(),
   ])
-  const total = countRes.rows[0].count
+  assert.ok(countRes.rows[0])
```
gobengo, line 52: now that node.js assert.ok is typed as a tsc type guard, this is how i chose to get type safety. 

It's at risk of throwing more errors.
but less at risk of more mysterious runtime errors.

@alanshaw lmk if you'd rather I just use a type assertion (which may be incorrect at runtime) rather than the asserts.

gobengo:
I have to admit I don't really understand why all the #1945 commits appear in here. I think it's because I didn't rebase, but merged. But the rebase seemed like it would take forever and be bug prone? If it's important to have a cleaner commit log on this, I could use some advice on how to do that. (IMO it isn't too important?)

packages/cron/src/jobs/dagcargo.js
alanshaw: We have not typically used `assert` outside of tests. It seems to add a lot of noise to the code base and (at least for me) is making it quite difficult to read the actual code. It looks like you might end up asserting on types multiple times in multiple places and I think it would be more DRY to just use a `@typedef` or `@type` once. There's also probably some performance overhead we're suffering by continually type checking everything.

It definitely makes sense to use `assert` to validate types for inputs (such as in a REST endpoint, or user input from a CLI) at _runtime_ (and likely in some other situations) but I wouldn't use it like to obtain type saftey like this in your editor or for the typescript compiler.

alanshaw (changes requested):

packages/cron/src/jobs/dagcargo.js
gobengo: I will update to use less runtime assert and instead use a [type assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) via js comment.

gobengo:
All the places where the new lint rules were unhappy due to unsafety around `pg.query`, I added a type assertion of `import('pg').QueryResult<Something>`.

packages/cron/src/lib/pinata.js
```diff
@@ -16,17 +24,21 @@ export class Pinata {
...
+      await fetchJSON(this.limiter, url.toString(), {
+        method: 'POST',
+        headers: {
+          authorization: `Bearer ${this.apiToken}`,
+          'content-type': 'application/json',
+        },
+        body: JSON.stringify({ hashToPin: cid, ...(options || {}) }),
+      })
+    )
+    const pin = /** @type {PinataPin} */ (json)
```
vasco-santos, line 41: wow, do we need the extra json const here to please the types?

vasco-santos (approved):

PR #2005, opened 2022-06-20
fix: link to NFTStorage constructor in API reference docs

alanshaw:
The `NFTStorage` class page is where folks need to be directed to.

Also adds a menu item to the reference docs.


yusefnapora (approved):
LGTM, thanks 👍

PR #2026, opened 2022-06-27
feat!: switch to Elastic IPFS

alanshaw:
This PR switches to uploading to the Elastic IPFS S3 bucket by default and uploading to IPFS Cluster asynchronously after the response is sent to the client.

We pin the root CID to Cluster and if the upload to Cluster fails, we hope that Cluster can fetch it from elastic provider later.

A dockerised minio is added to dev and test which stands in for a real life S3 in these environments.

BREAKING CHANGE: uploaded files are no longer instantly available via gateways or on the DHT. They may take a few seconds to become indexed by Elastic IPFS.


packages/api/src/routes/nfts-upload.js
```diff
@@ -158,16 +157,27 @@ export async function uploadCarWithStat(
     mime_type: mimeType,
     type: uploadType,
     content_cid: stat.rootCid.toV1().toString(),
-    source_cid: stat.rootCid.toString(),
-    dag_size: stat.structure === 'Complete' ? added.bytes : stat.size,
+    source_cid: sourceCid,
+    dag_size: stat.structure === 'Complete' ? stat.blocksSize : stat.size,
```
vasco-santos, line 161: Shouldn't this be only stat.size? We do that in web3 https://github.com/web3-storage/web3.storage/blob/1d101812cbb8db2099dc2fb03447bf2f94f72f40/packages/api/src/car.js#L138 and was expecting same here.

alanshaw: Yeah, ok I'm going to revert that `blocksSize` change and always use `stat.size` as in web3.storage.

Turns out using `added.bytes` here was **not** necessarily the size of the DAG (it's just the number of bytes that were imported by cluster). In one of our tests we add _two blobs_ with the _same name_ and the importer creates a DAG with just two blocks (the directory and the second blob) even though the CAR contains 3 blocks (one is orphan).

Resolving this issue will fix the edge case (either by disallowing or allowing it and creating a directory with two files with the same name): https://github.com/ipfs/js-ipfs-unixfs/issues/232

packages/api/test/nfts-upload.spec.js
```diff
@@ -386,7 +388,9 @@ describe('NFT Upload ', () => {
     assert.equal(upload.backup_urls[0], backupUrl)
   })
 
-  it('should backup chunked uploads, preserving backup_urls for each chunk', async () => {
+  it('should backup chunked uploads, preserving backup_urls for each chunk', async function () {
+    this.timeout(10_000)
```
alanshaw, line 392: IDK this just needed a bit more time on my laptop

vasco-santos (approved):
LGTM

olizilla (approved):
VERY GOOD

hugomrdias (commented):
niceee !

PR #2027, opened 2022-06-27
feat: abortable API method calls

### LINKED ISSUE #1973: Timeout for requests taking too long
Maybe use abortsignal/controller in the fetch? I have been doing one at a time, 98% take two seconds, then randomly 20,30, or 60+ seconds. It's for a mint on demand system so 60 will make them give up especially if it's two seconds almost always.

Thanks for considering.

alanshaw:
Adds `signal` (an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal)) as an optional parameter to methods that call the API.

resolves #1973


vasco-santos (approved):
LGTM

PR #2031, opened 2022-06-27
fix: no more PSA mailto link

dchoi27:


alanshaw:
How come this is going away?

JeffLowe:
We now have a webform on /manage page for users to request Pinning Service access.  However, we could leave this GH issue template in and instead point to the /manage page or to the docs page https://nft.storage/docs/how-to/pinning-service/#requesting-access that explains how to request access

dchoi27:
I think we just get rid of it in GH issues - users should really just be using their account page to manage things (and focus on GH for bug reports + feature requests). But if folks feel strongly about leaving something in GH, yes just a link to where they can submit the form would be fine too (no strong opinions)

alanshaw (approved):

drewdelano (approved):
LGTM

Issue #2033, opened 2022-06-28
nft-ttr metrics should have label for which gateway it was using


gobengo:
Motivation:
* so dashboards can distinguish and contextualize the retrieval_duration_seconds observations

gobengo:
It's added now
https://protocollabs.grafana.net/d/9wQvbG37k/bengo-nft-ttr?orgId=1&from=1656687354543&to=1656708954543&viewPanel=10

cc @alanshaw @dchoi27

Issue #1994, opened 2022-06-16
cron: nft-ttr job should also track the time it took to upload the data


gobengo:
Motivation:
* https://github.com/nftstorage/nft.storage/pull/1945#discussion_r892340411

gobengo:
`store_duration_seconds` cc @alanshaw @dchoi27 

https://protocollabs.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22Prometheus-k8s%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22store_duration_seconds_sum%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D

PR #2040, opened 2022-06-28
feat(cron): nft-ttr logs store_duration_seconds to prometheus

gobengo:
Motivation:
* https://github.com/nftstorage/nft.storage/issues/1994
* https://github.com/nftstorage/nft.storage/issues/2033


alanshaw (approved):
IMHO it's not worth the extra code to pass console around everywhere just to have pretty test output. I'd just use [`debug`](https://www.npmjs.com/package/debug) so that when you need the debugging info you can get it.

Otherwise this looks fine.

PR #2065, opened 2022-07-08
chore: convert tests to use miniflare + ava

yusefnapora:
This is a start at converting our test setup to use ava and miniflare.

I started pulling this thread yesterday, since @hugomrdias mentioned earlier in a sync call that it would be nice to replace playwright-test with miniflare. (cc @redaphid)

Since we're starting to use testcontainers in a few places, I figured this would be a good chance to try that approach here, since I was getting rid of the playwright hooks that start the docker services anyway. It turned out to work really well - I originally had it starting up one docker environment per test file, but that was crazy resource intensive, so I found a way to get Ava to do the container setup in a shared worker so all the test cases can use it. So in this branch the docker environment for test is managed in js (see `test/scripts/containers.js`) and is independent of the dev docker-compose setup.

There's still two test cases that I'm skipping, so it's not quite ready to merge.  I also set the minimum node version to 18.x, but I'm not 100% sure it's needed. That's likely why CI is failing tho.

Would love :eyes: on it if anyone has time - you'll need node 18.4.0 or later (lower versions trip a miniflare bug with ed25519 keys).

Update (2022-07-13):

I think all the blockers are sorted. There's still one test that's being skipped due to a bug in Miniflare, but I [opened a PR for it](https://github.com/cloudflare/miniflare/pull/309) and hopefully we can re-enable it soon.

A possible improvement could be to rewrite the container setup code to use docker compose. I found out after I wrote that bit that testcontainers supports docker compose, so we could just define everything in the compose file instead of duplicating the setup in code. I might play with that a bit & see if it simplifies things.


yusefnapora:
I updated all the workflows to use node 18.x, but Cloudflare Pages only supports up to 17.x [in their build environment](https://developers.cloudflare.com/pages/platform/build-configuration/). If we can't roll back to node 16, we could always build the site in an action with node 18 and manually publish it with `wrangler`.

Also, with the `engines` directive in the api package, you'll need at least 18.4.0 to run `yarn install`. Lower versions should be able to use the other packages (client, etc) when installed from npm, but they won't be able to build the project. Anyway, I'll see if I can get it working with node 16 on Monday.

olizilla:
Rad!!! 
Re: cloudflare pages... Let's move to building the site in our CI and then publishing the files to CF as in https://github.com/web3-storage/web3.storage/pull/1603

yusefnapora:
I figured out why the test I'm skipping gives a different result in Miniflare than prod, and made a PR to fix it in Miniflare: https://github.com/cloudflare/miniflare/pull/309

I also made a change to how the test helpers work, which complicates things in some ways but simplifies in others :smile:

The complicated bit is that I'm building a second CF worker that's only used in tests, which gets "mounted" by Miniflare to the URL `http://test.mf`. Test helpers that need to use Web APIs (`fetch`, `crypto.subtle`, etc) now live in this test worker, and unit tests invoke the helpers by dispatching `FetchEvent`s with Miniflare to `http://test.mf/mt-test-helper-endpoint`, e.g. `http://test.mf/create-user`.

The main motivation here is to avoid the hack I was using of checking whether `crypto` exists in global scope and dynamically importing `node:crypto` if not. This was failing the dev deploy (and would presumably fail prod, etc), since `wrangler` checks to make sure you're not using node builtin APIs.

I _thought_ that this would also let me stop special-casing the Ed25519 signature stuff - Miniflare claims to support the special `NODE-ED25519` algorithm, but even when I'm definitely running in Miniflare I get "not supported" errors. I did get rid of the fallback to `tweetnacl` though, and just use the standard node `Ed25519` algorithm name, which works with node's `webcrypto` implementation.

Anyway, this is basically ready apart from the CF pages website build, assuming we're okay with bumping the minimum node version to 18.4.0. 

I'll revisit the version thing once I get the site build / deploy action working - I think it might be easier to roll back to 16.x now that I moved the `fetch` stuff, but it might mean bringing `tweetnacl` back.

yusefnapora:
Huzzah for the :heavy_check_mark:!

I was able to get us back on node 16 by moving the test code that uses our cluster client into the test worker (it uses `fetch`) and rewriting a couple other things so that no Web API stuff gets imported outside of a miniflare worker.

I do think we should build the site in an action like @olizilla said, node version notwithstanding, but that works better as its own PR.

Let me know if I left any debug cruft in the diff - I'll go through again tomorrow when I've got a better forest / tree perspective :smile:.

packages/api/src/utils/car.js
```diff
@@ -1,5 +1,6 @@
 import { CID } from 'multiformats'
 import { CarWriter } from '@ipld/car'
+import { Blob } from '@web-std/blob'
```
hugomrdias, line 3: can we just use node 18 https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#class-blob ?

packages/api/src/utils/ed25519.js
```diff
@@ -11,24 +11,29 @@ const CLOUDFLARE_ED25519 = {
  * Verifies an Ed25519 signature over a given `message`.
  *
  * Will try to use the native Ed25519 implementation provided by the CloudFlare Worker API
- * if possible, falling back to js-based verification if not running in a CF Worker (e.g. in tests).
+ * if possible, falling back to node's builtin 'Ed25519' algorithm if that fails.
```
hugomrdias, line 14: just use https://github.com/paulmillr/noble-ed25519 everywhere

hugomrdias (changes requested):
first pass LGTM but the extra worker seems too complicated

we should just use node 18 for the API packages and polyfill with

https://github.com/paulmillr/noble-ed25519
https://github.com/fission-suite/one-webcrypto
https://github.com/web-std/io/blob/main/packages/fetch or undici 

where needed

yusefnapora:
@hugomrdias sounds good to me - I'll fix it up this afternoon

vasco-santos (commented):
I also agree with Hugo that it would be nicer to use Node 18 and polyfill to avoid having the extra worker complexity

yusefnapora:
Thanks for the reviews @hugomrdias and @vasco-santos - I force-pushed over this branch, since I reverted back to before the test worker change. With node 18 we don't need to polyfill `fetch`, `Blob`, or `File`, which is pretty sweet. Removing the `engines` restriction totally did the trick, thanks Hugo :)

I also changed the version test a bit so that we don't need to special case the worker build for the test environment. Instead the test just uses the git module to figure out what the correct version string should be.

The last commit pins some docs-related packages in the website dependencies - a new beta release broke the build and got pulled in when I did a `yarn clean` / `yarn install`. So I changed the version specifier to be exact, which feels like the right thing to do for a beta dependency.

yusefnapora:
Added a commit that cleans up the testcontainer code by using it's docker compose support. To make it work, I split off two new compose files that are only used for dev - `docker-compose-local-ports.yml` exposes the DB, etc on local host ports, and `docker-compose-volumes.yml` bind-mounts stateful volumes. 

Later we can make the volumes optional to make persistent state opt-in, but I wanted to leave the dev setup alone in this PR.

yusefnapora:
Some flailing in the last few commits to debug a docker thing. Apparently the docker compose version I have locally names its services with hyphens, e.g. `service-1`, but the one on CI uses `servce_1`, and the underscore was confusing things. I added a `container_name` entry to the service definitions so we don't depend on the auto-generated name, which seems to fix things up.

yusefnapora:
Oh, I also added a `services pull` command to the cli.js, since I was thinking that we were timing out during test setup when pulling images and wanted to pull stuff before the clock started ticking. I don't think that's what the problem actually was, but it doesn't seem to hurt.

packages/api/test/db-client.spec.js
```diff
@@ -1,58 +1,58 @@
...
-        iss: 'nft-storage',
-        iat: Date.now(),
-        name: 'key1',
-      },
-      config.SALT
-    )
-    await client.client.createKey({
+test.serial('getUser should list only active keys', async (t) => {
+  const client = await createClientWithUser(t)
+  console.log('test')
```
vasco-santos, line 15:
```suggestion
```

packages/api/test/nfts-store.spec.js
```diff
@@ -1,144 +1,158 @@
...
+      status: 'PinQueued',
+      service: 'IpfsCluster3',
+    },
+  ])
+})
 
-  it('should store dir wrapped image', async () => {
-    const { token, userId } = await createTestUser()
+// Miniflare doesn't currently preserve pathnames in File objects uploaded via FormData.
+// TODO: re-enable this test once https://github.com/cloudflare/miniflare/pull/309 is merged
```
vasco-santos, line 98: Can we open an issue in our repo pointing to that one?

packages/api/test/scripts/test.env
```diff
@@ -0,0 +1,30 @@
...
+
+S3_ENDPOINT=http://127.0.0.1:9000
+S3_REGION=us-east-1
+S3_ACCESS_KEY_ID=minioadmin
+S3_SECRET_ACCESS_KEY=minioadmin
+S3_BUCKET_NAME=dotstorage-test-0
+
+NFT_STORAGE_VERSION=test-version
+NFT_STORAGE_BRANCH=test-branch
+NFT_STORAGE_COMMITHASH=test-commit
```
vasco-santos, line 30: For now we can keep this. We should create an issue to track having a single place for all these

packages/api/test/scripts/test-context.js
```diff
@@ -0,0 +1,118 @@
...
+
+  if (!noContainers) {
+    const sharedWorker = await registerSharedWorker({
+      filename: path.resolve(__dirname, 'containers.js'),
+      supportedProtocols: ['ava-4'],
+    })
+
+    // wait for the shared worker to publish an object with env var overrides
+    for await (const message of sharedWorker.subscribe()) {
+      // console.log('got message from ava shared worker', message)
```
vasco-santos, line 62:
```suggestion
```

packages/api/test/scripts/test-context.js
```diff
@@ -0,0 +1,118 @@
...
+      if (!message.data || typeof message.data !== 'object') {
+        continue
+      }
+      if (!('overrides' in message.data)) {
+        continue
+      }
+      // console.log('setting environment overrides', message.data)
```
vasco-santos, line 69:
```suggestion
```

packages/api/test/scripts/test-context.js
```diff
@@ -0,0 +1,118 @@
...
+      // @ts-ignore
+      overrides = { ...overrides, ...message.data.overrides }
+      break
+    }
+  }
+
+  const mf = makeMiniflare(overrides)
+
+  const bindings = await mf.getBindings()
+  // console.log('miniflare bindings', bindings)
```
vasco-santos, line 79:
```suggestion
```

vasco-santos (commented):
Great PR Yusef! This is looking good to me.

The extra worker now seems way simpler. There is room for improvements here, but I feel like this is a good first iteration.

Let's wait for @hugomrdias 's feedback

packages/website/package.json
```diff
@@ -27,8 +27,8 @@
     "magic-sdk": "^6.2.1",
     "multiformats": "^9.6.4",
     "next-mdx-remote": "^4.0.0-rc.1",
-    "nextra": "^2.0.0-beta.5",
-    "nextra-theme-docs": "^2.0.0-beta.5",
+    "nextra": "2.0.0-beta.5",
+    "nextra-theme-docs": "2.0.0-beta.5",
```
hugomrdias, line 31: can we revert this please

packages/api/tsconfig.json
```diff
@@ -26,5 +26,5 @@
     "test",
     "scripts"
   ],
-  "exclude": ["node_modules/", "dist/"]
+  "exclude": ["node_modules/", "dist/", "test/test-worker"]
```
hugomrdias, line 29: this is not needed right ?

packages/api/test/scripts/test-context.js
```diff
@@ -0,0 +1,118 @@
...
+  const serviceConfig = serviceConfigFromVariables(bindings)
+  if (bindGlobals) {
+    // optionally pull cloudflare bindings into the global scope of the test runner
+    defineGlobals(bindings)
+  }
```
hugomrdias, line 84: why do we need this ?

packages/api/scripts/cmds/services.js
```diff
@@ -5,12 +5,31 @@ import { MINIO_API_PORT } from './minio.js'
...
+]
+
+const composeFileArgs = composeFiles.flatMap((filename) => [
+  '--file',
+  path.join(composeDir, filename),
+])
 
 const PG_PORT = 3000
 const PGRST_PORT = 5432
 const CLUSTER_PORT = 9094
```
hugomrdias, line 22: do we need this ?

packages/api/docker/docker-compose-local-ports.yml
```diff
@@ -0,0 +1,15 @@
...
+  db:
+    ports:
+      - 5432:5432
+  cluster:
+    ports:
+      - '127.0.0.1:9094:9094'
+  minio:
+    ports:
+      - '9000:9000'
+      - '9001:9001'
```
hugomrdias, line 15: why do we need this file ? i would expect that with test-containers we could have dynamic port

hugomrdias (changes requested):
LGTM left some comments inline

so dev loads with persistence and fixed ports and test with no persistence and dynamic ports, is this correct?

packages/website/package.json
yusefnapora: There seems to have been a breaking change in nextra 2.0.0-beta.6, which I didn't have time to figure out the details of last week. It will show up on `main` if you delete the lockfile and `yarn install` again. 

I'd rather pin the dependency here and fix the issue in a separate PR, if that works for you.

packages/api/test/scripts/test-context.js
yusefnapora: I'm using it to test the config loading code in config.spec.js, so we can test the `loadConfigVariables` function outside of the worker context. I'll take another look at those tests and see if there's another way to do it.

packages/api/docker/docker-compose-local-ports.yml
yusefnapora: This file and the other one for persistent volumes are only used for dev. I wanted to keep this PR focused on test, so I tried to make the dev setup function the same as before, with fixed ports and persistent data.

That said, we totally can use testcontainers to bring up the dev environment also. I had a PoC of that in #2073, which I closed since it was based on this PR branch and it seemed best to wait for things to settle down here.

The main thing about using testcontainers with ephemeral ports for the dev environment is that you need to set some environment variables for Miniflare, so it can find the services on their mapped ports. I did that in #2073 by using the Miniflare JS api to bring up the dev server, with a `bindings` option to set the service URLs.

Happy to either pull the dev setup into this PR or make a new one.

vasco-santos: Let's keep this PR strictly for test. It makes sense that docker compose should not have host ports so that testcontainers just grab an available one. And that means the need for specifying them for dev

packages/api/test/scripts/test.env
yusefnapora: Hmm... actually, we might not need this after all.

I added this early in the process hoping we'd be able to get rid of it, and now that I've thought a bit more, the only differences between this and the dev `env.tpl` are the service ports, which are overriden anyway, and the S3 bucket name. But since dev and test will be using different minio images, so they can have the same bucket name without collisions, so we can just use `dotstorage-dev-0` for both.

Let me see if I can get things passing without this file - that would be sweet.

packages/api/scripts/cmds/services.js
yusefnapora: I think so, since this PR is still using fixed ports for dev (see other comment).

packages/api/test/scripts/test-context.js
yusefnapora: I rewrote the config tests to avoid the `bindGlobals` option - the tests now set the global variables directly as part of the test setup.

yusefnapora:
I realized that there was a problem with using a fixed `container_name` in the docker compose file. If it's set, testcontainers doesn't use a unique prefix any longer, so the container names are just `ipfs`, `db`, etc. This means that you can't run tests while the dev environment is also running.

I found a simpler way to fix the issue that caused me to add the `container_name` though. Apparently you can use the `COMPOSE_COMPATIBILITY` environment variable to force both versions of docker compose to use the same `_1` suffix for container names. This lets the testcontainers code find the containers regardless of which version you have installed.

packages/api/test/nfts-store.spec.js
yusefnapora: Good call - I made #2088 to remind us to re-enable once the bug is fixed.

packages/api/test/scripts/containers.js
```diff
@@ -0,0 +1,111 @@
...
+      'db_1',
+      Wait.forLogMessage('PostgreSQL init process complete')
+    )
+    .withWaitStrategy('ipfs_1', Wait.forLogMessage('Daemon is ready'))
+    .withWaitStrategy('cluster_1', Wait.forLogMessage('IPFS Cluster is READY'))
+    .withWaitStrategy('minio_1', Wait.forLogMessage('1 Online'))
+    .up()
+
+  const ports = {
+    minio: compose.getContainer('minio_1').getMappedPort(9000),
```
hugomrdias, line 35: what does getMappedPort mean ? is it dynamic or not ? is the 9000 just like "try 9000 if that fail give me a free port" ?

packages/api/src/errors.js
```diff
@@ -1,7 +1,21 @@
...
+  TokenExpired: 'ERROR_DIDT_EXPIRED',
+  TokenCannotBeUsedYet: 'ERROR_DIDT_CANNOT_BE_USED_YET',
+  IncorrectSignerAddress: 'ERROR_INCORRECT_SIGNER_ADDR',
+  FailedRecoveryProof: 'ERROR_FAILED_RECOVERING_PROOF',
+  ApiKeyMissing: 'ERROR_SECRET_API_KEY_MISSING',
+  MalformedTokenError: 'ERROR_MALFORMED_TOKEN',
+  ServiceError: 'SERVICE_ERROR',
+  ExpectedBearerString: 'EXPECTED_BEARER_STRING',
+}
+
```
hugomrdias, line 18: any way to avoid this ? did we file an issue ?

hugomrdias (changes requested):
LGTM, left some comments inline but look ready to merge.

packages/api/test/scripts/containers.js
vasco-santos: Give me the docker container port mapped to the internal service running on 9000

vasco-santos: so, for the host port it is dynamic yes, the 9000 is defined in docker compose for the service inside the container

hugomrdias: got it thank you

yusefnapora:
So, I've got some sad news about this branch - when running on Github Actions, it seems to get "stuck" sometimes in a way that gives no useful debug output - the `yarn test:api` step just hangs until you cancel the test or it times out (about an hour). I first noticed this earlier in the week, but hoped it was caused by the fixed `container_name` thing. Sadly, it happened on the last commit, which didn't contain any code changes (just a comment typo).

Re-running the failed test seems to succeed, or at least it has so far when I've tried it. But having the tests just randomly hang forever seems unacceptably flaky, so I don't think we can merge this until we figure it out :cry:

I'm going to try to reproduce locally... first guess is to try running the workflow with [act](https://github.com/nektos/act) to get closer to the CI environment. Could also be that I'm not seeing the issue locally because I'm using docker-compose v2, so I'll try with v1 and see if it triggers the issue.

packages/api/src/errors.js
yusefnapora: Oh man, thanks for reminding me - I forgot about this. I'll dig a little deeper and see if I can figure out a solution

vasco-santos:
@yusefnapora let me know if you need any help

yusefnapora:
So, I lost another day to this, but I think I may have done it :crossed_fingers: 

To figure out why things were hanging, I wanted to eliminate testcontainers as a variable and replaced it with a script that runs `docker compose` before starting the tests, pretty similar to what we had before, but still isolated from the dev env.

Sadly, things still got stuck, with ava refusing to exit even after all tests passed. However, with testcontainers out of the mix, I no longer need to use a shared worker to coordinate the container bring-up, which means that I can tell ava to run each test file in its own subprocess instead of using one shared process with a bunch of isolated `Worker`s.

That seems to have fixed the hanging issue, and I've now been running the tests in a loop for ~30m without getting stuck. However, this means that even though testcontainers specifically wasn't causing the hanging issue, we still can't use testcontainers + ava, because that needs the shared worker, which requires `workerThreads: true`, which is the thing that seems to keep failing.

Also, I know that we said we'd keep this focused on test only, but I got in a hyper focus zone and couldn't resist using the docker compose script to bring up the dev env also.

The new script that runs docker compose is at `docker/run-with-dependencies.sh`, and it starts docker compose, then runs the command you pass in and cleans up docker compose at the end. If you set the `PERSIST_VOLUMES` environment var, it will also create named docker volumes for all the stateful services and use them. This replaces the bind mount to the `docker/compose` directory, and is wired up to `yarn:dev:persist`

So, to summarize:

- `docker/run-with-dependencies.sh` runs whatever command you tell it, sandwiched between `docker compose up` and `docker compose down` calls
  - it also exports overridden values for `DATABASE_URL`, etc
    - fun fact: if you list a port in your compose file but don't bind it to a host port (e.g. you use `3000` instead of `3000:3000`), docker will pick a random open port. You can then use `docker compose port <service-name> <port-number>` to find the mapped host port.
  - if you set `PERSIST_VOLUMES=anything` before running `run-with-dependencies.sh`, it will created named docker volumes (e.g. `nftstorage-ipfs-data`) and add a second compose file (`docker-compose-persist.yml`) to point the services at those external volumes instead of creating new ephemeral volumes
- `scripts/run-dev.sh` does the stuff that was previously in the `predev` and `dev` npm scripts. It's designed to be wrapped by the `run-with-dependencies` script. 
  - `yarn dev` now runs `./docker/run-with-dependencies.sh ./scripts/run-dev.sh`.
  - `yarn dev:persist` runs `yarn dev` with the `PERSIST_VOLUMES` var set
  - `yarn dev:clean` runs `./docker/clean-persistent-data.sh`, which just deletes the named volumes
- `scripts/run-test.sh` uses the cli script to setup the db schema and minio bucket and runs the test suite with `ava`
  - this is also designed to be wrapped with the docker script - `yarn test` now runs `./docker/run-with-dependencies.sh ./scripts/run-test.sh`

Miniflare and ava are both running on the host (outside of docker), and they pick up the environment variables set by the `run-with-dependencies` script. The miniflare cli does this by default, but I had to add a couple lines to the test setup to pull in the service vars from `process.env`.

@vasco-santos sorry about bundling up dev and test into one PR - I really need to context switch out of this problem so I wasn't sure I'd be able to follow through on a separate PR, and it was super easy to do the dev setup once I'd written the `run-with-dependencies` script.

vasco-santos:
Thanks for the new iteration @yusefnapora 

So, the shared process goal was an optimization to not restart containers for every test suite right? Do you have a rough estimate of how long it was taking? Anyway, I am not against using this setup with docker compose and move this discussion to an issue. @yusefnapora Could you also open an issue in Ava with your findings? They can potentially just get this fixed :)

I did not have a chance to look into the code changes yet. I would really prefer you to split the dev part in a follow up PR though, so that we can have this unblocked and you to move to other tasks. I know @hugomrdias is not keen on having a setup with shell scripts, and I am inclined to agree with him for consistency. We don't have shell scripts for setup in any other project, so it would just be easier for everyone to have this executions in the JS scripts with execa. The rest of the dev bits should be fine, but I know that the shell scripts will need discussion to go through. 

We can get this PR in without them, otherwise we should block on getting @hugomrdias back.

yusefnapora:
@vasco-santos I reverted the changes to the dev setup, so it's back to the status quo.

About the shared worker, I started using it because my local machine was running out of resources trying to run the full test suite - each test file would spin up its own testcontainers setup, and it ate up all my RAM :) There may be a way to share the testcontainers setup without the shared worker - testcontainers has a `withReuse` option that lets multiple threads / processes share a setup, but I haven't tried using it.

I haven't files an ava issue yet, mostly because I'm not sure how to reproduce it in a standalone way. Maybe I'll fork this branch at a commit that shows the bug and put a link in the issue.

BTW, if the objection to the dev changes is just about shell vs javascript, I could rewrite the `run-with-dependencies` thing to work as a node script instead.

yusefnapora:
I looked a bit this morning at converting the shell script for the test setup into a command in the js cli - it's a non-trivial thing to get right, so I don't thing we should block this PR on it, since I won't have time to keep rebasing this branch against changes to main.

@vasco-santos lmk if you're okay with merging this, now that it's test-only. If not, we either need to find someone else to take over this PR or close it, I think. Otherwise it will keep drifting from `main` whenever a new test is merged.

As far as the dev setup goes, I currently don't have the time or motivation to work on getting that over the line, and as I said earlier, I don't think I can commit to following up with another PR. I personally think we should just merge the thing that works, so @cmunns can use it to help onboard the new dev on their team, and we can fix the stylistic objections "in post". 

But since that doesn't seem like it's going to fly, and I can't commit to being available to work on this further, I'll just leave a link to the commit that reverted the working shell script solution, so whoever does pick this up can use it as inspiration:  https://github.com/nftstorage/nft.storage/pull/2065/commits/0b90ca971e1dc0ec188cb588d8ee6d72c7224f83

packages/api/package.json
```diff
@@ -9,9 +9,10 @@
     "deploy": "wrangler publish --env production",
     "predev": "./scripts/cli.js services start && ./scripts/cli.js db-sql --cargo --testing --reset && ./scripts/cli.js minio bucket create dotstorage-dev-0",
     "dev": "miniflare dist/worker.js --watch --debug --env ../../.env",
-    "clean": "./scripts/cli.js services stop --clean && rm -rf docker/compose",
+    "dev:persist": "PERSIST_VOLUMES=true npm run dev",
+    "dev:clean": "./docker/clean-persistent-data.sh",
```
vasco-santos, line 13:
```suggestion
```

vasco-santos (approved):
Changes look good, I am super happy to have the tests using Ava and Miniflare. I think this has direct value for everyone contributing to the project.

There are concerns around the shell scripts as we previously talked about. We need to get them into a JS CLI, but I agree that we should not block this anymore on that.

We can do it in a follow up PR, rather than keeping this branch running in parallel with main and having the continuous burden of rebasing when tests change. Also, Ava + miniflare already offer a great value that I don't want to drop for the shell scripts.

@yusefnapora can I just ask you to create an issue to discuss/refactor the shell scripts added here?

yusefnapora:
Sounds good, thanks @vasco-santos :) I'll make an issue about the bash -> JS thing so we can keep track of it, and hopefully I'll have some time to pick it up next week.

Issue #2066, opened 2022-07-12
/login returns 404 for GET and 500 for POST


olizilla:
### Using incorrect http method

Here we send `GET /login` lbut the route is defined as `POST /login` so we end up returning a 404, even tho the error message is also warning us that we will need a bearer key... 

```console
curl -I "https://api.nft.storage/login"
HTTP/2 404 

curl "https://api.nft.storage/login"
{"ok":false,"error":{"code":"EXPECTED_BEARER_STRING","message":"API Key is missing, make sure the `Authorization` header has a value in the following format `Bearer {api key}`."}}
```

If we added a bearer key we'd still get a 404 so we should be returning a `405 Method Not Allowed`.

### Not providing authorization header

If we send POST /login but without auth credentials:

```console
curl -I -X POST "https://api.nft.storage/login"
HTTP/2 500 

curl -X POST "https://api.nft.storage/login"
{"ok":false,"error":{"code":"SyntaxError","message":"Unexpected end of JSON input"}}
```

This should return a `401 Unauthorized`, and the response body should be as in the preview example

```
{"ok":false,"error":{"code":"EXPECTED_BEARER_STRING","message":"API Key is missing, make sure the `Authorization` header has a value in the following format `Bearer {api key}`."}}
```

related, it looks like we fixed this once before https://github.com/nftstorage/nft.storage/pull/95

olizilla:
we convert the error code here
https://github.com/nftstorage/nft.storage/blob/f503a4cca62b8fb9c39ecc3dd629370a08cbb82f/packages/api/src/errors.js#L73-L77

why it not return i wonder...

olizilla:
oh wat, POST is 500
```console
 curl -I -X POST "https://api.nft.storage/login"
HTTP/2 500 

curl -X POST "https://api.nft.storage/login"
{"ok":false,"error":{"code":"SyntaxError","message":"Unexpected end of JSON input"}}
```

olizilla:
👀 a request for `GET /login` is getting handled by the`GET /:cid` route, which includes the `withAuth` middleware-ish fn which then calls [validate](https://github.com/nftstorage/nft.storage/blob/f503a4cca62b8fb9c39ecc3dd629370a08cbb82f/packages/api/src/utils/auth.js#L36), which fails with when it tries to parse the auth header.

olizilla:
_root route catch-alls considered harmful_

olizilla:
[...]

PR #2067, opened 2022-07-13
wip: router changes to allow method-not-allowed

olizilla:
add tests for
- GET /login to return 405 Method Not Allowed
- POST /login to return 401 Unauthorized if auth header not present

initial work on getting router to find path-but-not-method matches

hampered by the `GET /:cid` route matching `GET /login`

see #2066

License: (Apache-2.0 AND MIT)

### DIFF
--- a/packages/api/src/index.js
+++ b/packages/api/src/index.js
@@ -124,7 +124,9 @@ r.add(
 // Upload
 r.add('get', '/check/:cid', withMode(nftCheck, RO), [postCors])
 r.add('get', '', withAuth(withMode(nftList, RO)), [postCors])
-r.add('get', '/:cid', withAuth(withMode(nftGet, RO)), [postCors])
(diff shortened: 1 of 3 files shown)


packages/api/src/index.js
```diff
@@ -124,7 +124,9 @@ r.add(
 // Upload
 r.add('get', '/check/:cid', withMode(nftCheck, RO), [postCors])
 r.add('get', '', withAuth(withMode(nftList, RO)), [postCors])
-r.add('get', '/:cid', withAuth(withMode(nftGet, RO)), [postCors])
+r.add('get', /^\/(?<cid>ba\S+|Qm\S+)/i, withAuth(withMode(nftGet, RO)), [
```
alanshaw, line 127: What if we added:

```js
r.add('get', '/login', () => new Response({ status: 405, statusText: 'Method Not Allowed' }), [postCors])
r.add('get', '/:cid', withAuth(withMode(nftGet, RO)), [postCors])
```

* No regex
* Allows other multibases to be supported if we want

PR #2073, opened 2022-07-15
chore: use testcontainers to orchestrate dev server dependencies

yusefnapora:
This builds on #2065 to also use testcontainers to bring up the environment for the Miniflare dev server.

This works by way of a new `dev.js` script that uses testcontainers to load the docker-compose environment, gets the mapped port numbers for all the services, and starts Miniflare with overridden environment variables so it can reach the services. 

It also prints the env vars out to the console when you start, so you can muck with the DB manually, etc if you need to:

```
❯ yarn dev
yarn run v1.22.19
$ ./scripts/dev.js
🚢📦 Starting service containers...
$ scripts/cli.js build
🚢📦 Containers started.

🌎 Environment overrides: 
export S3_ENDPOINT=http://localhost:49343
export DATABASE_URL=http://localhost:49344
export CLUSTER_API_URL=http://localhost:49345
export DATABASE_CONNECTION=postgresql://postgres:postgres@localhost:49341/postgres

📡 Dev API server listening on http://localhost:8787
```

By default it will create ephemeral volumes for the stateful services, but you can also run `yarn dev --persist` to have it create volumes in `packages/api/docker/compose`, like before.

@hugomrdias how do you feel about this approach? It seems like having our own entrypoint might make attaching a debugger to the miniflare process easier if we ever need to, although I haven't tried that yet.


yusefnapora:
Going to close this, but may reopen a similar one once #2065 lands

packages/api/scripts/containers.js
```diff
@@ -0,0 +1,108 @@
...
+    CLUSTER_API_URL,
+    DATABASE_CONNECTION,
+  }
+  return { overrides }
+}
+
+/**
+ * @param {string} connectionString
+ */
+async function initDBSchema(connectionString) {
```
redaphid, line 75: I'm a _huge_ fan of how simple the database init is :)

PR #2078, opened 2022-07-19
refactor: remove pinata

### LINKED ISSUE #260: Rework Piñata backup logic
The v2 cluster has 6 nodes. You can only pass 5 of them to Piñata as origins:

> you can help Pinata find your content faster by optionally providing us with the "multiaddresses" up to five host nodes that your content already resides on.

With the current logic, we would pass all 6, as we pass through the static list provided by CLUSTER_ADDRS in https://github.com/ipfs-shipyard/nft.storage/blob/b93614ece6806611addea215726ff43f5f7f98bc/packages/api/wrangler.toml#L22 (currently 3, but will become 6 when we adopt the v2 cluster)

We should upgrade the pinning logic to query cluster to find out which nodes actually host the pins, and verify they are online before passing them to Piñata:

> ⚠️ Please make sure every node provided is online. Pinata will attempt to connect to all nodes before pinning the content, and if any these nodes are offline, your request will eventually fail.

ref: https://github.com/PinataCloud/Pinata-SDK#host-node-multiaddresses

### LINKED ISSUE #391: Continuous Pinata backup job on Heroku
Issue for similar work to: https://github.com/web3-storage/web3.storage/pull/411

Basically trying to pin to Pinata continuously so we fall less behind given its rate limits.

alanshaw:
This PR removes the pinata cron job (attempting to pin things), and removes the pinata default pin that gets inserted when an upload is created.

The pinata entry in the database service enum remains because a) we have loads of rows referencing that value and b) you can't remove a postgres enum value - you need to drop the whole enum and recreate (which would be really difficult considering the number of rows we have referencing it).

This effectively stops NFT.Storage pinning to Pinata.

closes #260
closes #391


hugomrdias (approved):
LGTM

Issue #2092, opened 2022-07-28
Rate limiting UCAN authorized requests


Gozala:
Our DB got overwhelmed due to vast amount of UCAN authorized requests that did not got rate limited by our rate limiting rules. That is because each request gets unique UCAN token for the invocation so authorization header based rate limiting did not apply.

In a discussion with @hugomrdias and @alanshaw we have came up with a following solution to address this problem:

1. [ ] Require additional header e.g `x-agent-did` with a `did` of the actor issuing requests, that is outmost token issuer.
2. [ ] Deny request without above header in place on a new API endpoint e.g. `/ucan-upload` and rewire UCAN authorized uploads to that endpoint.
3. [ ] Add rate limiting rules based on `x-agent-did` header similar to ones that are in place for `authorization` header on than new API endpoint.
4. [ ] Update docs to reflect changes.
5. [ ] Inform users about changes. Provide info about this changes in the error responses.

Gozala:
As separate item we also decided to followup on the general rate limiting strategies with CF team. Specifically it we had a Web Application Firewall (WAF) API in the workers that we could check against specific inputs we would be able to rate limit from the worker directly by asserting against agent `did` and potentially a root issuer `did` as well. 

We have even considered idea of creating `/waf` endpoint that we could trigger form our workers so we could check if received request should be rate limited or not and if so deny service.

PR #2093, opened 2022-07-28
feat!: rate limiting by ucan agent did

Gozala:
fixes https://github.com/nftstorage/nft.storage/issues/2092 

As per discussion today this changes ucan authorized request route to `/ucan-upload` and requires `x-agent-did` header to apply apply rate limiting.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -31,7 +31,7 @@
     "nanoid": "^3.1.30",
     "regexparam": "^2.0.0",
     "toucan-js": "^2.4.1",
-    "ucan-storage": "^ 1.3.0",
+    "ucan-storage": "^1.3.0",
     "uint8arrays": "^3.0.0"
   },
   "devDependencies": {
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "api",
-  "version": "3.0.1",
+  "version": "3.0.2",
   "description": "NFT Storage API",
   "private": true,
   "type": "module",
@@ -31,7 +31,7 @@
     "nanoid": "^3.1.30",
     "regexparam": "^2.0.0",
     "toucan-js": "^2.4.1",
-    "ucan-storage": "^1.0.0",
+    "ucan-storage": "^ 1.3.0",
     "uint8arrays": "^3.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


packages/api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "api",
-  "version": "3.0.1",
```
hugomrdias, line 3: dont change this manually

packages/api/src/utils/auth.js
```diff
@@ -37,19 +39,35 @@ export async function validate(event, { log, db, ucanService }, options) {
...
-    const user = await db.getUser(root.audience())
-    if (user) {
-      log.setUser({ id: user.id })
-      return {
-        user: filterDeletedKeys(user),
-        db,
-        ucan: { token, root: root._decoded.payload, cap },
-        type: 'ucan',
+  if (Ucan.isUcan(token)) {
+    if (options?.checkUcan) {
```
hugomrdias, line 43: this should be the other way around right ?

hugomrdias (changes requested):

packages/api/package.json
```diff
@@ -1,6 +1,6 @@
...
+  "version": "3.0.2",
```
alanshaw, line 3: This is managed by release please. Please revert.

packages/api/package.json
```diff
@@ -31,7 +31,7 @@
     "nanoid": "^3.1.30",
     "regexparam": "^2.0.0",
     "toucan-js": "^2.4.1",
-    "ucan-storage": "^1.0.0",
+    "ucan-storage": "^ 1.3.0",
```
alanshaw, line 34:
```suggestion
    "ucan-storage": "^1.3.0",
```

alanshaw (commented):
This LGTM.

Please can you also do a small tweak to https://nft.storage/docs/how-to/ucan/ to explain that uploads need to go to a different endpoint and also add the new endpoint to the OpenAPI [spec](https://github.com/nftstorage/nft.storage/blob/main/packages/website/public/schema.yml) (as seen here: https://nft.storage/api-docs/).

packages/api/package.json
Gozala, line 3:
```suggestion
  "version": "3.0.1",
```

packages/api/src/utils/auth.js
Gozala: No this is correct if it is a UCAN but we don't checkUcans we throw error on line 69 to tell user to go to other route instead

alanshaw (approved):

PR #2094, opened 2022-07-28
fix: fix error sourcemaps with wrangler

hugomrdias:
wrangler needs `no_bundle = true` or else its using its own bundling thing....


vasco-santos (approved):
wow, how did this change silently

hugomrdias:
@vasco-santos i cant find the issue about this stuff can you help ?

vasco-santos:
Not sure if there was an issue. I only remember this PR https://github.com/web3-storage/web3.storage/pull/1593

PR #2097, opened 2022-07-29
feat!: rewire ucan endpoint back to /upload

Gozala:


alanshaw (changes requested):
Just noticed:

https://github.com/nftstorage/nft.storage/blob/b29d3d5a431f4775fabd147bc66353956511e8f9/packages/api/src/errors.js#L238-L256

Can you fix `this.code` in both and add missing `ErrorAgentDIDRequired.CODE`.

packages/client/src/lib.js
```diff
@@ -112,15 +113,26 @@ class NFTStorage {
...
 
   /**
    * @hidden
-   * @param {string} token
+   * @param {object} options
+   * @param {string} options.token
+   * @param {string} [options.did]
    */
-  static auth(token) {
+  static auth({ token, did }) {
```
alanshaw, line 129: Signature changed but none of the callers

Gozala: That is because I use spread to capture both e.g see line 170

alanshaw: Oh yes my apologies!

alanshaw (changes requested):
`this.code` in `ErrorDIDNotFound` is still not correct!

packages/api/src/errors.js
```diff
@@ -251,12 +251,7 @@ export class ErrorAgentDIDRequired extends HTTPError {
...
     super(msg, status)
     this.name = 'ErrorAgentDIDRequired'
-    this.code = ErrorUnauthenticated.CODE
-  }
-}
-
-export class ErrorInvalidRoute extends HTTPError {
-  constructor(msg = 'Invalid route for the request', status = 400) {
-    super(msg, status)
+    this.code = ErrorAgentDIDRequired.CODE
```
Gozala, line 254:
```suggestion
    this.name = 'AgentDIDRequired'
    this.code = ErrorAgentDIDRequired.CODE
```

packages/api/src/errors.js
```diff
@@ -242,7 +242,7 @@ export class ErrorDIDNotFound extends HTTPError {
     this.code = ErrorMaintenance.CODE
```
Gozala, line 242:
```suggestion
    this.code = ErrorDIDNotFound.CODE
```

alanshaw (approved):
LGTM

PR #2101, opened 2022-08-01
fix: website workflow release build

vasco-santos:
This PR fixes website workflow build. Automatic build is now disabled in CF, and as a result we need the ENV variables to be specified in the build here instead of tied with CF Dashboard


vasco-santos:
Merging for testing

PR #2108, opened 2022-08-01
fix: fdw fault tolerance

alanshaw:
We sometimes see request failures for calls to FDW. In this case we actually get sent back non-JSON data and our postgrest client fails to parse it and throws i.e. does not _return_ an error.

We're seeing this in the upload flow because we immediately request our upload after it has been created. Our database is not under load, but the FDW database seems to sometimes not be available.

This PR handles the case when the call to FDW fails with a thrown error _not_ a returned one.

499 is [H27 - Client Request Interrupted](https://devcenter.heroku.com/articles/error-codes#h27-client-request-interrupted):

Screenshot 2022-08-01 at 18 09 55

Here is the error our client throws because it receives HTML:

Screenshot 2022-08-01 at 18 10 23

.

### DIFF
--- a/packages/api/src/utils/db-client.js
+++ b/packages/api/src/utils/db-client.js
@@ -498,25 +498,29 @@ export class DBClient {
    * @param {string[]} cids
    */
   async getDealsForCids(cids = []) {
-    const rsp = await this.client.rpc('find_deals_by_content_cids', {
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):
LGTM

PR #2111, opened 2022-08-02
fix: always retry s3 upload

alanshaw:
We're seeing more and more of the following errors. My suspicion is that the S3 client is not (or no longer) returning an error with `name: 'BadDigest'`. This PR changes the behaviour here to just always retry a failed upload once and adds the thrown error as the `cause` of a new `Error` so that we can retain context here (since the stack in sentry is not helpful).

Screenshot 2022-08-02 at 11 35 02


vasco-santos (approved):
LGTM
this s3 client 😢

PR #2130, opened 2022-08-10
chore: replace docker shell script with js CLI commands and use for dev and test environments

### LINKED ISSUE #2121: refactor run-with-dependencies shell script & fold into cli.js
As part of the epic journey that was #2065, I ended up adding a shell script called `run-with-dependencies.sh` to bring the docker compose environment up and down, and a `run-test.sh` script to kick off the test suite. I used shell scripts because they were easy to get going with as I was trying to figure out if testcontainers was the cause of a show-stopping issue, and we merged them in to avoid that PR branch going stale.  This issue is to track replacing them with a JS solution for better maintainability and consistency.

The `run-with-dependencies` script is essentially doing the same thing as `services start` and `services stop`, except that it will run an arbitrary command in between, and export some config variables to point to the right service ports. I really like the "with dependencies" pattern, since it ensures that things always get cleaned up even if things exit abnormally. But I think it may make more sense as a flag, e.g. `--with-services`, or something.

So, my tentative plan is:

- [ ] add new `run-dev` and `run-test` subcommands to the cli, to setup and run the dev server and ava, respectively. 
  - These can either use `execa` + the command line tools, or use the Miniflare / Ava JS APIs.
- [ ] give those new commands a `--with-services` flag, which causes them to be sandwiched between docker compose `up` and `down` calls (using `execa`)
  - this also causes the `DATABASE_URL` and other port-dependent config vars to be overridden and made available to the `run-dev` / `run-test` commands, so they can use the dynamic values instead of whatever is in `.env`
- [ ] also add a `--persist` flag, to enable persistent data volumes.

This seems like a pretty flexible solution - you can run dev and test either with or without an automatic environment, so if you have a custom set of volumes for debugging or something, you can set that up manually and run `cli.js run-dev` without the `--with-services` flag. But the npm scripts would include the flag, so by default everything is nicely automatic and isolated.

Anyone with ideas for improvement or a better plan for hooking this up should please chime in!

yusefnapora:
This closes #2121 by replacing the `run-with-dependencies.sh` script I added with a `runWithServices` function in `scripts/cmds/services.js`.

This does essentially the same thing as the shell script, which is to run `docker compose up` to bring up the services, get the local ports for the running services and override some config variables, then run the action you give it.

I replaced the existing `services start` and `services stop` commands with `services exec`, which does the `runWithServices` thing to an arbitrary shell command you give it. The motivation here is to allow using a unique docker compose project name for each run (except in persistent mode, where the project name is always `nft-storage-dev-persistent`). With separate `start` and `stop` commands, you would need to write the project name to a file or something to keep track of it between invocations, but with the start and cleanup commands in the same process you can just keep the project name in memory. If you want something like the old behavior (because you want to manually poke at the db, say), you can run `cli.js services exec bash` to run a shell, and then run whatever you want inside the shell.

I also wired this up to `yarn dev` and added a new `yarn dev:persist` command, which I believe meets the requirements we were trying to hit in #1888, but with less boilerplate and without putting miniflare or the test runner into a container.

The persistent mode works by creating named docker volumes on the host, and using the `docker-compose-volumes.yml` file to point the services at those long-lived volumes instead of the ephemeral volumes specified in the main `docker-compose.yml`.


yusefnapora:
cc @funkytek - would be great if you guys could play with this branch a bit and see if there are any rough spots I missed :)

PR #2137, opened 2022-08-18
feat: cron runs nft-ttr twice as often (every 15min)

gobengo:
Motivation:
* I want to have more frequent metrics in grafana dashboard
* I want to verify that changing this is all that's required to increase metric frequency (it's possible the constraint is actually based on how often prometheus scrapes the pushgateway)


Gozala (approved):

olizilla:
the scrape interval for prom to fetch metrics from push gateway is 5m at time of writing defined here https://github.com/protocol/monitoring-infra/blob/master/ansible/roles/prometheus-operator/templates/prometheus-config.yml#L615 (private repo, sorry)

PR #2145, opened 2022-08-22
chore: change npm token var

hugomrdias:
after we merge this we can delete the repo NPM_AUTH_TOKEN secret


hugomrdias:
> Neat. I think it is wise to use the org level secrets here. Is NPM_TOKEN set to be a `automation` token for the new dig-house npm bot?

yes

PR #2151, opened 2022-08-24
feat(utils): add timers to logging

ericssontadashi:
migrating from web3 logging

### DIFF
--- a/packages/api/src/utils/logs.js
+++ b/packages/api/src/utils/logs.js
@@ -13,6 +13,21 @@ const buildMetadataFromHeaders = (/** @type {Headers} */ headers) => {
   return responseMetadata
 }
 
+/**
+ * @typedef {'log' | 'debug' | 'info' | 'warn' | 'error' | 'time'} Level
+ * @typedef {{
+ * dt: string;
+ * level: Level;
+ * context?: any;
+ * metadata: any;
+ * message: string;
+ * stack?: string;
+ * duration?: number}} LogEntry
+ */
+
+/**
+ * Logging
+ */
 export class Logging {
   /**
    * @param {FetchEvent} event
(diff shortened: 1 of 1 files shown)


vasco-santos (commented):
The set of changes look good to me, but would be great to know more about the motivation of this PR, so that I can better support in case other solutions exist.

As a side note, I would really prefer if we simply drop this file and used the common code in workers/utils https://github.com/web3-storage/workers/blob/master/packages/utils/src/logging.js

From what I see, it already has what is in here?

packages/api/src/utils/logs.js
```diff
@@ -268,9 +298,25 @@ export class Logging {
     if (this.opts?.debug) {
       console.log(`[${this._date()}] `, `${name}: ${duration} ms`)
     }
-    return timeObj
```
alanshaw, line 271: No return statement anymore?

packages/api/src/utils/logs.js
```diff
@@ -237,28 +266,29 @@ export class Logging {
...
     this._timesOrder.push(name)
   }
 
   /**
    * @param {string} name
+   * @param {boolean} [shouldLog]
+   * @param {any} [context]
    */
-  timeEnd(name) {
+  timeEnd(name, shouldLog, context) {
```
alanshaw, line 284: This should match the [`timeEnd`](https://developer.mozilla.org/en-US/docs/Web/API/console/timeEnd) method.

packages/api/src/utils/logs.js
```diff
@@ -268,9 +298,25 @@ export class Logging {
     if (this.opts?.debug) {
       console.log(`[${this._date()}] `, `${name}: ${duration} ms`)
     }
-    return timeObj
+
+    if (shouldLog) {
```
alanshaw, line 302: Note that timing execution is not effective in production and I believe that is why this method does not send logs to logtail.

https://developers.cloudflare.com/workers/learning/security-model/#step-1-disallow-timers-and-multi-threading

alanshaw (changes requested):

hugomrdias:
yes @ericssontadashi exactly what @vasco-santos said drop this file and use https://github.com/web3-storage/workers/blob/master/packages/utils/readme.md#logging directly if you have problems or questions doing that ping me

Issue #2138, opened 2022-08-18
Can create NFT API Key with same name and cant delete it 


ktaware1:
![Screenshot (10)](https://user-images.githubusercontent.com/108401277/185399099-e0b3a4e2-f0bb-405f-bed5-7a7164096a68.png)
This is creating a problem to delete and have same API Key and API Name

PR #2158, opened 2022-08-26
fix: duplicate key name

drewdelano:
closes https://github.com/nftstorage/nft.storage/issues/2138

This includes 3 fixes each of which should fix the issue
1. Enable the fix the popup menus so they're distinct
2. Add a unique constraint to the database to prevent duplicates (this might be a bad idea if there are duplicates already in PROD)
3. Added logic to check if a duplicate already exists and message that in the API layer

### DIFF
--- a/packages/api/src/utils/db-client.js
+++ b/packages/api/src/utils/db-client.js
@@ -525,6 +525,18 @@ export class DBClient {
     /** @type {PostgrestQueryBuilder<definitions['auth_key']>} */
     const query = this.client.from('auth_key')
 
+    const { data: dataExists } = await query
+      .select()
+      .match({
+        user_id: key.userId,
+        name: key.name,
+      })
+      .is('deleted_at', null)
+
+    if (dataExists?.length) {
+      throw new Error('Duplicate key name.')
+    }
+
     const { data, error } = await query
       .upsert({
         name: key.name,
--- a/packages/website/pages/manage.js
+++ b/packages/website/pages/manage.js
@@ -99,6 +99,14 @@ export default function ManageKeys({ user }) {
     </Button>
   )
 
+  /**
+   * @param {string} keyName
+   * @param {number} keyIndex
+   */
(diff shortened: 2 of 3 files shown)


packages/website/pages/manage.js
```diff
@@ -99,6 +99,14 @@ export default function ManageKeys({ user }) {
     </Button>
   )
 
+  /**
+   * @param {string} keyName
+   * @param {number} keyIndex
+   */
+  const getButtonKey = (keyName, keyIndex) => {
```
dashcraft, line 106: Suggestion (non-blocking):
I don't know if this is particularly valuable, it sorta just indirectly does what string template literals do already.
I feel like it wouldn't be too much to just do this check inline, but i'd leave it to preference.

packages/api/src/utils/db-client.js
```diff
@@ -525,6 +525,18 @@ export class DBClient {
     /** @type {PostgrestQueryBuilder<definitions['auth_key']>} */
     const query = this.client.from('auth_key')
 
+    const { data: dataExists } = await query
```
dashcraft, line 528: comment (blocking):
Let's get a test to check logic and behavior please!

dashcraft (changes requested):
Might also run prettier against this file too, some spaces are misaligned, etc. This file is kinda rough in general

packages/website/pages/manage.js
drewdelano: it's used in 4 places so i was just going with reuse here.  i could explode it to just do templating, but this seemed easier to maintain (one place to edit, instead of 4).

packages/api/src/utils/db-client.js
drewdelano: done!

dashcraft (approved):
LGTM!

drewdelano:
Dropped DB changes commit

PR #2159, opened 2022-08-29
chore: re-enable skipped test & workaround miniflare formdata issue

### LINKED ISSUE #2088: re-enable skipped test when miniflare bug is fixed
This issue is a reminder to re-enable the test in `nfts-store.spec.js` that was disabled by #2065 once the issue with file paths not being preserved in Miniflare is sorted.

Revisit once https://github.com/cloudflare/miniflare/pull/309 is merged (or the bug is fixed some other way :smile:).

yusefnapora:
This closes #2088 by re-enabling the test we were skipping, now that miniflare preserves the full file path.

It also works around a bug I found while updating. Miniflare v2.7+ fails when you provide an instance of `FormData` from `@web-std/form-data` as the body of a `dispatchFetch` call. It also seems to fail when using the `FormData` built into node 18+. It likes `undici`'s `FormData` though :smile: 

Miniflare updated their `undici` dependency in the 2.7 release, so that seems likely to be the cause of the issue. It manifests as the body being sent with a content-type of `text/plain;charset=utf-8` instead of `multipart/form-data`, so the api route handler doesn't know what to do with it.

In `nfts-upload.spec.js` we were constructing a `FormData` directly, so the fix is just to import `FormData` from undici instead of using the platform version. The tests in `nfts-store.spec.js` use the `Token.encode` fn from the client. I didn't want to make changes to the client just to make API tests happy, so I added a little helper to convert the `Token.encode` output to an undici `FormData`.

BTW, this will likely affect any clean builds from `main`, since we currently depend on miniflare `^2.5.0`, so if you don't have a lockfile it will resolve to the latest 2.x release.

### DIFF
--- a/packages/api/test/nfts-store.spec.js
+++ b/packages/api/test/nfts-store.spec.js
@@ -9,6 +9,7 @@ import {
   setupMiniflareContext,
 } from './scripts/test-context.js'
 import { File, Blob } from 'nft.storage/src/platform.js'
+import { FormData } from 'undici'
 
 test.beforeEach(async (t) => {
   await setupMiniflareContext(t)
@@ -34,7 +35,7 @@ test('should store image', async (t) => {
       ],
     },
   }
-  const body = Token.encode(metadata)
+  const body = fixFormData(Token.encode(metadata))
 
   const res = await mf.dispatchFetch('http://miniflare.test/store', {
     method: 'POST',
@@ -94,9 +95,7 @@ test('should store image', async (t) => {
   ])
 })
 
-// Miniflare doesn't currently preserve pathnames in File objects uploaded via FormData.
-// TODO: re-enable this test once https://github.com/cloudflare/miniflare/pull/309 is merged
-test.skip('should store dir wrapped image', async (t) => {
+test('should store dir wrapped image', async (t) => {
   const config = getTestServiceConfig(t)
   const mf = getMiniflareContext(t)
   const { token, userId } = await createTestUser(t)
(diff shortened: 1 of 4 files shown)


packages/api/test/nfts-store.spec.js
```diff
@@ -94,19 +95,22 @@ test('should store image', async (t) => {
...
+test('should store dir wrapped image', async (t) => {
   const config = getTestServiceConfig(t)
   const mf = getMiniflareContext(t)
   const { token, userId } = await createTestUser(t)
 
   const metadata = {
     name: 'name',
     description: 'stuff',
-    image: new File(['fake image'], '/dir/cat.png', { type: 'image/png' }),
+    // Note that the multipart form parser used by miniflare does not remove
```
hugomrdias, line 106: did you report this upstream ?

packages/api/test/nfts-store.spec.js
```diff
@@ -156,3 +160,20 @@ test.skip('should store dir wrapped image', async (t) => {
...
+
+/**
+ * Current versions of Miniflare (2.7.1 as of this writing) don't set the
+ * correct content type when given an instance of FormData from the `@web-std/form-data`
+ * package. Until we sort out the root cause, this just converts from the `@web-std` FormData
+ * object to the one from `undici`, which is what miniflare uses internally.
+ *
+ * @param {FormData} fd
+ */
+function fixFormData(fd) {
```
hugomrdias, line 172: we should probably update @web-std to use undici formdata

hugomrdias (changes requested):

packages/api/test/nfts-store.spec.js
yusefnapora: good idea :smile: https://github.com/cloudflare/miniflare/issues/352

Also made one about the FormData problem: https://github.com/cloudflare/miniflare/issues/351

olizilla:
Oh my. I've just spent a day trying to figure what the heck was going on with our tests failing when I had to update miniflare to get r2 support for #2204

So happy you found the fix @yusefnapora ! let's get it merged!

olizilla (approved):
🚀 nice digging!

PR #2185, opened 2022-09-14
fix: fix typescript nodenext resolution

rayzr522:
tiny fix here: while your ESModule build is working totally fine, when downstream consumers are set up to use `nodenext` for their `module` & `moduleResolution` settings, typescript reads your `package.json` and sees that the `exports.node` points to the `src` folder and tries to also look there for the typescript typings

as per [the typescript docs](https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing) this just lets typescript know to still keep looking in the `dist/src` folder for the types :)

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -16,16 +16,19 @@
     ".": {
       "browser": "./src/lib.js",
       "require": "./dist/src/lib.cjs",
+      "types": "./dist/src/lib.d.ts",
       "node": "./src/lib.js"
     },
     "./src/platform.js": {
       "browser": "./src/platform.web.js",
       "require": "./dist/src/platform.cjs",
+      "types": "./dist/src/platform.cjs",
       "node": "./src/platform.js"
     },
     "./src/token.js": {
       "browser": "./src/token.js",
       "require": "./dist/src/token.cjs",
+      "types": "./dist/src/token.cjs",
       "node": "./src/token.js"
     },
     "./dist/bundle.esm.min.js": {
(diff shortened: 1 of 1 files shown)


rayzr522:
bump! could we get this merged?

packages/client/package.json
```diff
@@ -16,16 +16,19 @@
     ".": {
       "browser": "./src/lib.js",
       "require": "./dist/src/lib.cjs",
+      "types": "./dist/src/lib.d.ts",
       "node": "./src/lib.js"
     },
     "./src/platform.js": {
       "browser": "./src/platform.web.js",
       "require": "./dist/src/platform.cjs",
+      "types": "./dist/src/platform.cjs",
```
alanshaw, line 25:
```suggestion
      "types": "./dist/src/platform.d.ts",
```

packages/client/package.json
```diff
@@ -16,16 +16,19 @@
...
       "node": "./src/platform.js"
     },
     "./src/token.js": {
       "browser": "./src/token.js",
       "require": "./dist/src/token.cjs",
+      "types": "./dist/src/token.cjs",
```
alanshaw, line 31:
```suggestion
      "types": "./dist/src/token.d.ts",
```

alanshaw (changes requested):

rayzr522:
@alanshaw thanks for the review! can't believe I missed it til now, and all just because of such a silly typo hehe

rebased on main & made the changes you requested :3

alanshaw (approved):

PR #2204, opened 2022-09-26
feat: put to r2 and check linkdex-api for dag structure

olizilla:
Porting https://github.com/web3-storage/web3.storage/pull/1785

### Adds
- put CARs to R2 in parallel with S3. Both must succed for a succesful upload, so we add p-retry around both
- linkdex-api to fetch dag completeness report over CARs in S3, to replace what ipfs-cluster was doing for us.
- `updatePinStatus` to the db client, so we can update the status immediately if linkdex-api tells us the DAG is now complete... previously this only happened via a direct db update in a cron job.
- ElasticIpfs as an entry in the ClusterService enum, and changes so that we add a "pin" record for the one we have on e-ipfs... the e-ipfs pin will now be the only one for an upload, and it's status will determine if we tell the user it is "Pinned" or "Pinning"... and we can now find the dag status immediately after upload, so the time to mark a thing as Pinned will be massively reduced.
- mocking fetch requests in miniflare, to simulate linkdex-api responses. See: https://miniflare.dev/core/standards#mocking-outbound-fetch-requests

adds env vars
- `CARPARK` - an R2 Bucket binding
- `CARPARK_URL` - the public url prefix to use when recording the r2 backup_url
- `LINKDEX_URL` - url for linkdex-api

### Removes
- pinning to ipfs-cluster after upload completes. It's all e-ipfs now.

## Notes

- It looks like we've been using the wrong `dagSize` for things uploaded via `/store`. Uploading a single CBOR block CAR to cluster and asking it for the dag size just gives you the total bytes of the blocks in the CAR... not the size of the reachable dag from that root. This PR will change that to store the full dagSize as the sum of the size of all the blocks.
see: https://github.com/ipfs-cluster/ipfs-cluster/blob/f376cf5106deeeb903b58e7e2431fa63bdea6900/adder/adder.go#L303

## TODO

- [x] Add `LINKDEX_URL` to env
  - [x] on staging
  - [x] on prod
- [ ] Run the DB migration to add ElasticIpfs as a cluster service [packages/api/db/migrations/007-add-service.sql](https://github.com/nftstorage/nft.storage/pull/2204#diff-ce523ba6159cba899351f139330b643178f3a1bcf102ad006eb466c7f33b361f)
  - [x] on staging
  - [ ] on prod

License: MIT


olizilla:
Needs #2159 to get tests working

packages/api/scripts/run-test.sh
```diff
@@ -9,10 +9,6 @@ $CLI db-sql --cargo --testing --reset
 echo "creating minio bucket..."
 $CLI minio bucket create dotstorage-dev-0
 
-cd $THIS_DIR/..
```
olizilla, line 12: Moved out `tsc` checking to a `posttest` run script to separate it from our tests, but still have it run after the tests have run by the magic of npm lifecycle hook scripts, so I can run the tests without appeasing tsc first.

packages/api/src/utils/db-types.d.ts
```diff
@@ -18,6 +18,7 @@ export interface paths {
         query: {
           user_id?: parameters['rowFilter.admin_search.user_id']
           email?: parameters['rowFilter.admin_search.email']
+          github_id?: parameters['rowFilter.admin_search.github_id']
```
olizilla, line 21: This PR didn't change this field, but I had to re-generate this massive db-types file as I've added a new `ElasticIpfs` option to the ClusterService enum...

packages/api/package.json
```diff
@@ -8,10 +8,11 @@
   "scripts": {
     "deploy": "wrangler publish --env production",
     "predev": "./scripts/cli.js services start && ./scripts/cli.js db-sql --cargo --testing --reset && ./scripts/cli.js minio bucket create dotstorage-dev-0",
-    "dev": "miniflare dist/worker.js --watch --debug --env ../../.env",
+    "dev": "miniflare dist/worker.js --watch --debug --env ../../.env --r2 CARPARK",
     "dev:persist": "PERSIST_VOLUMES=true npm run dev",
     "build": "scripts/cli.js build",
     "test": "./docker/run-with-dependencies.sh ./scripts/run-test.sh",
+    "posttest": "tsc",
```
vasco-santos, line 15: nit: we usually do this in pretest. I have no strong opinion, just raising for keeping things uniform

packages/api/wrangler.toml
```diff
@@ -14,6 +14,11 @@ no_bundle = true
 ENV = "dev"
 DEBUG = "true"
 DATABASE_URL = "http://localhost:3000"
+CARPARK_URL = "https://carpark-dev.web3.storage"
+
+[[r2_buckets]]
```
vasco-santos, line 19: we need prod and staging

packages/api/test/nfts-upload.spec.js
```diff
@@ -203,6 +208,95 @@ test.serial('should upload a single CAR file', async (t) => {
...
+
+    const pin = await pRetry(
+      async () => {
+        const { data: pin } = await db
+          .from('pin')
+          .select('*')
+          .match({ content_cid: cid, service: 'ElasticIpfs' })
+          .single()
+        if (pin.status !== 'Pinned') {
+          throw new Error('retry, the status is updated in a waitUntil')
```
vasco-santos, line 288: we can await on waitUntil with MF (search waitUntil in https://miniflare.dev/core/fetch/#http-requests)

packages/api/test/nfts-upload.spec.js
```diff
@@ -438,6 +537,10 @@ test.serial(
       maxChunkSize: chunkSize,
     })
     const splitter = await TreewalkCarSplitter.fromBlob(car, chunkSize)
+    const linkdexMock = getLinkdexMock(t)
+    // return repsonse "Partial" 5 times, then 'Complete once
```
vasco-santos, line 541:
```suggestion
    // return response "Partial" 5 times, then 'Complete once
```

vasco-santos (changes requested):
Great work. Some minor things can be improved, and wrangler toml is missing prod+staging buckets

packages/api/test/nfts-upload.spec.js
olizilla: oh nice! that is better. Fixing!

DEVELOPMENT.md
```diff
@@ -6,7 +6,7 @@ This doc should contain everything you need to know to get a working development
 
 You'll need at least the following:
 
-- Node.js v16+
```
yusefnapora, line 9: Thanks for updating this!

packages/api/wrangler.toml
olizilla: Added config for prod and staging buckets

yusefnapora (approved):
This looks great. Thanks for updating the config stuff - I definitely wasn't expecting non-string global values from miniflare when I wrote that.

So cool to not have cluster in the loop!

vasco-santos (approved):
LGTM

alanshaw (changes requested):
🎉 Awesome!

There's two other files where uploads happen: [`metaplex-upload.js`](https://github.com/nftstorage/nft.storage/blob/main/packages/api/src/routes/metaplex-upload.js) which I think is fine, but worth double checking if you haven't already and [`nfts-store.js`](https://github.com/nftstorage/nft.storage/blob/main/packages/api/src/routes/nfts-store.js) which we need to deprecate (the client builds this now and sends a CAR) however there are quite a lot of folks still using it:
 
Screenshot 2022-10-11 at 12 10 47

I can't remember why this wasn't switched to creating and uploading a CAR to S3 when the rest was. Probably just forgotten.

Everything else seems great aside from some minor comments.

packages/api/db/migrations/007-add-service.sql
```diff
@@ -0,0 +1,13 @@
+-- Service type is the place/location/organisation that is pinning the content.
+CREATE OR UPDATE TYPE service_type AS ENUM (
```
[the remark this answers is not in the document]
olizilla: Done!

packages/api/src/routes/nfts-upload.js
```diff
@@ -162,20 +178,23 @@ export async function uploadCarWithStat(
...
-        await cluster.pin(sourceCid)
-      } catch (err) {
-        console.warn('failed to pin to cluster', err)
-      }
-      await cluster.addCar(car, { local: true })
-    })()
-  )
+  // ask linkdex for the dag structure across the set of CARs in S3 for this upload.
+  const checkDagStructureTask = async () => {
+    const structure = await ctx.linkdexApi?.getDagStructure(s3Backup.key)
```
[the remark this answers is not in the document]
olizilla: tsc not smart enough here. Only happy if I caputre lindexApi as a block local const after the falsey check.
Screenshot 2022-10-12 at 15 38 48

packages/api/src/utils/db-client.js
```diff
@@ -209,6 +209,33 @@ export class DBClient {
...
 
+  /**
+   * @param {string} contentCid
+   * @param {import('./db-client-types').CreateUploadInputPin} newPin
+   * @returns
+   */
+  async updatePinStatus(contentCid, { service, status: newStatus }) {
+    const now = new Date().toISOString()
+    const query = this.client.from('pin')
+    const { data: pin, status } = await query
```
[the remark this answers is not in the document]
olizilla: Oh wat. Yes it absolutely should. I've managed to get this far without using postgres-js, so looked at the existing `updateUpload` function for the pattern... I can't think of a good readon why this one doesn't need to check for  `error` as well.

https://github.com/nftstorage/nft.storage/blob/ce81d2a88eeaaa97344b909105900b63894c6d9b/packages/api/src/utils/db-client.js#L218-L246

packages/api/src/routes/nfts-upload.js
```diff
@@ -132,7 +132,7 @@ export async function uploadCar(params) {
 }
 
 /**
- * @param {UploadCarInput} data
+ * @param {Omit<UploadCarInput, 'car'>} data
```
olizilla, line 135: the `car` property is no longer needed by `uploadCarWithStat` as `CarStat` now provides the `carBytes` as a `UInt8Array`

packages/api/src/utils/car.js
```diff
@@ -8,7 +8,7 @@ import { sha256 } from 'multiformats/hashes/sha2'
 
 /**
  * @param {CID[]} roots
- * @param {AsyncIterable<Block>|Iterable<Block>} blocks
+ * @param {AsyncIterable<Block>|Iterable<Block>|AsyncGenerator<{cid: CID, bytes: Uint8Array}>} blocks
```
olizilla, line 11: No longer needed, reverting!

alanshaw: I _think_ you could have used `AsyncIterableIterator<Block>` here instead of all those.

alanshaw (approved):
Lovely job.

packages/api/src/routes/nfts-store.js
```diff
@@ -1,83 +1,213 @@
...
-      setIn(dag, path, cid)
-      files.push({ name: file.name, type: file.type })
+  // accumlate blocks here, until we are ready to export it as a CAR
+  const bs = new MemoryBlockStore()
+
+  for (const [name, content] of form.entries()) {
+    if (name === 'meta') {
+      continue
+    }
+    const file = /** @type {File} */ (content)
```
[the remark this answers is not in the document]
olizilla: ah yes, nice. fixing.

PR #2205, opened 2022-09-26
feat: update stats page

e-schneid:


packages/website/pages/stats.js
```diff
@@ -109,30 +109,97 @@ export default function Stats({ logos }) {
...
+            </p>
+            <p className={pClass}>
+              Of course NFTs can be more than just art, but NFTs without
+              trustless storage fall into the same traps that web3 is aiming to
+              solve.
+            </p>
+          </div>
+        </div>
+        <div className="flex lg:w-1/2 items-center justify-center pl-2">
+          <Img
```
johnathonroach, line 139: @e-schneid for some reason, the `decentralized.png` image is showing as "not found" on the preview link. It is using the cloudflare image loader and my guess is that it may only be effecting the preview instance but I wanted to point this out.

prm-eric-schneider: thanks @johnathonroach .  I'll do some research to make sure this is the case

e-schneid: thanks for testing it out!  I'll do some research to make sure it's just the preview.

PR #2218, opened 2022-10-03
chore: add note about /store endpoint perf & recommend /upload

yusefnapora:
This adds a note to the HTTP API docs about `/store` being slower than `/upload`, since we had a few people surprised at the `/store` perf earlier.


dchoi27 (approved):
made some suggested changes, feel free to revert or edit text if you disagree

Issue #2202, opened 2022-09-26
More browser friendly rate limits/IP block lengths


zpalmtree:
Hello,

I would like to request that the rate limits for the `nftstorage.link` domain be slightly altered. 

The current rate limits appear to be 200 requests per minute, or 3.33 requests per second.

> The NFT.Storage gateway is currently rate limited to 200 requests per minute for a given IP address. In the event of a rate limit, the IP will be blocked for 30 seconds.

I have an app which fetches lots of NFTs in browser, and if I accidentally hit the rate limit, a 30 second delay makes for a quite poor experience for my users - I have to block loading for 30 seconds until the rate limit expires.

I would like to request that instead of applying the rate limit over a minute period, with a 30 second block period, perhaps this be lowered to a 10 second period, with a, for example, 5 second block period.

The requests per second would still stay the same of course, so 33 requests per 10 second period.

This change would allow me to provide a much nicer experience for my users - if I accidentally hit the rate limit, I can slow down my fetches and resume loading data only 5 seconds later.

vasco-santos:
Hey @zpalmtree 

We would love to decrease the blocking interval (and that docs is wrong, it is actually 1min... @zpalmtree can you point me to where in docs it is?).

The core issue is the firewall provider we use for rate limiting (CF only allows us to block for the duration of time it is tracked or higher). So, we can't for example has 200 requests per minute, with 15s blocking.

Keeping the same throughput, but reducing the interval would be an option. But there are also cons on that approach. If we consider a web page that loads and performs right away 80 requests to the gateway to load their resources would be problematic as some would be blocked... We can try to get a better balance like 100 req/30s, but we need to evaluate the implications of that. Currently other IPFS gateway providers have also have same 1 minute and we should do a study together.

For what is worth, we are working on ways of getting higher rate limits. That will probably be the way to go

zpalmtree:
Here is the link I saw that snippet: https://nft.storage/docs/concepts/gateways/#rate-limits

PR #2231, opened 2022-10-17
fix: docs with incorrect blocking duration due to rate limits

vasco-santos:
I would to be 30 seconds (I think 🤔 ) but now CF Firewall does not allow blocking duration smaller than the interval using to track rate limit. In other words, if we say 200req/min we need to block in same time unit (i.e 60 seconds).

There were some thoughts on doing some changes to accomodate this, like 100req/30s but we might hit other issues and need to think about this more carefully - read more at https://github.com/nftstorage/nft.storage/issues/2202#issuecomment-1273543142

Closes https://github.com/nftstorage/nft.storage/issues/2202


yusefnapora (approved):
Thanks @vasco-santos

PR #2233, opened 2022-10-17
chore: api Logging#setUser expects user.id as string instead of number

gobengo:
Motivation:
* this way `Logging#setUser` can accomodate user ids that correspond to bigints
* avoid this mistake https://github.com/web3-storage/web3.storage/issues/2040

Related
* I noticed that the db-types user definition has `id: number`. I wonder if it should change to `bigint` or `text` (and underlying queries will need to change). https://github.com/nftstorage/nft.storage/blob/6fcd38923506965012505473b57afde2ce22973e/packages/api/src/utils/db-types.d.ts#L1188
  * nft.storage prod counter isn't at risk of bumping into this soon, but it is a bigint underneath so it could happen.

### DIFF
--- a/packages/api/src/utils/logs.js
+++ b/packages/api/src/utils/logs.js
@@ -46,7 +46,7 @@ export class Logging {
     }
     this.metadata = {
       user: {
-        id: 0,
+        id: '0',
       },
       request: {
         url: request.url,
@@ -68,10 +68,10 @@ export class Logging {
    * Set user
    *
    * @param {Object} user
-   * @param {number} [user.id]
+   * @param {string} user.id
    */
   setUser(user) {
-    this.metadata.user.id = user.id || 0
+    this.metadata.user.id = user.id
     this.opts.sentry.setUser({
       id: `${user.id}`,
     })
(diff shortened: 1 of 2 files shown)


yusefnapora (approved):

olizilla:
> I noticed that the db-types user definition has id: number. I wonder if it should change to bigint or text (and underlying queries will need to change).

I've not dug in, but my guess is we need to fix the schema so it's handled as text or else the timebomb remains active.

packages/api/src/utils/logs.js
```diff
@@ -68,10 +68,10 @@ export class Logging {
    * Set user
    *
    * @param {Object} user
-   * @param {number} [user.id]
+   * @param {string} user.id
    */
   setUser(user) {
-    this.metadata.user.id = user.id || 0
+    this.metadata.user.id = user.id
```
olizilla, line 74: minor: this removes the default, so we'll start to pass `''` instead of `'0'` where we don't have a user. Both seem odd... I think we should make the call to `this.opts.sentry.setUser` if we have a user id, otherwise skip it.

olizilla: it's probably not an issue in practice as it's all just metadata to sentry, and it has to do a bunch of work to filter out noise from reports anyway. I suspect it would impact our "# users affected" reports for a given bug... with the default to `0` i suspect we would see issues that impact unknown users all logged as impacting just 1.

PR #2234, opened 2022-10-17
fix: update toucan-js to report error cause

olizilla:
toucan-js now reports error.cause
see: https://github.com/robertcepa/toucan-js/releases/tag/v2.7.0

License: MIT


olizilla:
Currently error.cause is not sent to sentry. This update means we can find out what actually went wrong when an error has a cause.

alanshaw (approved):

PR #2238, opened 2022-10-18
feat: write DUDEWHERE link and SATNAV car indexes.

olizilla:
- write a CARv2 index so we can use the `SATNAV` bucket to find where Blocks are.
- write a key to map root CID to car CID so we can use the `DUDEWHERE` bucket to find where our CAR is in `CARPARK` (for a given root CID).

This enables [freeway](https://github.com/web3-storage/freeway) to lookup CAR CID(s) for a given root data CID and UnixFS export data directly from R2.

port of https://github.com/web3-storage/web3.storage/pull/2035

License: MIT

### DIFF
--- a/packages/api/src/utils/uploader/r2-uploader.js
+++ b/packages/api/src/utils/uploader/r2-uploader.js
@@ -9,11 +9,11 @@ import { createCarIndex } from '../car.js'
  */
 export class R2Uploader {
   /**
-   * @param {object} opts
-   * @param {R2Bucket} opts.carpark
-   * @param {R2Bucket} opts.dudewhere
-   * @param {R2Bucket} opts.satnav
-   * @param {string} opts.publicUrl
+   * @param {object} config
+   * @param {R2Bucket} config.carpark
+   * @param {R2Bucket} config.dudewhere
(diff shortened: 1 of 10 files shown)


packages/api/src/utils/uploader/r2-uploader.js
```diff
@@ -1,27 +1,28 @@
...
  */
 export class R2Uploader {
   /**
-   * @param {R2Bucket} bucket
-   * @param {string} publicUrl
+   * @param {object} opts
+   * @param {R2Bucket} opts.carpark
+   * @param {R2Bucket} opts.dudewhere
+   * @param {R2Bucket} opts.satnav
+   * @param {string} opts.publicUrl
```
alanshaw, line 16: When it's not optional options, I like to call it config...

```suggestion
   * @param {object} config
   * @param {R2Bucket} config.carpark
   * @param {R2Bucket} config.dudewhere
   * @param {R2Bucket} config.satnav
   * @param {string} config.publicUrl
```

alanshaw (approved):

vasco-santos (approved):
🕺🏼

PR #2240, opened 2022-10-18
ci: add w3s.link to nft-ttr cronjob

gobengo:
Motivation:
* so we measure this gateway compared to others

### DIFF
--- a/.github/workflows/cron-nft-ttr.yml
+++ b/.github/workflows/cron-nft-ttr.yml
@@ -34,6 +34,7 @@ jobs:
             --minImageSizeBytes=10000000 \
             --gateway https://nftstorage.link \
             --gateway https://dweb.link \
+            --gateway https://w3s.link \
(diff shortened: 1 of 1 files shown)


olizilla:
I think we'd expect pretty similar perf from nftstorage.link and w3s.link, @vasco-santos what do you think? would it be useful to have time to availability metrics for both?

vasco-santos:
oh, I did not know this cron existed, nice!

Yes, I would pitch not to add w3s.link in. We expect exactly same performance in both for first read given they are essentially the same codebase behind the scenes https://github.com/web3-storage/reads/tree/main/packages/edge-gateway

Also, I like that we "warm" up the cache and that is another reason to not add w3s.link. At the moment, we rely on 2 domains with their own setup because each one offers us a Cache LRU (one per domain), so getting stuff from w3s.link that likely will be requested from nftstorage.link will mean that we will be removing things from the LRU more relevant for w3s.link.

PR #2265, opened 2022-11-21
feat: add plausible analytics

cmunns:
Wire up plausible and keep count.ly while we test this implementation.


drewdelano (approved):
LGTM

PR #2341, opened 2023-03-09
feat: pins-add catches 400 from cluster and logs to sentry

gobengo:
Motivation:
* get `responseText` [in sentry so I can tell a bit more about what's going on](https://protocol-labs-it.sentry.io/issues/3226038374/?query=is%3Aunresolved&referrer=issue-stream)
* got a grafana alert about errors and I looked in sentry and there were 100+ in the last hour so figured this would give me more visibility next time it happens


gobengo:
Didn't realize those build errors.
I donno if this is even a good idea, but if anyone wants this, comment and I'll pick it up again.

PR #2351, opened 2023-03-22
feat: send `delegates` on pin add response

olizilla:
allows undialable nodes to connect to us when pinning via the pinning service api

see: https://github.com/web3-storage/web3.storage/pull/2255
see: https://github.com/web3-storage/pickup/pull/134

License: MIT

### DIFF
--- a/packages/api/src/routes/pins-add.js
+++ b/packages/api/src/routes/pins-add.js
@@ -46,9 +46,11 @@ export async function pinsAdd(event, ctx) {
     )
   }
 
-  await cluster.pin(cid.sourceCid, {
+  const pinRes = await cluster.pin(cid.sourceCid, {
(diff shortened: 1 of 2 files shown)


packages/api/src/routes/pins-add.js
```diff
@@ -67,5 +69,6 @@ export async function pinsAdd(event, ctx) {
     name: pinData.name,
   })
 
-  return new JSONResponse(toPinsResponse(upload))
+  // @ts-expect-error cluster client types expect Record<string, string> but delegates is string[]
```
olizilla, line 72: i can fix this in a subsequent PR, but i'd like to get this out asap

alanshaw (approved):

PR #2396, opened 2023-05-22
feat: custom chunk size

alanshaw:
Since we upped the default chunk size, folks uploading directories with thousands of small files are sometimes finding the worker exceeds resource limits 🤦‍♂️. This allows them to `encodeDirectory` and then `storeCar` with a smaller chunk size to allow the directory to upload successfully.

e.g.

```js
import { NFTStorage } from 'nft.storage'
import { filesFromPaths } from 'files-from-path'

const token = 'API_KEY' // your API key from https://nft.storage/manage

async function main() {
  const storage = new NFTStorage({ token })
  const files = await filesFromPaths(['/path/to/files'])
  
  const { cid, car } = await NFTStorage.encodeDirectory(files)
  console.log(`File CID: ${cid}`)

  console.log('Sending file...')
  await storage.storeCar(car, {
    maxChunkSize: 1024 * 1024 * 10, // 10MB
    onStoredChunk: (size) => console.log(`Stored a chunk of ${size} bytes`)
  })

  console.log('✅ Done')
}
main()
```

### DIFF
--- a/packages/client/src/lib.js
+++ b/packages/client/src/lib.js
@@ -169,11 +169,11 @@ class NFTStorage {
   static async storeCar(
     { endpoint, rateLimiter = globalRateLimiter, ...token },
     car,
-    { onStoredChunk, maxRetries, decoders, signal } = {}
+    { onStoredChunk, maxRetries, maxChunkSize, decoders, signal } = {}
   ) {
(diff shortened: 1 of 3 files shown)


vasco-santos (approved):
LGTM

PR #2399, opened 2023-05-23
docs: add encodeDirectory example

alanshaw:
We have one for `encodeBlob`...not `encodeDirectory`...until now!


vasco-santos (approved):
lgtm

PR #2518, opened 2024-03-13
docs: add decisions/20240313-try-w3up.md

gobengo:
Motivation:
* write decision to implement first part of https://github.com/w3s-project/project-tracking/issues/7

Status
* [x] initial plan drafted
* [x] update based on feedback


decisions/20240313-try-w3up.md
```diff
@@ -0,0 +1,142 @@
...
+##### `W3_NFTSTORAGE_PRINCIPAL` Environment Variable
+
+configures how nft.storage will authenticate to web3.storage when sending invocations to store things. e.g. it may contain an ed25519 keypair as a CID with raw multihash.
+
+##### `W3_NFTSTORAGE_PROOF` Environment Variable
+
+configures the capabilities that nft.storage has access to when interacting with web3.storage to store nfts. These capabilities will usually be UCAN delegations whose audience is the identifier of `W3_NFTSTORAGE_PRINCIPAL`.
+W3_NFTSTORAGE_PROOF needs to have proof rooted in W3_NFTSTORAGE_SPACE that authorize W3_NFTSTORAGE_PRINCIPAL to store in W3_NFTSTORAGE_SPACE.
+
+##### `NFTSTORAGE_W3S_ALLOW_ACCOUNTS` Environment Variable
```
vasco-santos, line 71: I assume you want to use this for initial testing? If so, that sounds good, but then we can remove it as we want all the things

decisions/20240313-try-w3up.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Risks
+
+### degrading nft.storage's existing functionality
+
+We wouldn't want these changes to have any negative impact on users that rely on nft.storage's current setup. We only want to enable evaluation of new ways of delivering on that functionality that are seamless for end-users.
+
+#### Mitigation: feature switch for initial rollout
+
+We'll deploy this functionality to use web3.storage initially behind an opt-in feature switch. Once deployed, most uploads will not use the web3.storage functionality. Only accounts we add to a configured allow-list will get persisted to w3up. This removes risk of everyday users being impacted by bugs in the new code, while still enabling employees we allowlist to test the code in production. After testing in production, we can add more to the allowlist (e.g. perhaps nft.storage employees), or we can remove the feature switch to roll out the functionality for all users of nft.storage.
```
vasco-santos, line 89: ok, answered my first comment

decisions/20240313-try-w3up.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Does this project have any dependencies?
+
+#### dependency on collaboration with nft.storage
+
+This project can't succeed without a sponsor from nft.storage and willingness to try out the proposed changes to the nft.storage codebase in a production-like environment. If we have this, we can help prove out nft.storage storing data via web3.storage and relying on web3.storage productized services for serving on ipfs and filecoin dealmaking.
+
+## Alternative Solutions
+
+I don't know of any alternate solutions I can link to that provide a path to servicing nft.storage's web3 storage needs without relying on dagcargo. If you know of some, please add them here.
```
vasco-santos, line 112: use w3filecoin directly. But that seems already out of the table :)

vasco-santos (approved):
Thanks for putting this together. This is great

decisions/20240313-try-w3up.md
```diff
@@ -0,0 +1,142 @@
...
+## Goals
+
+* enable nft.storage to store new uploads using web3.storage
+* enable nft.storage to migrate to web3.storage, have an account there we have in our CRM, access console.web3.storage, etc
+
+## Non-Goals
+
+* migrate old nft.storage uploads
+* fully remove reliance on dagcargo
+  * however, this work may serve as useful input to that. If nft.storage can upload using w3up, then w3up can probably handle the filecoin aggregation via [w3-filecoin](https://github.com/web3-storage/specs/blob/main/w3-filecoin.md). And if we have that, it reduces reliance on dagcargo.
```
vasco-santos, line 27: @gobengo do you mean by this that first iteration is behind a feature flag via ENV VAR + specific accounts? Or do you mean for first iteration we still need dagcargo DB, but not the aggregator given all uploads will go via w3up, but we still need DB until second stage?

vasco-santos, line 27:
```suggestion
  * however, this work may serve as useful input to that. If nft.storage can upload using w3up, filecoin aggregation will happen via [w3-filecoin](https://github.com/web3-storage/specs/blob/main/w3-filecoin.md). And if we have that, it reduces reliance on dagcargo.
```

gobengo: > do you mean by this that first iteration is behind a feature flag via ENV VAR + specific accounts?

I mostly meant this. 'fully remove reliance on dagcargo' is a nongoal for first iteration of just making sure we can do at least some uploads via w3up. But I wrote this bullet to say that, while it's a non-goal this sprint, I hope this moves us closer to removing dagcargo by replacing it with w3up.

Assuming this scouting goal goes well, 'fully remove reliance on dagcargo' could be next goal.

vasco-santos: For me that is fine if the date dagcargo goes off is before sprint ends? worth double checking with @reidlw

decisions/20240313-try-w3up.md
gobengo: @vasco-santos Do you think using w3filecoin directly might be a more promising approach than what we describe here? It can be on the table, I just don't understand what that would entail and pros/cons.

decisions/20240313-try-w3up.md
gobengo: I made a note up above to make it more explicit: https://github.com/nftstorage/nft.storage/pull/2518/commits/8481203c5bbd1da8fc6986b1a5832b0bb615a16b

decisions/20240313-try-w3up.md
gobengo: Yep. I added a note to this section to make it more explicit this env var may be removed later once it has served its purpose to control rollout. https://github.com/nftstorage/nft.storage/pull/2518/commits/8481203c5bbd1da8fc6986b1a5832b0bb615a16b

decisions/20240313-try-w3up.md
gobengo: > For me that is fine if the date dagcargo goes off is before sprint ends? 

I think the sprint ends on March 27, and my last day of availability is March 22. I'll be back April 1. I don't see a path to dagcargo going off before the sprint ends.

decisions/20240313-try-w3up.md
vasco-santos: Here is the proposal for it I wrote a way back and discussed with Alan and Hannah https://hackmd.io/@vasco-santos/Skxt0yd5a/edit

w3filecoin was designed for multiple storefronts, so w3up/others, and my idea was to not go through w3up to avoid mixing the data at that level. And this together with the separate aggregation process for both. The general sense though is that using w3up makes more sense on a perspective on having nft.storage as a user of w3up in the long run. But if the expectation we have is it to be 3 months, making it a client of w3filecoin, while continuing to write to old bucket and so on, would make more sense to me.

But after talking with Alan, Hannah, Reid, we opted on the w3up lane, and that is why I say that it seems out of the table. I generally agree with using w3up because it is simpler if we won't care about splitting aggregation + have nft.storage in the long run as a client of w3up

decisions/20240313-try-w3up.md
```diff
@@ -0,0 +1,145 @@
+# nft.storage should try w3up for web3 storage of new uploads
+
+Authors
+- bengo@protocol.ai
+
+## Problem Statement
+
+nft.storage needs to offer web3 storage and currently does so using a homegrown solution that may be expensive to operate, maintain, and scale.
```
hannahhoward, line 8: It seems a little odd to not mention anywhere in this doc the reality that nft.storage's custom Filecoin pipeline (DAGCargo) is about to break for taking new data - so the experience explicitly will stop working. 

It feels like someone revisiting this ADR in a year might want to know this?

hannahhoward (approved):
Generally LGTM.

Time crunch issues around staged rollout with feature flag may be a thing, but that can be dealt with outside this doc.

decisions/20240313-try-w3up.md
gobengo: > not mention anywhere in this doc the reality that nft.storage's custom Filecoin pipeline (DAGCargo) is about to break for taking new data

Do you have any ideas for what I could link to that publicly explains that reality?

gobengo: @hannahhoward I guess what I'm feeling like is missing is: Where is the ADR for the root cause of why dagcargo is about to break? For me that isn't a decision I'm making, it's a research task to go find out the root cause. I could do that instead of implementing this plan, but I don't think we'll regret doing this plan as described this sprint no matter what.

gobengo: I mentioned dagcargo briefly. Thanks for the suggestion. https://github.com/nftstorage/nft.storage/pull/2518/commits/dc2bf27f8599038d886ad1d43a4795a6f4102112

decisions/20240313-try-w3up.md
gobengo: I added a section for this alternate solution. Thanks for sharing it and context. 🙏  https://github.com/nftstorage/nft.storage/pull/2518/commits/b3c22cdb8e70be29f4b6cf66f70e6f9f4c24e07e

decisions/20240313-try-w3up.md
```diff
@@ -0,0 +1,169 @@
...
+Prototype Happy path
+
+- [x] @gobengo set up nft.storage localdev to be able to hack on packages/api
+- [ ] nft.storage/api: can be configured with `W3_NFTSTORAGE_PRINCIPAL` and `W3_NFTSTORAGE_PROOF` and use them to get a w3up-client at runtime
+- [ ] nft.storage/api: modify POST /upload/ handler to store using w3up-client@12
+- [ ] nft.storage/api: add feature switch requiring allowlist for new uploads to store in web3.storage
+
+Rollout
+
+- [ ] release modified nft.storage to staging (production ok too) but with web3.storage disabled-by-default via feature switch
```
prodalex, line 143: Elizabeth is planning to release a new Website/console which is currently in development by some contractors. We need to understand what she is changing and what the impact on this roll out plan will be.

vasco-santos: this is an API change, so as soon a new website/console uses same API everything would be fine. But from what I know (or in other words, pieces I put together from all the meetings I attended), the new website/console they are putting together would be a complete new thing that would mean all of this is deprecated

gobengo: > We need to understand what she is changing and what the impact on this roll out plan will be.

sounds like a good thing for next sprint

PR #2522, opened 2024-03-19
feat: POST /uploads can write uploads to w3up for web3 serving and dagcargoless filecoin deal-making

gobengo:
Motivation:
* part of https://github.com/w3s-project/project-tracking/issues/7
* implement https://github.com/nftstorage/nft.storage/pull/2518

Note
* many files I modified only because the upgrade to new typescript compiler or eslint required it
* w3up-client requires node 18 so this drops support for node 16 from the api package
  * I think the client will still work on node 16, but I couldn't find a way of getting the ci for client to, using node 16, install just the client package deps and not the api package deps (which tries to install w3up-client, and then `yarn` errored because that requies node 18)
* w3up-client requires node 18, but everything had been running on node 16
  * [ ] if we can't find a way of continuing to use v16, then when this merges to main, we probably need to go into cloudflare workers settings and change NODE_VERSION from 16 to 18+


packages/api/src/utils/w3up.js
```diff
@@ -0,0 +1,135 @@
...
+import { connect } from '@ucanto/client'
+import { CAR, HTTP } from '@ucanto/transport'
+
+/**
+ * @param {object} env
+ * @param {string|undefined} [env.principal]
+ * @param {string|undefined} [env.proof]
+ */
+export async function getW3upClient({ principal, proof } = {}) {
+  const signer = principal ? ed25519.parse(principal) : await ed25519.generate()
```
alanshaw, line 20: Fail if undefined

packages/api/src/utils/w3up.js
```diff
@@ -0,0 +1,135 @@
...
+  } catch (/** @type {any} */ err) {
+    console.error(`Error: failed to import proof: ${err.message}`)
+    throw err
+  }
+}
+
+/**
+ * @param {import('@ucanto/interface').Delegation} delegation - delegation to encode
+ */
+export async function encodeDelegationAsCid(delegation) {
```
alanshaw, line 78: Move to test helpers?

packages/api/src/utils/w3up.js
```diff
@@ -0,0 +1,135 @@
...
+ * @param {object} options
+ * @param {string} options.url
+ * @param {string} options.principal
+ * @param {string} options.proof
+ */
+export async function createW3upClientFromConfig(options) {
+  const url = new URL(options.url)
+  const principal = ed25519.parse(options.principal)
+  const connection = connect({
+    id: { did: () => 'did:web:web3.storage' },
```
alanshaw, line 115: Need service DID to come from env too. If we change `W3UP_URL`, then it won't match...e.g. for staging.

packages/api/src/routes/nfts-upload.js
```diff
@@ -105,9 +111,48 @@ export async function nftUpload(event, ctx) {
...
+          cause: /** @type any */ (error)?.cause,
+        })
+        throw error
+      }
+    }
+  } else {
+    console.warn(
+      'nftUpload route skipping w3up storage due to missing ctx.w3up'
+    )
+  }
```
alanshaw, line 133: This needs to be pushed into `uploadCar` and the upload to w3up should happen _instead_ of upload to s3/r2. There is more than one route that calls this. e.g. https://github.com/nftstorage/nft.storage/blob/main/packages/api/src/routes/nfts-store.js

alanshaw (changes requested):

packages/api/src/routes/nfts-upload.js
travis: ah yea - I think this was intentional and intended to limit the potential scope of changes - is there any downside to pushing the data through both upload paths for now, aside from some extra computation and associated $ costs? If the only problem is that doing both uploads will result in extra infra costs for the next couple months, and we can limit potentially disruptive changes then I'm in favor of just having this code do the upload both ways and then potentially removing the "classic" upload path once we're confident things are working... what do you think?

alanshaw: I'm probbaly more pro replacing it. For one it makes it easier to test if everything is working or not - no false positives since it's not going through the old pipeline as well. It also saves money. I think it mostly boils down to a conditional over https://github.com/nftstorage/nft.storage/blob/06712ad02d8df58b654b77a3a4db0eb5fa213fb7/packages/api/src/routes/nfts-upload.js#L152-L155 ...but then linkdex API will not be able to determine completeness since it won't be put in S3. I reckon we could quite easily deploy an equivalent service that works on cloudflare and looks at R2.

packages/api/src/utils/w3up.js
travis: done

packages/api/src/utils/w3up.js
travis: done

packages/api/src/utils/w3up.js
travis: done

packages/api/src/routes/nfts-upload.js
travis: ok! I started to look into this but I think it's a decent chunk of work and I wouldn't be able to finish before the end of the day, so I'll leave this one for you tomorrow!

alanshaw (approved):

PR #2534, opened 2024-03-29
fix: upgrade @ipld/dag-cbor

travis:
we've hit a problem where the CBOR-decoding part of parsing the w3up delegation seems to be erroring and falling back to the JWT codepath here:

https://github.com/ipld/js-dag-ucan/blob/21bae4e0ad7ff76e0c7b73938dcc228726854a43/src/lib.js#L52

it's hard to tell exactly what's going wrong since this seems to work locally, so this is a shot in the dark - upgrading `@ipld/dag-cbor` should be harmless and may fix the problem we're seeing


travis:
merged without review since the code change is trivial and I'm anxious to get this into staging to test

PR #2535, opened 2024-03-29
fix: log proof for debugging

travis:
current theory is that this variable isn't what we expect - logging so we can debug in staging


travis:
merging since trivial code change

PR #2542, opened 2024-03-29
fix: add a fake piece hasher to try to fix staging upload issues

travis:
we're seeing memory issues in staging that may be due to the piece hasher - as suggested by @alanshaw fake it out to try to get rid of those issues


travis:
merging to test in staging

Gozala (approved):

PR #2546, opened 2024-04-01
chore: expose simple route to debug

Gozala:
I was thinking to try something along these lines

### DIFF
--- a/packages/api/src/repl.js
+++ b/packages/api/src/repl.js
@@ -1,5 +1,5 @@
 import { JSONResponse } from './utils/json-response.js'
-/** @type {import('../bindings').Handler} */
+/** @type {import('./bindings').Handler} */
 export const repl = async (event, ctx) => {
   const { w3up } = ctx
   const { headers } = event.request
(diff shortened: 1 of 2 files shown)


travis (approved):

packages/api/src/repl.js
```diff
@@ -0,0 +1,21 @@
+import { JSONResponse } from './utils/json-response.js'
+/** @type {import('../bindings').Handler} */
```
Gozala, line 2:
```suggestion
/** @type {import('./bindings').Handler} */
```

packages/api/src/repl.js
```diff
@@ -0,0 +1,21 @@
...
+    onShardStored: ({ cid }) => {
+      console.log(`SHARD ${cid}`)
+    },
+    // @ts-expect-error TODO adjust upstream type
+    pieceHasher: null,
+  })
+  console.log('UPLOADED CAR')
+
+  return new JSONResponse({ ok: true, value: result })
+}
```
Gozala, line 21:
```suggestion
}

```

PR #2548, opened 2024-04-02
fix: try older w3up client

hannahhoward:
# Goals

I found memory issues went away with older w3up-client in my upload test. 

This latest client appears not to be compatible with whatever APIs exist on local.storage, though perhaps web3.storage is updated -- I was actually trying to resolve this and accidentally discovered that I didn't have a memory crash any more.

Debugging in VS Code put the error in PieceHashing code, so it makes sense that the latest client, which adds filecoin/offer running locally, would potentially consume more memory.


gobengo (approved):
I think this is low risk to try (due to only affecting functionality enabled when feature switch allows), and will be interesting to know whether this helps avoid the OOM.

We had thought we were running with the piece hasher disabled, but this is a good way of trying it with it *definitely* disabled (i.e. running code from before it was added).

PR #2549, opened 2024-04-02
fix: Revert "fix: try older w3up client"

gobengo:
Reverts nftstorage/nft.storage#2548

Why?
* that downgrade of w3up-client didn't fix everything
* @hannahhoward mentioned that in local testing, she thinks that the newer version doesn't use extra memory when configured with a `null` pieceHasher, which is what we're doing
* I can't think of a good reason to be on old version vs latest version


hannahhoward (approved):
LGTM

travis (approved):

PR #2550, opened 2024-04-02
chore: simple test case to check limits

Gozala:
This updates `/repl` endpoint to simply upload CARs so we can test it's limits, if works out we can rewire primary endpoint.

### DIFF
--- a/packages/api/src/repl.js
+++ b/packages/api/src/repl.js
@@ -11,16 +11,11 @@ export const repl = async (event, ctx) => {
       throw new HTTPError('empty payload', 400)
     }
 
-    const result = await w3up.uploadCAR(blob, {
-      onShardStored: ({ cid }) => {
-        console.log(`SHARD ${cid}`)
-      },
-      // @ts-expect-error TODO adjust upstream type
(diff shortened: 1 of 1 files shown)


hannahhoward (approved):
LGTM

packages/api/src/repl.js
```diff
@@ -11,16 +11,11 @@ export const repl = async (event, ctx) => {
...
     }
 
-    const result = await w3up.uploadCAR(blob, {
-      onShardStored: ({ cid }) => {
-        console.log(`SHARD ${cid}`)
-      },
-      // @ts-expect-error TODO adjust upstream type
-      pieceHasher: null,
-    })
+    const car = await w3up.capability.store.add(blob)
```
travis, line 14: I believe this only invokes the `store/add` capability - I think we need to get the URL from the result and send the bits to it

travis: disregard! https://github.com/web3-storage/w3up/blob/6fa7797ba6f5c0bbd359c807e257882a7d5d6fb8/packages/upload-client/src/store.js#L99-L113

PR #2556, opened 2024-04-03
feat: lock w3up uploads feature switch open

travis:
commenting code out here because I'll make a cleanup pass to remove the switch entirely once this is stable in production

### DIFF
--- a/packages/api/test/nfts-upload.spec.js
+++ b/packages/api/test/nfts-upload.spec.js
@@ -173,7 +173,7 @@ test.serial('should forward uploads to W3UP_URL', async (t) => {
       headers: { Authorization: `Bearer ${client2.token}` },
       body: file,
     })
-    // should not have incremented
+    // should have incremented because feature switch allows w3up for all uploaders
     t.is(mockW3upStoreAddCount, storeAddCountBeforeClient2 + 1)
   }
 })
--- a/packages/api/test/nfts-upload.spec.js
+++ b/packages/api/test/nfts-upload.spec.js
@@ -330,92 +330,91 @@ test.serial('should upload a single CAR file', async (t) => {
   t.is(data.content.dag_size, 15, 'correct dag size')
 })
 
-test.serial(
-  'should check dag completness with linkdex-api for partial CAR',
-  async (t) => {
-    const client = await createClientWithUser(t)
-    const config = getTestServiceConfig(t)
-    const mf = getMiniflareContext(t)
+// TODO verify with @alanshaw that we don't need to do this in the new upload flow
+// TODO remove this once we remove legacy uploads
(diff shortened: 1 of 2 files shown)


packages/api/test/nfts-upload.spec.js
```diff
@@ -330,92 +330,91 @@ test.serial('should upload a single CAR file', async (t) => {
   t.is(data.content.dag_size, 15, 'correct dag size')
 })
 
-test.serial(
-  'should check dag completness with linkdex-api for partial CAR',
-  async (t) => {
-    const client = await createClientWithUser(t)
-    const config = getTestServiceConfig(t)
-    const mf = getMiniflareContext(t)
+// TODO verify with @alanshaw that we don't need to do this in the new upload flow
```
travis, line 333: @alanshaw @vasco-santos could you weigh in on whether this dag-completeness check still needs to happen? it looks like we are still doing this check but maybe need to mock out `upload/get` in order to get this test passing? it looks like it's failing here because we have not mocked this out: https://github.com/nftstorage/nft.storage/blob/main/packages/api/src/routes/nfts-upload.js#L193

should we mock this out to get the test passing or just skip/remove this test?

travis:
@alanshaw @vasco-santos - could you two take a look at the tests we're skipping here and verify that they seem like things we no longer need to test? @gobengo and I were just not highly confident and this feels like a good moment to double check that the things these tests are trying to verify are no longer parts of what this code is expected to do...

packages/api/test/nfts-upload.spec.js
```diff
@@ -174,7 +174,7 @@ test.serial('should forward uploads to W3UP_URL', async (t) => {
       body: file,
     })
     // should not have incremented
```
gobengo, line 176:
```suggestion
    // should have incremented because feature switch allows w3up for all uploaders
```

packages/api/test/nfts-upload.spec.js
vasco-santos, line 333: of course this means we should remove it also from the code
https://github.com/nftstorage/nft.storage/blob/ad33faf83ae1c994f040f6ad1539496b16ad264a/packages/api/src/routes/nfts-upload.js#L190-L206
https://github.com/nftstorage/nft.storage/blob/ad33faf83ae1c994f040f6ad1539496b16ad264a/packages/api/src/routes/nfts-upload.js#L220-L225

travis: yep! I'm planning on removing the whole legacy codepath plus these tests once this is running reliably in production

gobengo (approved):

gobengo:
afaict this will be safe. I've been testing uploads from an account allowed by the feature switch, and haven't gotten any errors in prod.

PR #2560, opened 2024-04-04
feat: NFT.storage changes

travis:
Make all (most?) of the requested changes from https://github.com/orgs/w3s-project/projects/1/views/1?pane=issue&itemId=57594940

Note that once this PR is deployed, new user signups will be disabled - I did not put these behind a feature flag as there doesn't seem to be a strong need for one.

A few caveats:

1) I put the same banner at the top of every page - since the menu is in the global layout it's tricky to get different banners displaying on different pages. This means that the requests to have different pages link to different URLs are challenging to accommodate, so I've gone down the easy path.
2) The error message for new users is a little janky - I followed the letter of the request but it's not terribly pretty:

Screenshot 2024-04-04 at 12 28 33 PM

We could make this prettier but it's nontrivial work, so I'm proposing we ship with this and iterate but would be open to iterating before we ship.

### DIFF
--- a/packages/api/src/errors.js
+++ b/packages/api/src/errors.js
@@ -284,3 +284,27 @@ export class ErrorAgentDIDRequired extends HTTPError {
   }
 }
 ErrorAgentDIDRequired.CODE = 'ERROR_AGENT_DID_REQUIRED'
+
+/**
+ * Error indicating a new user signup was denied and probably will be indefinitely,
+ * and the user should try a new product instead.
+ */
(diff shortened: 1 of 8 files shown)


gobengo:
lgtm, but worth fixing the failed check. I think it might just be due to some lint rules (that IMO aren't that helpful here) https://github.com/nftstorage/nft.storage/actions/runs/8560717653/job/23460342853?pr=2560#step:5:121

gobengo (approved):

PR #2573, opened 2024-04-10
feat: show deal info from w3up

travis:
Fetch deal information from w3up.

There's no way to "multi-get" deal info for more than one upload so I've reworked the deal info in the files list - if it has deals from dagcargo it will just use those, and if it doesn't, it will display a "load deals" button:

Screenshot 2024-04-12 at 12 07 20 PM

When clicked, it will find the shards of the upload and find filecoin info for each of them. If it doesn't find any filecoin info in w3up it will display the "queuing" status and tooltip as it did in the past:

Screenshot 2024-04-12 at 12 07 27 PM
Screenshot 2024-04-12 at 12 07 34 PM

This required updating the "get NFT" API endpoint to look for filecoin info from w3up.

### DIFF
--- a/packages/api/src/utils/w3up.js
+++ b/packages/api/src/utils/w3up.js
@@ -1,7 +1,7 @@
 import * as W3UP from '@web3-storage/w3up-client'
 import * as ed25519 from '@ucanto/principal/ed25519'
 import { StoreMemory } from '@web3-storage/access/stores/store-memory'
-import contentClaims from '@web3-storage/content-claims/client'
+import * as contentClaims from '@web3-storage/content-claims/client'
 import { CID } from 'multiformats/cid'
 import { base64 } from 'multiformats/bases/base64'
 import { identity } from 'multiformats/hashes/identity'
--- a/packages/api/src/utils/w3up.js
+++ b/packages/api/src/utils/w3up.js
@@ -1,12 +1,14 @@
 import * as W3UP from '@web3-storage/w3up-client'
 import * as ed25519 from '@ucanto/principal/ed25519'
 import { StoreMemory } from '@web3-storage/access/stores/store-memory'
+import contentClaims from '@web3-storage/content-claims/client'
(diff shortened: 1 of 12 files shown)


packages/api/src/utils/w3up.js
```diff
@@ -106,3 +107,95 @@ export async function createW3upClientFromConfig(options) {
...
+        const info = result.ok
+        for (const deal of info.deals) {
+          filecoinInfos.push({
+            pieceCid: info.piece.toString(),
+            status: 'published',
+            batchRootCid: deal.aggregate.toString(),
+            miner: deal.provider,
+            chainDealID: Number(deal.aux.dataSource.dealID),
+            // TODO: figure this out
+            datamodelSelector: '',
```
travis, line 189: @alanshaw @vasco-santos @gobengo do any of you know what I should be using for `datamodelSelector` here? it's required and looks like `Links/19/Hash/Links/46/Hash/Links/0/Hash` in fixtures... is there a magic incantation that makes sense for deal info returned from w3up?

alanshaw: Yes, empty string is all we can do here. We could _add_ a field for the inclusion proof when the deal comes from the new pipeline, but not a priority right now.

packages/api/test/utils/w3up-testing.js
```diff
@@ -55,6 +57,21 @@ export async function createMockW3up(
...
+        const result = await options.onHandleFilecoinInfo?.(invocation)
+        if (result) {
+          return {
+            ok: result,
+          }
+        } else {
+          return {
+            error: {
+              name: 'UnexpectedError',
+              message: `onUploadGet was not defined or return ${result}`,
```
alanshaw, line 70:
```suggestion
              message: `onFilecoinInfo was not defined or return ${result}`,
```

alanshaw (approved):

vasco-santos (approved):

PR #2580, opened 2024-04-12
chore: remove legacy upload codepath

travis:
plus feature flag and related tests


vasco-santos (approved):
LGTM

Not sure if you are going for a second PR to remove linkdex, but I think we should/could also remove https://github.com/nftstorage/nft.storage/blob/feat/cleanup-w3up-feature-flag/packages/api/src/routes/nfts-upload.js#L219-L222

I think that once we write to R2 only that will fail, but I am not super aware on how linkdex is built

PR #2593, opened 2024-04-17
fix: more logging of upload get fail

travis:


travis:
merging without review so I can debug in staging

